PR #8, opened 2018-08-07
style(component) adding padding to Dropdown component

tagraha:
adding padding to Dropdown Component


tagraha:
adding some adjustment for dropdown component

PR #9, opened 2018-08-08
fix(component): fix slider component next/previous icon

tagraha:
Next/Prev icon on slider (color change)


tagraha:
adding disable color on Next/Prev icon (slider component)

PR #12, opened 2018-08-15
feature/job overview

tagraha:
JobOverview Component


tagraha:
Adding JobOverview component

PR #13, opened 2018-08-16
feat(component): adding collapsible component

tagraha:
Collapsible Component including the storybook doc


tagraha:
adding new component `<Collapsible>`

PR #34, opened 2018-10-16
Add in 1 more prop for job card target behaviour

seahyc:
Added target prop so that we can control the link opening behaviour. Updated documentation to reflect change. Also included default value in documentation.

### DIFF
--- a/stories/Application/JobCardStory.js
+++ b/stories/Application/JobCardStory.js
@@ -104,7 +104,7 @@ const JobCardStory = () => (
           <td>_blank|_self|_parent|_top|framename</td>
           <td>_self</td>
           <td>no</td>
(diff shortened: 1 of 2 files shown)


stories/Application/JobCardStory.js
```diff
@@ -92,7 +93,16 @@ const JobCardStory = () => (
...
+          <td>no</td>
+          <td>redirect URL when clicked</td>
+        </tr>
+        <tr>
+          <td>target</td>
+          <td>string</td>
+          <td>_blank|_self|_parent|_top|framename</td>
+          <td>_self</td>
           <td>no</td>
           <td>redirect URL when clicked</td>
```
fredyyanto: I think, the description can be "specifies where to open the linked document"

PR #38, opened 2018-10-19
Feature/accessibility

fredyyanto:


src/Application/JobCard/JobCardBody.js
```diff
@@ -2,11 +2,25 @@ import React from 'react';
...
-const JobCardBody = ({ children, ...defaultProps }:Props) => (
-  <JobcardBodyWrapper {...defaultProps}>
-    {children}
-  </JobcardBodyWrapper>
-);
+const JobCardBody = (props: Props) => {
+  const {
+    children,
+    className,
+    ...defaultProps
```
seahyc: nitpick: do you have any lint checks? we usually practise trailing commas

fredyyanto: yeah, we use lint.. once you put the comma, it says "[eslint] Parsing error: A trailing comma is not permitted after the rest element"

PR #72, opened 2019-06-10
Hotfix/search-filter-closes-on-scroll

franciscrispin:
Fixed problem of SearchFilter dropdown automatically closing when scrollbar is clicked

- Before:
  - We passed a handleOpen function to the input's onBlur prop which toggles the isOpen state everytime the input loses focus.
  - This caused the dropdown to close everytime the dropdown's scrollbar is clicked because the input loses focus.
- After:
  - We added a handleMouseDown eventListener which ignores the mousedown event if it is on the scrollbar.
  - Moving forward, the dropdown will only close when a mousedown event takes place outside the component or when a dropdown list item is clicked.


seahyc:
lgtm!

PR #77, opened 2019-07-01
ButtonStyle: Add active color for `PrimaryBtn`

franciscrispin:
Changed the `active` color of the `PrimaryBtn` to white.


franciscrispin:
@tagraha PTAL. Just a tiny change to fix this bug 

https://trello.com/c/AnPN7kl4/3220-when-users-clicked-the-apply-icon-it-should-be-changed-the-font-into-white-reported-by-ven

PR #79, opened 2019-08-01
Progress: Extend functionality with new props

franciscrispin:
- `content`: sets the content inside the component
- `percentageRange`: sets the percentage range to display a different color on the progress component when the percentage changes
- `size`: sets the size of the component

### DIFF
--- a/src/Display/Progress/Progress.tsx
+++ b/src/Display/Progress/Progress.tsx
@@ -86,7 +86,7 @@ const Progress: React.FunctionComponent<Props> = (props) => {
 interface Props extends Omit<React.ComponentPropsWithoutRef<typeof ProgressContainer>, 'progress'> {
   percentage: number;
   percentageRange?: [] | [number] | [number, number];
-  size?: 1|2|3|4|5|6|7|8|9|10;
+  size?: number;
   content?: React.ReactNode;
 }
 
--- a/src/Display/Progress/Progress.tsx
+++ b/src/Display/Progress/Progress.tsx
@@ -14,36 +14,43 @@ import { PrimaryColor, SecondaryColor } from '../../Style/Colors';
 const Progress: React.FunctionComponent<Props> = (props) => {
   const {
     className,
-    percentage,
+    percentage = 100,
     percentageRange = [50],
     size = 8,
     content,
     ...defaultProps
   } = props;
-  
+
+  let normalizedPercentage;
+  let normalizedSize;
+
   if (percentage < 0 || percentage > 100) {
-    throw new Error(`Error: percentage prop on Progress component expected a value between 0-100. Received ${percentage} instead.`)
-  }  
-
+    normalizedPercentage = percentage < 0 ? 0 : percentage > 100 ? 100 : percentage;
+    console.warn(`Invalid prop value on Progress component: percentage prop expected a value between 0-100. Received ${percentage} instead.`)
+  } else {
+    normalizedPercentage = percentage;
+  }
   if (percentageRange.length > 2) {
-    throw new Error(`Error: percentageRange prop on Progress component expected a maximum of 2 numbers. Received [${percentageRange}] instead.`)
+    console.warn(`Invalid prop value on Progress component: percentageRange prop expected an array with a maximum of 2 numbers. Received [${percentageRange}] instead.`)
+  }
+  if (size < 1|| size > 10) {
+    normalizedSize = size < 1 ? 1 : size > 10 ? 10 : size;
+    console.warn(`Invalid prop value on Progress component: size prop expected a value between 1-10. Received ${size} instead.`)
+  } else {
+    normalizedSize = size;
   }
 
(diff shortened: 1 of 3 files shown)


src/Style/Display/ProgressStyle.ts
```diff
@@ -11,12 +11,17 @@ const spinning = keyframes`
   }
 `;
 
-export const ProgressContent = styled.div`
+export const ProgressContent = styled.div<ProgressContentProps>`
+  size: ${((props: ProgressContentProps) => props.size) as any};
```
no13000: Is `size` a legit css attribute? 😮

src/Display/Progress/Progress.tsx
```diff
@@ -2,50 +2,85 @@ import * as React from 'react';
...
     className,
     percentage,
+    percentageRange = [50],
+    size = 8,
+    content,
     ...defaultProps
   } = props;
+  
+  if (percentage < 0 || percentage > 100) {
+    throw new Error(`Error: percentage prop on Progress component expected a value between 0-100. Received ${percentage} instead.`)
```
no13000: Hmm as a user I'd prefer you to normalize the value for me instead of throwing an `Error` here which could potentially crash the app if I did not catch it properly

src/Display/Progress/Progress.tsx
```diff
@@ -2,50 +2,85 @@ import * as React from 'react';
...
+  }  
 
-  const progressValue = percentage > 100 ? 282.6 * (1 - (100 / 100)) : 282.6 * (1 - (percentage / 100));
-  const color = percentage > 50 ? SecondaryColor.green : SecondaryColor.orange;
+  if (percentageRange.length > 2) {
+    throw new Error(`Error: percentageRange prop on Progress component expected a maximum of 2 numbers. Received [${percentageRange}] instead.`)
+  }
+
+  if (size > 10 || size < 1) {
+    throw new Error(`Error: size prop on Progress component expected a number between 1 - 10. Received ${size} instead.`)
```
no13000: Same as above comment

src/Display/Progress/Progress.tsx
```diff
@@ -2,50 +2,85 @@ import * as React from 'react';
...
   } = props;
+  
+  if (percentage < 0 || percentage > 100) {
+    throw new Error(`Error: percentage prop on Progress component expected a value between 0-100. Received ${percentage} instead.`)
+  }  
 
-  const progressValue = percentage > 100 ? 282.6 * (1 - (100 / 100)) : 282.6 * (1 - (percentage / 100));
-  const color = percentage > 50 ? SecondaryColor.green : SecondaryColor.orange;
+  if (percentageRange.length > 2) {
+    throw new Error(`Error: percentageRange prop on Progress component expected a maximum of 2 numbers. Received [${percentageRange}] instead.`)
```
no13000: Same as above, perhaps can just ignore the extra values in the array

src/Style/Display/ProgressStyle.ts
franciscrispin: Oh I'm not familiar with typescript but I thought this allowed me to use the `props.size` below.
```
font-size: ${props => `${props.size}em`};
```
Lol sorry was blindly copying and pasting code.

src/Display/Progress/Progress.tsx
franciscrispin: Oh yeah first time dealing with such cases, wasn't sure how to handle the errors also. What do you mean by normalizing the value?

no13000: Meaning to make the `percentage` bound within 0 - 100
You can either use `lodash.clamp`
https://lodash.com/docs/4.17.15#clamp
or see
https://stackoverflow.com/questions/5842747/how-can-i-use-javascript-to-limit-a-number-between-a-min-max-value

franciscrispin: Oh I see. Think I will use this from the stack overflow link because I prefer not to add another dependency in glints-aries.
```
number = number > 100 ? 100 : number < 0 ? 0 : number;
```
@no13000 Do you think it's still useful to warn the user with something like `console.warn`? Afraid if we don't show any errors/warnings, the devs will not know that something is wrong with their props.

no13000: @franciscrispin sure that'll be alright

src/Style/Display/ProgressStyle.ts
```diff
@@ -11,12 +11,17 @@ const spinning = keyframes`
...
+export const ProgressContent = styled.div<ProgressContentProps>`
+  size: ${((props: ProgressContentProps) => props.size) as any};  
+
   position: relative;
   outline: none;
+  font-size: ${props => `${props.size}em`};
 `;
 
 export const ProgressContainer = styled.div<ProgressContainerProps>`
+  size: ${((props: ProgressContentProps) => props.size) as any};
```
no13000: Same as the comment above btw

PR #85, opened 2019-08-13
Hotfix/select infinite loading bug and refactor <Select> component

franciscrispin:
- Fix infinite loading bug when no options are returned
- Refactor `SelectList` into separate component
- Add `noOptionResult` prop in `SelectStory` and simplify the usage example


franciscrispin:
@no13000 Yo tony PTAL

Issue #97, opened 2019-08-26
Fix re-rendering of Alert Box, when autoClose is kept as a dynamic parameter.


ritikamotwani:
When changing the value of autoClose, the Alert Box will take in the value of the previous change and work accordingly.

PR #98, opened 2019-08-26
Fix rendering of Alert Box when autoClose changes dynamically

ritikamotwani:
Fixes: #97 

The Alert Box works according to the current value of the autoClose instead of taking the previous one.


franciscrispin:
@ritikamotwani LGTM

Issue #95, opened 2019-08-17
Wrong color used for different states of <Breadcrumb> 


franciscrispin:
[Trello Link](https://trello.com/c/3fgIY3DA/51-unexpected-behaviour-of-breadcrumb-reported-by-ven)

ritikamotwani:
I would like to work on this.
Another Trello card related to this: https://trello.com/c/zYxi9e62/3275-unexpected-behaviour-of-breadcrumb-reported-by-ven

PR #99, opened 2019-08-27
Fix the unexpected behavior of BreadCrumb

ritikamotwani:
1. In the active BreadCrumb item : Colour changed to grey and the cursor changed to default.
2. For the non active ones: Colour changed to Black.

To view the changes after installing the project :
In the file ` BreadcrumbStory.tsx ` , change the active Breadcrumb Item to div. (For now the active element is written in anchor tags which makes the cursor to be pointer by default.)

image

Fixes: #95


franciscrispin:
@ritikamotwani LGTM

Issue #104, opened 2019-08-28
<Select> component filter old options instead of displaying new options


BenedictSu:
When the new list of options has the same length as the current list, the existing check `if (React.Children.count(nextProps.children) !== prevState.childrenLength) {` assumes that there is no change in the children values and does not update the `filterVaule` state with the new list of options.

PR #105, opened 2019-08-28
Fix <Select> component filtering old options instead of displaying new options

BenedictSu:
src/Input/Select/Select.tsx: check for difference in children instead of just length when updating the state for filterValue

Fixes: https://github.com/glints-dev/glints-aries/issues/104

### DIFF
--- a/src/Input/Select/Select.tsx
+++ b/src/Input/Select/Select.tsx
@@ -56,31 +56,37 @@ class Select extends React.Component<Props, State> {
   }
 
   componentDidUpdate(prevProps: Props) {
-    const hasDifferentChildren = !isEqual(this.props.children,  prevProps.children);
+    const {
+      children,
+      isLoading,
+      value,
+    } = this.props;
+
+    const hasDifferentChildren = !isEqual(children,  prevProps.children);
     if (hasDifferentChildren) {
       this.setState({
-        selectedValue: this.props.value as string,
-        defaultValue: this.props.value as string,
-        floating: !!this.props.value,
-        filterValue: React.Children.map(this.props.children, data => data),
+        selectedValue: value as string,
+        defaultValue: value as string,
+        floating: !!value,
+        filterValue: React.Children.map(children, data => data),
       });
-    } else if (this.props.value !== prevProps.value) {
-      if (this.props.value && this.props.value !== this.state.defaultValue) {
+    } else if (value !== prevProps.value) {
+      if (value && value !== this.state.defaultValue) {
         this.setState({
-          selectedValue: this.props.value as string,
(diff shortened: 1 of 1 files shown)


src/Input/Select/Select.tsx
```diff
@@ -61,6 +55,38 @@ class Select extends React.Component<Props, State> {
...
+          floating: true,
+        });
+      } else if (this.props.value === '') {
+        this.setState({
+          selectedValue: this.props.value,
+          defaultValue: this.props.value,
+          floating: false,
+        });
+      }
+    } else if (this.props.isLoading !== prevProps.isLoading) {
```
franciscrispin: Think this can be simplified to remove the nested if condition
```
else if (this.props.isLoading && this.props.isLoading !== prevProps.isLoading) {
  this.setState({
    isLoading: this.props.isLoading,
  });
}
```

src/Input/Select/Select.tsx
```diff
@@ -61,6 +55,38 @@ class Select extends React.Component<Props, State> {
     document.addEventListener('click', this.handleOnBlur, false);
   }
 
+  componentDidUpdate(prevProps: Props) {
+    const hasDifferentChildren = !isEqual(this.props.children,  prevProps.children);
```
franciscrispin: This is just personal formatting preference but you can consider destructuring the props object to remove the need for `this.props`

`const { children, value, isLoading } = this.props`

src/Input/Select/Select.tsx
```diff
@@ -61,6 +55,38 @@ class Select extends React.Component<Props, State> {
...
+    if (hasDifferentChildren) {
+      this.setState({
+        selectedValue: this.props.value as string,
+        defaultValue: this.props.value as string,
+        floating: !!this.props.value,
```
franciscrispin: Again, more of personal preference but I'd prefer to make this more verbose by using `Boolean()` instead of the double bang `!!`. Some may argue that it is less performant but it is probably quite negligible so I'd go for more readable code :smile:

src/Input/Select/Select.tsx
BenedictSu: agreed, updated

src/Input/Select/Select.tsx
BenedictSu: got it, updated

src/Input/Select/Select.tsx
BenedictSu: got it, updated!

BenedictSu:
@franciscrispin thank you for the review and the insights! I have followed up on your suggestions

PR #112, opened 2019-09-14
Tabs: Refactor to function component with hooks and adjust tab index to allow for changing tabs with tab key

franciscrispin:
- add props on Tabs to allow devs to set the activeTab and the onTabClick behaviour
- add label prop on children to set the label when it is passed to TabPane
- adjust tab index to allow users to press tab key to switch focus of tabs
- adjust tab index to allow users to press enter to switch to a tab in focus

### DIFF
--- a/src/Display/Tabs/Tabs.tsx
+++ b/src/Display/Tabs/Tabs.tsx
@@ -10,84 +10,70 @@ import {
   TabsBody,
 } from '../../Style/Display/TabsStyle';
 
-class Tabs extends React.Component<Props, State> {
-  static Pane = TabPane;
+const Tabs: Tabs = ({activeTab, onTabClick, children, className}) => {
+  const [currentIndex, setCurrentIndex] = React.useState(0);
+  const activeTabOrIndex: any = activeTab || currentIndex;
+  const onClick = onTabClick || setCurrentIndex;
 
-  constructor(props: Props) {
-    super(props);
-    this.state = {
-      currentIndex: 0,
-    };
-  }
+  const childrenArray = React.Children.toArray(children);
+  const currentChild = childrenArray[activeTabOrIndex];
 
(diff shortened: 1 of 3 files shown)


franciscrispin:
Yo @lsping8, need your help to code review this :pray: 

- Added some props to the tab component to enable it to work with the new tabulation feature on `dst` https://trello.com/c/VOJXGW4a/195-m141-tabulation-tab-on-job-search-page-for-you-tab-2sp-gun-5-sp-francis
- Also adjusted the tab index so users can press tab then enter to switch tabs

src/Display/Tabs/Tabs.tsx
```diff
@@ -10,84 +10,70 @@ import {
   TabsBody,
 } from '../../Style/Display/TabsStyle';
 
-class Tabs extends React.Component<Props, State> {
-  static Pane = TabPane;
+const Tabs: Tabs = ({activeTab, onTabClick, children, className}) => {
+  const [currentIndex, setCurrentIndex] = React.useState(0);
+  const activeTabOrIndex: any = activeTab || currentIndex;
```
lsping8: @franciscrispin try not to use type `any`, in this case you can just use union types

src/Display/Tabs/Tabs.tsx
```diff
@@ -10,84 +10,70 @@ import {
...
+  const childrenArray = React.Children.toArray(children);
+  const currentChild = childrenArray[activeTabOrIndex];
 
-  handleClickTab = (index: number) => {
-    const listener = () => {
-      this.setState({
-        currentIndex: index,
-      });
-    };
+  const handleTabClick = (tab: any) => {
```
lsping8: @franciscrispin here also, try not to use `any`

franciscrispin:
@lsping8 PTAL again!

PR #120, opened 2019-10-02
Badge: Fix bug where a label of 0 causes the badge not to show

franciscrispin:
https://github.com/glints-dev/glints-aries/issues/119
- Change the type of label to string or number
- Only hide the badge when label is null or undefined


rosslo:
LGTM 😄

PR #121, opened 2019-10-02
Tabs: Restore the previous classNames

ritikamotwani:
Restore the previous classNames so that the current code is not disturbed.

### DIFF
--- a/src/Display/Tabs/Tabs.tsx
+++ b/src/Display/Tabs/Tabs.tsx
@@ -35,10 +35,10 @@ const Tabs: Tabs = (
   };
 
   return (
-    <TabsContainer className={classNames(`${alignment}-aries-tabs`, className)}>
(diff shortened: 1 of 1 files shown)


src/Display/Tabs/Tabs.tsx
```diff
@@ -49,7 +49,8 @@ const Tabs: Tabs = (
                     `tab-${tabLabel}`,
                     { active: activeTabOrIndex === tabLabel },
                     `${alignment}-tab`,
-                    `${variant}`
+                    `${variant}`,
+                    'tab'
```
franciscrispin: Was the className `'tab'` present before? If not we can leave it out.

ritikamotwani: No it wasn't.
Removed it! :smile:

PR #127, opened 2019-10-19
<Modal> Add unit tests

franciscrispin:


src/Display/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const modalDialog = getByRole('dialog');
+  return [modalContainer, onClose, modalDialog];
+}
+
+it('<Modal> should render with a title, content, footer and an onClick handler', () => {
+  const ModalSnapshot = renderer.create(OpenedModal).toJSON();
+  expect(ModalSnapshot).toMatchSnapshot();
+});
+
+describe('when modal is opened', () => {
```
iamfeek: Hey @franciscrispin , there is another `describe('when modal is opened')` block at `Line 62`. 

I believe this can be merged together.

src/Display/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+  });
+
+  it('should not be closed if the modal is clicked', () => {
+    const [, onClose, modalDialog] = setupModal(true);
+    fireEvent.click(modalDialog as Element);
+    expect(onClose).toHaveBeenCalledTimes(0);
+  });
+});
+
+describe('when modal is opened', () => {
```
iamfeek: Hey @franciscrispin, I have a question.

The group of tests here are checking whether or not the title and content and footer is present in accordance to the props.

However, at `Line 44`, there is a snapshot check. 

Is there a reason why this is repeated?
Also, which is a healthier way of testing?

src/Display/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+  it('should display the content', () => {
+    const { getByText } = render(OpenedModal);
+    const modalContent = getByText(props.content);
+    expect(modalContent).toHaveTextContent(props.content);
+  });
+
+  it('should display the footer', () => {
+    const { getByText } = render(OpenedModal);
+    const modalFooter = getByText(props.footer);
+    expect(modalFooter).toHaveTextContent(props.footer);
```
iamfeek: hey @franciscrispin , this checks that we have a `textContent`. 

In `setupModal`, the modal is given a `button` element that will then render `props.footer`.

I believe that we should also check that the button is rendered.
And that the button has this text content?

When I bring in the snapshot test, then it starts to get clearer also.
I understand that snapshot testing would also inspect HTML elements and their children?

src/Display/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+});
+
+describe('when modal is closed', () => {
+  it('should not be visible', () => {
+    const [modalContainer] = setupModal(false);
+    expect(modalContainer).toHaveStyle('visibility: hidden');
+  });
+});
+
+describe('onClose should be called once when:', () => {
```
iamfeek: Hey @franciscrispin 

There needs to be a test to cover the click on the `X` icon at the modal's header.

src/Display/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+    title={props.title}
+    isVisible={true}
+    footer={[<button key="button">{props.footer}</button>]}
+    onClose={props.onClose}
+  >
+    <p>{props.content}</p>
+  </Modal>
+);
+
+function setupModal(isVisible: boolean) {
```
iamfeek: hey @franciscrispin 

When referencing storybook, I saw a few more props that this modal can consume.

I believe that there is merit in testing the `size`, `hideHeader`, and `hideContentArea` props. 
I lack knowledge in testing for the `centering` prop -- I will need your advice on that.
(if you are also not aware, lets pair program this!)

src/Display/Modal/Modal.test.tsx
franciscrispin: Haha yeah they can. But wanted to separate them for clarity cause they didn't feel like they all fit together. But I guess i'll put them together anyway!

src/Display/Modal/Modal.test.tsx
franciscrispin: Oh hmm I think snapshot testing at the component level is to ensure that the component as a whole does not change unexpectedly. 

From the docs https://jestjs.io/docs/en/snapshot-testing
```
Snapshot tests are a very useful tool whenever you want to make sure your UI does not change unexpectedly.
```

Whereas the unit tests here are to test a specific part of the component. In a sense it is repeated but the purpose of the snapshot test is not to test each specific part, but more of testing the component as a whole.

src/Display/Modal/Modal.test.tsx
franciscrispin: Okay will check for the button!

Yeap, snapshot testing would also inspect HTML elements and their children.

src/Display/Modal/Modal.test.tsx
franciscrispin: Oh yes, was in my comments but I forgot about it lol

src/Display/Modal/Modal.test.tsx
franciscrispin: Hmm yeah I was considering those. My initial idea was to make sure every single prop was tested. I'm not so sure whether that's a good idea cause I would probably end up testing implementation details like checking for whether `align-items: center` (`centering`) exists or whether `background: transparent` (`hideContentArea`) exists. 

I recalled this from the `react-testing-library` [docs](https://testing-library.com/docs/intro) which seems to imply that doing the above is unfavourable. 
```
As a part of this goal, you want your tests to avoid including implementation details so refactors of your components (changes to implementation but not functionality) don't break your tests and slow you and your team down.
```

I'm not sure what's the right approach for this also so if you have any thoughts please do share them.

franciscrispin:
@iamfeek PTAL again!

src/Display/Modal/Modal.test.tsx
iamfeek: > As a part of this goal, you want your tests to avoid including implementation details so refactors of your components (changes to implementation but not functionality) don't break your tests and slow you and your team down.

I find that line above interesting and really changes the way we approach tests. 
I do agree that checking for implementation details is not healthy. 

I have a possible solution to test for `hideHeader` and `hideContentArea`. 
What if we test using the `has text` thing?
That way, _we are testing that the header text does not appear_. 

What do you think?

src/Display/Modal/Modal.test.tsx
```diff
@@ -75,26 +77,32 @@ describe('when modal is opened', () => {
   it('should display the footer', () => {
     const { getByText } = render(OpenedModal);
     const modalFooter = getByText(props.footer);
-    expect(modalFooter).toHaveTextContent(props.footer);
+    expect(modalFooter).toContainHTML(`<button>${props.footer}</button>`);
```
iamfeek: is this equivalent to _testing for implementation detail_?

src/Display/Modal/Modal.test.tsx
franciscrispin: I realised `hideContentArea` might be an unnecessary prop which may be deleted so I left out tests for that. Added test for `size`, `centering` and `hideHeader`. For the first two, I had to test the styling itself for it to work. Don't think there's another way around it.

src/Display/Modal/Modal.test.tsx
franciscrispin: Yeah it is. I don't think there is any other way to test for the `button` though. I think I prefer to leave it as `toHaveTextContent(props.footer);` in this case.

franciscrispin:
@iamfeek PTAL again :pray:

Issue #89, opened 2019-08-17
Create <Typography> component according to Glints UI Text Styles


franciscrispin:
Create `h1` - `h6` variants

https://app.zeplin.io/project/5ba343bf0c399c33db483653/screen/5bc6d879fbe15b18ca619fb0

PR #133, opened 2019-10-26
Feature/typography

franciscrispin:
- Created Typography component as per issue https://github.com/glints-dev/glints-aries/issues/89
- `Typography.Title` renders h1-h6 tags 
- `Typography.Paragraph` renders a paragraph tag with variants `'subtitle' | 'regular' | 'caption' | 'smallest'`

Glints UI: https://app.zeplin.io/project/5ba343bf0c399c33db483653/screen/5bc6d879fbe15b18ca619fb0

### DIFF
--- a/src/General/Typography/Paragraph.test.tsx
+++ b/src/General/Typography/Paragraph.test.tsx
@@ -15,7 +15,7 @@ const paragraphText =
   'As the first sign in the zodiac, the presence of Aries always marks the beginning of something energetic and turbulent. They are continuously looking for dynamic, speed and competition, always being the first in everything - from work to social gatherings.';
 
 describe('displays the correct text and font-size for each variant:', () => {
-  const variants: paragraphType[] = Object.values(PARAGRAPH_VARIANTS);
+  const variants = Object.values(PARAGRAPH_VARIANTS) as paragraphType[];
   variants.forEach(variant => {
     const fontSize = PARAGRAPH_FONT_SIZES[variant];
     it(`${variant}: ${fontSize}px`, () => {
--- a/src/General/Typography/Paragraph.test.tsx
+++ b/src/General/Typography/Paragraph.test.tsx
@@ -5,18 +5,26 @@ import { PrimaryColor } from '../../Style/Colors';
 import '@testing-library/jest-dom/extend-expect';
 import { render } from '@testing-library/react';
 
+import {
+  PARAGRAPH_FONT_SIZES,
+  PARAGRAPH_VARIANTS,
+  paragraphType,
+} from './ParagraphStyles';
+
 const paragraphText =
   'As the first sign in the zodiac, the presence of Aries always marks the beginning of something energetic and turbulent. They are continuously looking for dynamic, speed and competition, always being the first in everything - from work to social gatherings.';
 
-describe('displays the correct text for each variant:', () => {
-  const variants: any[] = ['subtitle', 'regular', 'caption', 'smallest'];
+describe('displays the correct text and font-size for each variant:', () => {
+  const variants: paragraphType[] = Object.values(PARAGRAPH_VARIANTS);
   variants.forEach(variant => {
-    it(`${variant}`, () => {
+    const fontSize = PARAGRAPH_FONT_SIZES[variant];
+    it(`${variant}: ${fontSize}px`, () => {
       const { getByText } = render(
         <Paragraph variant={variant}>{paragraphText}</Paragraph>
       );
       const paragraphTag = getByText(paragraphText);
-      expect(paragraphTag).toHaveTextContent(paragraphText);
+      expect(paragraphTag).toBeTruthy();
+      expect(paragraphTag).toHaveStyle(`font-size: ${fontSize}px`);
     });
   });
 });
--- a/src/General/Typography/Paragraph.test.tsx
+++ b/src/General/Typography/Paragraph.test.tsx
@@ -1,0 +1,46 @@
+import * as React from 'react';
+import Paragraph from './Paragraph';
+import { PrimaryColor } from '../../Style/Colors';
+
+import '@testing-library/jest-dom/extend-expect';
+import { render } from '@testing-library/react';
+
+const paragraphText =
+  'As the first sign in the zodiac, the presence of Aries always marks the beginning of something energetic and turbulent. They are continuously looking for dynamic, speed and competition, always being the first in everything - from work to social gatherings.';
+
+describe('displays the correct text for each variant:', () => {
+  const variants: any[] = ['subtitle', 'regular', 'caption', 'smallest'];
+  variants.forEach(variant => {
+    it(`${variant}`, () => {
+      const { getByText } = render(
+        <Paragraph variant={variant}>{paragraphText}</Paragraph>
+      );
+      const paragraphTag = getByText(paragraphText);
+      expect(paragraphTag).toHaveTextContent(paragraphText);
+    });
+  });
+});
+
+it('displays the text in bold', () => {
+  const { getByText } = render(<Paragraph bold>{paragraphText}</Paragraph>);
+  const paragraphTag = getByText(paragraphText);
+  expect(paragraphTag).toHaveStyle('font-weight: bold');
+});
+
+it('displays the correct color', () => {
+  const { getByText } = render(
+    <Paragraph color={PrimaryColor.glintsred}>{paragraphText}</Paragraph>
+  );
+  const paragraphTag = getByText(paragraphText);
+  expect(paragraphTag).toHaveStyle(`color: ${PrimaryColor.glintsred}`);
+});
+
+it('display the ellipsis', () => {
+  const { getByText } = render(<Paragraph ellipsis>{paragraphText}</Paragraph>);
+  const paragraphTag = getByText(paragraphText);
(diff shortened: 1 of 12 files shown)


franciscrispin:
Yo @rosslo, when you have the time could you help to code review this :pray:

src/General/Typography/Paragraph.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+describe('displays the correct text for each variant:', () => {
+  const variants: any[] = ['subtitle', 'regular', 'caption', 'smallest'];
+  variants.forEach(variant => {
+    it(`${variant}`, () => {
+      const { getByText } = render(
+        <Paragraph variant={variant}>{paragraphText}</Paragraph>
+      );
+      const paragraphTag = getByText(paragraphText);
+      expect(paragraphTag).toHaveTextContent(paragraphText);
```
rosslo: Because `getByText` matches full strings and case-sensitive by default,  `toHaveTextContent` seems to be redundant.
Can use `expect(paragraphTag).toBeTruthy()` or `expect(paragraphTag).toContainHTML('p')` to instead.

rosslo: I think we should add testing for `font-size` here

```
const fontSize: number[] = [18, 16, 14, 12];
...
const size = fontSize[index];
expect(paragraphTag).toHaveStyle(`font-size: ${size}px`);
```

src/General/Typography/Paragraph.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const Paragraph: React.FunctionComponent<Props> = props => {
+  const {
+    className,
+    children,
+    variant = 'regular',
+    bold,
+    color,
+    ellipsis,
+    ...defaultProps
```
rosslo: I think `restProps` makes more sense than `defaultProps`

src/General/Typography/Paragraph.tsx
```diff
@@ -0,0 +1,39 @@
...
+  ParagraphProps,
+} from './ParagraphStyles';
+
+const Paragraph: React.FunctionComponent<Props> = props => {
+  const {
+    className,
+    children,
+    variant = 'regular',
+    bold,
+    color,
```
rosslo: Could move default color from `Styles.ts` file to here for putting all default props together?

stories/index.tsx
```diff
@@ -61,6 +61,7 @@ import TextareaStory from './Input/TextareaStory';
 import TextFieldStory from './Input/TextFieldStory';
 import ToastStory from './Display/ToastStory';
 import TooltipStory from './Display/TooltipStory';
+import TypographyStory from '../src/General/Typography/TypographyStory';
```
rosslo: Just curious about the file path.
Seems to the other `Story.tsx` files are under `/storeies/General/`; 
like `import ProfilePictureStory from './General/ProfilePictureStory';`

src/General/Typography/Paragraph.tsx
```diff
@@ -0,0 +1,39 @@
...
+  } = props;
+
+  return (
+    <StyledParagraph
+      className={classNames('aries-typography-paragraph', className)}
+      variant={variant}
+      bold={bold}
+      color={color}
+      ellipsis={ellipsis}
+      {...defaultProps}
```
rosslo: same as above

src/General/Typography/Title.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+import Title from './Title';
+import { PrimaryColor } from '../../Style/Colors';
+
+import '@testing-library/jest-dom/extend-expect';
+import { render } from '@testing-library/react';
+
+const titleText = 'Glints Aries';
+
+describe('displays the correct text and uses the correct html tag for each tag:', () => {
+  const tags: any[] = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'];
```
rosslo: Can use the following to instead of `any[]`
``` 
 type tagsType = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
 const tags: tagsType[] = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'];
```

src/General/Typography/Paragraph.tsx
```diff
@@ -0,0 +1,39 @@
...
+    >
+      {children}
+    </StyledParagraph>
+  );
+};
+
+interface Props extends ParagraphProps {
+  className?: string;
+  children: React.ReactNode;
```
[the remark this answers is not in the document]
franciscrispin: Hmm it already extends the `ParagraphProps` which contains those props though.

stories/index.tsx
franciscrispin: Yeah i'm trying to slowly migrate all the `Story` files to be in the same directory as their component haha. Think it makes more sense for them to all be in the same directory.

franciscrispin:
@rosslo PTAL again :pray:

src/General/Typography/TypographyStory.tsx
```diff
@@ -0,0 +1,244 @@
...
+      require: 'no',
+      description: 'Display ellipsis when the title overflows.',
+    },
+    {
+      name: 'tag',
+      type: 'string',
+      defaultValue: <code>h1</code>,
+      possibleValue: (
+        <code>
+          {TITLE_VARIANTS.h1} | {TITLE_VARIANTS.h2} | {TITLE_VARIANTS.h3} |{' '}
```
rosslo: Is `{' '}` intended change?

src/General/Typography/TitleStyles.ts
```diff
@@ -0,0 +1,88 @@
...
+        return `
+        font-size: ${TITLE_FONT_SIZES.h5}px;
+        line-height: 1.4;
+      `;
+      case TITLE_VARIANTS.h6:
+        return `
+        font-size: ${TITLE_FONT_SIZES.h6}px;
+        line-height: 1.4;
+      `;
+      default:
```
rosslo: Maybe can set h3 - h6 as the default case to make this switch statement shorter

```
${props => {
    switch (props.as) {
      case TITLE_VARIANTS.h1:
        return `
        font-size: ${TITLE_FONT_SIZES.h1}px;
        font-weight: 900;
        line-height: 1.2;
      `;
      case TITLE_VARIANTS.h2:
        return `
        font-size: ${TITLE_FONT_SIZES.h2}px;
        line-height: 1.2;
      `;
      default:
        return `
        font-size: ${TITLE_FONT_SIZES[props.as]}px;
        line-height: 1.4;
      `;
    }
}}
```

src/General/Typography/TypographyStory.tsx
franciscrispin: Oh that's prettier auto formatting

src/General/Typography/TitleStyles.ts
franciscrispin: Hmm In this case I actually prefer to leave it so it's easier for the dev to see that all the variants are accounted for. Also it would be easier for future devs to add specific styling for each variant if necessary.

src/General/Typography/Title.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  tags.forEach(tag => {
+    const fontSize = TITLE_FONT_SIZES[tag];
+    it(`${tag}`, () => {
+      const { container, getByText } = render(
+        <Title tag={tag}>{titleText}</Title>
+      );
+      const heading = getByText(titleText);
+      expect(heading).toBeTruthy();
+      expect(container).toContainHTML(tag);
```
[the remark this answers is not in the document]
franciscrispin: https://github.com/testing-library/jest-dom#tocontainhtml

```
Assert whether a string representing a HTML element is contained in another element:
```

In this case, the `h1` tag should be contained inside the `container`. So I think it is more appropriate to say the container contains the html tag.

src/General/Typography/Paragraph.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  PARAGRAPH_FONT_SIZES,
+  PARAGRAPH_VARIANTS,
+  paragraphType,
+} from './ParagraphStyles';
+
+const paragraphText =
+  'As the first sign in the zodiac, the presence of Aries always marks the beginning of something energetic and turbulent. They are continuously looking for dynamic, speed and competition, always being the first in everything - from work to social gatherings.';
+
+describe('displays the correct text and font-size for each variant:', () => {
+  const variants: paragraphType[] = Object.values(PARAGRAPH_VARIANTS);
```
rosslo: `const variants = Object.values(PARAGRAPH_VARIANTS) as paragraphType[];`

rosslo: @franciscrispin do you have an error `Property 'values' does not exist on type 'ObjectConstructor'` for `Object.values`?
According to https://stackoverflow.com/questions/52304422/property-values-does-not-exist-on-type-objectconstructor, I think we can simply add `es2017.object` to `compilerOptions` `lib` array to resolve it.

franciscrispin: @rosslo Yeap that error is fixed in the `v4` branch already. I just need to merge it in and it should be fine.

rosslo:
@franciscrispin  LGTM 😄

PR #135, opened 2019-11-01
Feature/toggle switch

iamfeek:
**TESTS WILL BE DEFERRED**

_I ran out of time to properly settle the tests and am already pressed to have this feature out._

### DIFF
--- a/src/Input/ToggleSwitch/Styles.ts
+++ b/src/Input/ToggleSwitch/Styles.ts
@@ -1,0 +1,49 @@
+import styled from 'styled-components';
+import { SecondaryColor } from '../../Style/Colors';
+
+const DEFAULTS = {
+  activeBackgroundColor: SecondaryColor.actionblue,
+  inactiveBackgroundColor: SecondaryColor.grey,
+  inactiveBallColor: SecondaryColor.white,
+  activeBallColor: SecondaryColor.white,
+};
+
+interface ToggleProps {
+  active: Boolean;
+}
+
+const Toggle = styled.button<ToggleProps>`
+  cursor: pointer;
+  display: flex;
+  align-items: center;
+  overflow: hidden;
+  position: relative;
+  transform: translate3d(0, 0, 0);
+  background-color: ${({ active }) =>
+    active ? DEFAULTS.activeBackgroundColor : DEFAULTS.inactiveBackgroundColor};
+  height: 25px;
+  width: 56px;
+  border-radius: 14px;
+  padding: 0px;
+  border: none;
+`;
+
+const ToggleBall = styled.div<ToggleProps>`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  border-radius: 50%;
+  transition: transform 0.3s cubic-bezier(1, 0.19, 0.15, 0.7);
+  transition-delay: 0.1s;
+  background-color: ${DEFAULTS.inactiveBallColor};
+  height: 19px;
+  width: 19px;
+  margin: 3px;
+  transform: ${props =>
+    props.active ? `translateX(31px)` : 'translateX(0px)'};
+  &:active {
+    background-color: ${DEFAULTS.activeBallColor};
+  }
+`;
+
+export { ToggleBall, Toggle };
--- a/src/Input/ToggleSwitch/ToggleSwitch.tsx
+++ b/src/Input/ToggleSwitch/ToggleSwitch.tsx
@@ -1,0 +1,58 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+import { has } from 'lodash';
+
+import { SecondaryColor } from '../../Style/Colors';
+import Icon from '../../General/Icon';
+import { Toggle, ToggleBall } from './Styles';
+
+export interface ToggleSwitchProps {
+  defaultActive?: boolean;
+  iconOptions?: {
+    active: string;
+    inactive: string;
+  };
+  onChange?: (active: boolean) => void;
+  className?: string;
+}
+
+const ToggleSwitch: React.FunctionComponent<ToggleSwitchProps> = ({
+  defaultActive,
+  iconOptions,
+  onChange,
+  className,
+  ...defaultProps
+}) => {
+  const [active, setActive] = React.useState(defaultActive);
+
+  return (
+    <Toggle
+      aria-checked={active}
+      role="switch"
+      className={classNames('aries-toggle-switch', className)}
+      onClick={() => {
+        const newActiveState = !active;
+        setActive(newActiveState);
+        if (onChange) onChange(newActiveState);
+      }}
+      active={active}
+      {...defaultProps}
+    >
+      <ToggleBall active={active}>
+        {has(iconOptions, 'active') && has(iconOptions, 'inactive') && (
+          <Icon
+            color={active ? SecondaryColor.actionblue : SecondaryColor.grey}
+            name={active ? iconOptions.active : iconOptions.inactive}
+          />
+        )}
+      </ToggleBall>
+    </Toggle>
+  );
+};
+
+ToggleSwitch.defaultProps = {
+  defaultActive: false,
+};
+
+export default ToggleSwitch;
--- a/src/Input/ToggleSwitch/ToggleSwitchStory.tsx
+++ b/src/Input/ToggleSwitch/ToggleSwitchStory.tsx
@@ -1,0 +1,67 @@
+import * as React from 'react';
+
+import ToggleSwitch from './ToggleSwitch';
+import StorybookComponent from '../../../stories/StorybookComponent';
+import { SecondaryColor } from '../../Style/Colors';
+
+const propsObject = {
+  ToggleSwitch: [
+    {
+      name: 'defaultActive',
+      type: 'boolean',
+      defaultValue: <code>false</code>,
+      require: 'No',
+      possibleValue: <code>true | false</code>,
+      description:
+        'This will determine whether or not the component renders with an active or inactive state at the start.',
+    },
+    {
+      name: 'iconOptions',
+      type: 'object',
+      require: 'No',
+      possibleValue: (
+        <code>{'{ active: "eye", inactive: "eye-slashed" }'}</code>
+      ),
+      description:
+        'The name of icons that will be used in both active ' +
+        'and inactive states. Both icons have to be defined when opting to use ' +
+        'this.',
+    },
+    {
+      name: 'onChange',
+      type: 'function',
+      defaultValue: '',
+      require: 'No',
+      possibleValue: 'function',
+      description:
+        'Called when the state of the toggle has been changed. Returns true or false depending on whether the toggle is active or inactive.',
+    },
+  ],
+};
+
+const ToggleSwitchStory = () => (
+  <StorybookComponent
+    title="Toggle Switch"
+    code="import { ToggleSwitch } from 'glints-aries'"
+    propsObject={propsObject}
+    usage={`
+      <ToggleSwitch
+        defaultActive={false}
+        iconOptions={{
+          active: 'eye',
+          inactive: 'eye-slashed',
+        }}
+      />
+    `}
+  >
+    <ToggleSwitch
+      defaultActive={false}
+      iconOptions={{
+        active: 'eye',
+        inactive: 'eye-slashed',
+      }}
+    />
+  </StorybookComponent>
+);
+
+export default ToggleSwitchStory;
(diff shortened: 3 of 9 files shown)


src/Input/ToggleSwitch/ToggleSwitch.story.tsx
```diff
@@ -0,0 +1,81 @@
...
+import ToggleSwitch from './ToggleSwitch';
+import { ToggleSwitchProps } from './ToggleSwitch.interface';
+import StorybookComponent from '../../../stories/StorybookComponent';
+import { SecondaryColor } from '../..';
+
+const onClick = (e: any) => {}
+
+const props: ToggleSwitchProps = {
+  defaultActive: false,
+  activeBallColor: SecondaryColor.white,
```
franciscrispin: Hmm we typically don't provide styling options unless the user is likely to want to change them. ([Point 4](https://github.com/glints-dev/glints-aries#code-procedure)). 

Another approach is to check with the designer whether they have some standard `theme` we can apply (like how we do it for the Button https://github.com/glints-dev/glints-aries/blob/v4/src/General/Button/Button.tsx#L14)

This comment applies to these props:
```
activeBackgroundColor,
inactiveBackgroundColor,
activeBallColor,
inactiveBallColor,
```

src/Input/ToggleSwitch/ToggleSwitch.story.tsx
```diff
@@ -0,0 +1,81 @@
...
+  inactiveBallColor: SecondaryColor.white,
+  activeBackgroundColor: SecondaryColor.actionblue,
+  inactiveBackgroundColor: SecondaryColor.grey,
+  activeIconName: 'eye',
```
franciscrispin: Since `activeIconName` and `inactiveIconName` are closely related, consider putting them together as an object instead. E.g.:

```
iconOptions: {
    active: 'eye',
    inactive: 'eye-slashed',
  },
```

src/Input/ToggleSwitch/ToggleSwitch.story.tsx
```diff
@@ -0,0 +1,81 @@
...
+  inactiveIconName: 'eye-slashed',
+  handleClick: onClick,
+};
+
+const propsObject = {
```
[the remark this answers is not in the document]
franciscrispin: Also for the `iconOptions` prop, we should mention that while it is optional, both are needed if a dev decides to include them (meaning you can't have `activeIconName` but leave out `inactiveIconName`). 

I'm not sure if there's a use-case for a dev setting either one without setting the other. I don't have strong opinions on that so you can decide which is better.

src/Input/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -0,0 +1,55 @@
...
+  activeBackgroundColor,
+  inactiveBackgroundColor,
+  activeBallColor,
+  inactiveBallColor,
+  onChange,
+}) => {
+  const [active, setActive] = React.useState(defaultActive || false);
+
+  return (
+    <ToggleSwitchWrapper>
```
franciscrispin: Is this wrapper necessary? Doesn't seem to make a difference if it is removed.

src/Input/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+const ToggleSwitch: React.FunctionComponent<ToggleSwitchProps> = ({
+  defaultActive,
+  activeIconName,
+  inactiveIconName,
+  activeBackgroundColor,
+  inactiveBackgroundColor,
+  activeBallColor,
+  inactiveBallColor,
+  onChange,
```
franciscrispin: Pass any remaining props as `...defaultProps` and pass that to the `Toggle`  component.

```
<Toggle 
  otherProps
  {...defaultProps}
/>
```

This allows the dev to add other props such as `style`.

src/Input/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -0,0 +1,55 @@
...
+}) => {
+  const [active, setActive] = React.useState(defaultActive || false);
+
+  return (
+    <ToggleSwitchWrapper>
+      <Toggle
```
franciscrispin: Set a default className `aries-toggle-switch` and pass the className prop to `<Toggle>`. We usually set this on the root element. In this case your root element is the `<ToggleSwitchWrapper>` but that seems like it can be removed so you can pass it to `<Toggle>` instead.

Example: 
```
className={classNames('aries-toggle-switch', className)}
```
Reference: https://github.com/glints-dev/glints-aries/blob/v4/src/Input/Switch/Switch.tsx#L40

src/Input/ToggleSwitch/Styles.ts
```diff
@@ -0,0 +1,47 @@
+import styled from 'styled-components';
+import { ToggleBallProps, ToggleProps } from './ToggleSwitch.interface';
+
+const ToggleSwitchWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  position: relative;
+`;
+
+const Toggle = styled.div<ToggleProps>`
```
franciscrispin: In terms of html semantics, it seems the conventional element for a Toggle component is a `button` instead of a `div`. You will have to add some css to override the default button styles but I think that's an acceptable trade-off for having semantically accurate html. 

One example I found is this Toggle component found here https://scottaohara.github.io/aria-switch-control/
image

src/Input/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -0,0 +1,55 @@
...
+        onClick={() => {
+          const newActiveState = !active;
+          setActive(newActiveState);
+          if (onChange) onChange(newActiveState);
+        }}
+        active={active}
+        activeBackgroundColor={activeBackgroundColor}
+        inactiveBackgroundColor={inactiveBackgroundColor}
+      >
```
franciscrispin: Accessibility is a key feature for component libraries. I'm not super familiar with it as well, but what I usually do is to copy the `aria` attributes from similar components. I suggest copying the them by inspecting this component found here https://scottaohara.github.io/aria-switch-control/

image
image


Seems like you need the following attribtues:
```
role="switch"
aria-checked="true" | "false"
```

The `data` attributes are unnecessary.

src/Input/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -0,0 +1,55 @@
...
+  defaultActive,
+  activeIconName,
+  inactiveIconName,
+  activeBackgroundColor,
+  inactiveBackgroundColor,
+  activeBallColor,
+  inactiveBallColor,
+  onChange,
+}) => {
+  const [active, setActive] = React.useState(defaultActive || false);
```
franciscrispin: You don't need `|| false` since it is already set in the defaultProps.

src/Input/ToggleSwitch/Styles.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const ToggleBall = styled.div<ToggleBallProps>`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  z-index: 2;
+  border-radius: 50%;
+  transition: transform 0.3s cubic-bezier(1, 0.19, 0.15, 0.7);
+  transition-delay: 0.1s;
+  will-change: transform;
```
franciscrispin: Hmm I didn't know about this but found this out on the docs https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

```
Important: will-change is intended to be used as a last resort, in order to try to deal with existing performance problems. It should not be used to anticipate performance problems.
```

I don't foresee any performance issues so I suggest removing it.

src/Input/ToggleSwitch/Styles.ts
```diff
@@ -0,0 +1,47 @@
...
+  height: 25px;
+  width: 56px;
+  border-radius: 14px;
+`;
+
+const ToggleBall = styled.div<ToggleBallProps>`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  z-index: 2;
```
franciscrispin: This seems like it can be safely removed since `z-index` can't work without setting the `position` of the element

src/Input/ToggleSwitch/ToggleSwitch.interface.ts
```diff
@@ -0,0 +1,25 @@
+interface ToggleSwitchProps {
```
franciscrispin: This is purely a matter of convention/personal preference. In aries, it seems that most interfaces are defined in the component that needs them. In this case `ToggleSwitchProps` belongs only to `ToggleSwitch.tsx` while the other two belong to `Styles.ts` so I think it more appropriate to just define them in their own files rather than a separate file.

It might be the case that we eventually decide that all interfaces should reside in their own file, but for now I think we can follow the convention.

src/Input/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -0,0 +1,55 @@
+import * as React from 'react';
+
+import { Icon, SecondaryColor } from '../../index';
+import { Toggle, ToggleBall, ToggleSwitchWrapper, ToggleSwitchProps } from './';
+
+const ToggleSwitch: React.FunctionComponent<ToggleSwitchProps> = ({
+  defaultActive,
```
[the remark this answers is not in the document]
franciscrispin: Okay this is subjective, I'll leave it up to you haha

franciscrispin:
Oh @iamfeek I'm not sure why some of my comments aren't showing on this page. You might have to go to the files changed here https://github.com/glints-dev/glints-aries/pull/135/files to see the comments.

src/Input/ToggleSwitch/ToggleSwitch.story.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from 'react';
```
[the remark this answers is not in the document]
iamfeek: I dont think they are any clear advantages to the naming convention. 
IIRC, I saw that convention somewhere and it got registered in my head as the defacto way of handling naming in React. 🤦‍♂ 

I will convert the files to be consistent with the current codebase

src/Input/ToggleSwitch/ToggleSwitch.tsx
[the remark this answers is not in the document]
iamfeek: I will leave it!

iamfeek:
@franciscrispin PTALA!

src/Input/ToggleSwitch/ToggleSwitch.story.tsx
iamfeek: Good insight!

I think for this, I will just default it to what I currently have from my designer, then I will iterate when the need arise.

src/Input/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -1,68 +1,55 @@
...
+      onClick={() => {
+        const newActiveState = !active;
+        setActive(newActiveState);
+        if (onChange) onChange(newActiveState);
+      }}
+      active={active}
+      {...defaultProps}
+    >
+      <ToggleBall active={active}>
+        {iconOptions && (
```
franciscrispin: This will error if either `active` or `inactive` properties are not present, or an empty object is passed. Can use lodash `has` or `hasOwnProperty` to check for both properties.

src/Input/ToggleSwitch/ToggleSwitchStory.tsx
```diff
@@ -65,20 +16,13 @@ const propsObject = {
...
+      name: 'iconOptions',
+      type: 'object',
+      require: 'No',
+      possibleValue: `{ active: "eye", inactive: "eye-slashed" }`,
+      description: 'The name of icons that will be used in both active ' +
+        'and inactive states. Both icons have to be defined when opting to use ' +
+        'this.',
     },
     {
       name: 'onChange',
```
franciscrispin: Can just add one sentence to the `onChange` description
```
description: 'Called when the state of the toggle has been changed. Returns true or false depending on whether the toggle is active or inactive.',
```

src/Input/ToggleSwitch/ToggleSwitchStory.tsx
```diff
@@ -65,20 +16,13 @@ const propsObject = {
...
-      name: 'inactiveIconName',
-      type: 'string',
-      require: 'Yes if activeIconName is provided',
-      possibleValue: 'All icon names from the Icon component',
-      description:
-        'The icon that will be displayed when the toggle is in the inactive state.',
+      name: 'iconOptions',
+      type: 'object',
+      require: 'No',
+      possibleValue: `{ active: "eye", inactive: "eye-slashed" }`,
```
franciscrispin: Wrap with `<code>`:

```
possibleValue: (
        <code>{'{ active: 'eye', inactive: 'eye-slashed' }'}</code>
      ),
```

src/Input/ToggleSwitch/Styles.ts
```diff
@@ -1,56 +1,49 @@
 import styled from 'styled-components';
+import { SecondaryColor } from '../../Style/Colors';
```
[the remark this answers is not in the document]
iamfeek: This is used

PR #145, opened 2019-11-08
Checkbox: Refactor to use hooks and add unit and snapshot tests

franciscrispin:
- Refactor to use hooks 
- Add unit and snapshot tests

Note: I wanted to test the `:before` and `:after` pseudo-elements to make sure that the checkbox exists. However I couldn't find a good way to do it because the `renderer` does not tell us anything about pseudo-elements. 

https://stackoverflow.com/questions/54950000/test-for-pseudo-class-using-jest-and-react-test-renderer

I tried using `getComputedStyle` but that didn't seem to work for the same reason https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle

### DIFF
--- a/src/Input/Checkbox/Checkbox.test.tsx
+++ b/src/Input/Checkbox/Checkbox.test.tsx
@@ -1,0 +1,73 @@
+import * as React from 'react';
+
+import * as renderer from 'react-test-renderer';
+import '@testing-library/jest-dom/extend-expect';
+import { fireEvent, render } from '@testing-library/react';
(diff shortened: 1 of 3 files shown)


franciscrispin:
@rosslo PTAL. Not urgent, just whenever you are free :smile: 

@iamfeek can take a look if you're interested too! But I think ross will handle the review since you have done the past few already.

rosslo:
@franciscrispin LGTM!

PR #149, opened 2019-11-11
add launch.json

chong901:


franciscrispin:
Hi @rosslo @tingtingdisco @westwood846, @aaa59891 set this up to allow us to debug code more easily but I couldn't get it to work locally on my machine for reasons neither of us know haha. If you think it's worth keeping, you can try to fix it before it goes stale :smile:

westwood846:
Seems like this has gone stale after all :D Closing, if anybody needs this can make new PR, doesn't seem too complicated.

Issue #148, opened 2019-11-11
Color: Update color library


franciscrispin:
- Update `Primary` and `Secondary` colors. Some colors have been removed. Need to check if they are used in our existing codebases and refactor them accordingly. In the interim, we should show deprecation warnings without completely removing the colors.
- Add `Greyscale` colors 

Glints UI: https://app.zeplin.io/project/5ba343bf0c399c33db483653/screen/5bbf16daf616cc09a6264519

PR #150, opened 2019-11-12
Colors: Update color library

franciscrispin:
https://github.com/glints-dev/glints-aries/issues/148

- Remove some colors from SecondaryColor
- Add colors to Grayscale

Glints UI: https://app.zeplin.io/project/5ba343bf0c399c33db483653/screen/5bbf16daf616cc09a6264519

### DIFF
--- a/src/Style/Colors.ts
+++ b/src/Style/Colors.ts
@@ -1,31 +1,47 @@
 export const PrimaryColor = {
+  glintsyellow: '#FFF240',
   glintsred: '#EC272B',
   glintsblue: '#0BAEEC',
-  glintsyellow: '#FFF240',
 };
 
 export const SecondaryColor = {
-  white: '#FFFFFF',
-  whitesmoke: '#F3F3F3',
+  actionblue: '#017EB7',
+  green: '#93BD49',
+  darkgreen: '#008464',
+  magenta: '#FF00E5',
+  violet: '#9013FE',
+  orange: '#F5A623',
+
+  // deprecated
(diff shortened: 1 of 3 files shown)


franciscrispin:
@aaa59891 PTAL :pray: I decided not to show the deprecation messages because 

1. I can just refactor all the instances of the deprecated colors in our codebases
2. When I showed the deprecation message using the `IIFE`, the moment you import the color, all the deprecation messages will show at the same time which is not very helpful

src/Style/Colors.ts
```diff
@@ -1,31 +1,47 @@
...
   lightgreen: '#EEF6F4',
-  darkgreen: '#008464',
-  orange: '#F5A623',
   lightorange: '#FFF6E8',
   lightred: '#FFF2F3',
   lightblue: '#D6F4FF',
   darkblue: '#004b68',
+
+  // renamed
+  purplepizza: '#FF00E5',
```
aaa59891: oh there two are renamed, but it seems they are duplicated, and in the storybook, I only see `magenta` and `violet`, which two do you want to be used?

franciscrispin: We will use this:
```
magenta: '#FF00E5',
violet: '#9013FE',
```

I'm leaving `purplepizza` and `bluegem` there until I refactor all the usage of them in our codebases.

aaa59891: ok!

aaa59891:
LGTM

PR #151, opened 2019-11-13
Optimize icon size

rosslo:
# Why
Currently, `glints-aries` is using a string name to map the icon, which can not load on demand.

# What changed
- Add script `build:icon` to automatically generate individual icon component files and export them from `src/index.ts`.
- Change the Dropdown's prop `leftIconName: string` to `leftIcon: element` 
- Show warning message for legacy icon usage way
image

- Change Icon usage way from

```
Import { Icon } from 'glints-aries';
...
<Icon name="add" />
```
to
```
Import { AddIcon } from 'glints-aries';
...
<AddIcon />
```

# Result for `dst` bundle
Before:
image

After:
image


franciscrispin:
@rosslo for the time being, I think we need to maintain the old IconLibrary until we refactor all the usages of Icons in all our codebases. So for now we can just leave the old `IconLIbrary` as is and push the changes you made. 

Can we also show some deprecation warning when a user uses `<Icon name="iconName" />`?

PR #152, opened 2019-11-13
Alert: Fix the closing of the alert on clicking below the close icon

ritikamotwani:
Currently for the Alert component clicking on the area below the close icon results in closing of the alert box. This is caused by the `div` wrapping the icon taking up the height of its parent.

image

Fixes (only for the desktop) : #158

### DIFF
--- a/src/Style/General/AlertStyle.ts
+++ b/src/Style/General/AlertStyle.ts
@@ -43,7 +43,7 @@ export const AlertContainer = styled.div<AlertContainerProps>`
   font-size: 1em;
 
   @media ${Device.mobileM} {
-    min-width: 100%;
+    min-width: calc(100% - 30px);
   }
 
   ${({ type }) => {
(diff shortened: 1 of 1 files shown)


franciscrispin:
@ritikamotwani could you create an issue for this too? Thanks!

tingtingdisco:
@ritikamotwani could you also update the style that 
1. there should be space between the mobile screen and the banner
2. the text on the banner is too closed to the close button

https://docs.google.com/document/d/1s8XQm-09AGqmtsy5UY9hzzfPo1cJDEO4xIkXVoisKB4/edit

src/Style/General/AlertStyle.ts
```diff
@@ -43,7 +43,7 @@ export const AlertContainer = styled.div<AlertContainerProps>`
   font-size: 1em;
 
   @media ${Device.mobileM} {
-    min-width: 100%;
+    min-width: 95%;
```
tingtingdisco: @ritikamotwani do you think this is better? That we keep the spacing 15px on both side.
```
min-width: calc(100% - 30px);
```

ritikamotwani: Thanks @tingtingdisco 
Updated this :smile:

tingtingdisco:
LGTM, thank u @ritikamotwani

Issue #91, opened 2019-08-17
<Modal> closes on mousedown inside modal and mouseup outside modal


franciscrispin:
Reported by: @yjwong 
[Tello Link](https://trello.com/c/gBBOp3Wo/114-modal-gets-dismissed-when-releasing-mouse-over-the-darkened-area-reported-by-yong-jie)

Users may accidentally close the modal while trying to highlight text inside the modal.

PR #153, opened 2019-11-13
Modal: Do not close modal if mousedown occurs inside modal but the subsequent mouseup occurs outside it

franciscrispin:
Fix for https://github.com/glints-dev/glints-aries/issues/91

Previously, the modal would close if a `mousedown` event occurred inside the modal, but the `mouseup` event occurred outside. This causd the modal to close when a user clicks inside the modal to copy some text, but releases the click outside the modal.

We fixed this by checking if the`mousedown` event occurred inside the modal, and ignoring the subsequent `click` event.

### DIFF
--- a/src/Display/Modal/Modal.tsx
+++ b/src/Display/Modal/Modal.tsx
@@ -1,9 +1,7 @@
 import * as React from 'react';
-
 import classNames from 'classnames';
 
 import Icon from '../../General/Icon';
-
 import { escEvent } from '../../Utils/DomUtils';
 
 import {
@@ -17,6 +15,7 @@ import {
 
 class Modal extends React.Component<Props, State> {
   modalContentAreaRef: React.RefObject<HTMLDivElement>;
+  mouseDownTarget: HTMLElement;
 
   state = {
     isOpen: false,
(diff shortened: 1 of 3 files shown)


src/Display/Modal/Modal.tsx
```diff
@@ -58,6 +57,26 @@ class Modal extends React.Component<Props, State> {
...
+    }
+  };
+
+  handleClick = () => {
+    if (this.state.ignoreBackdropClick) {
+      this.setState({ ignoreBackdropClick: false });
+      return;
+    }
+    this.props.onClose();
+  };
```
rosslo: @franciscrispin  Could we use below to instead `ignoreBackdropClick` simply?
```
  handleMouseDown = (e: React.MouseEvent) => {
    const element = event.target as HTMLElement;
    this.mouseDownTarget = element;
  };

  handleClick = (e: React.MouseEvent) => {
    const element = event.target as HTMLElement;

    if (this.mouseDownTarget === element) {
      this.props.onClose();
    }
  };
```

franciscrispin: Oh yes this is great! Thanks @rosslo. PTAL again :smile:

PR #155, opened 2019-11-14
Test/textarea

franciscrispin:
- Textarea: Add unit and snapshot test
- TextField: Minor updates to tests

### DIFF
--- a/src/Input/Textarea/Textarea.test.tsx
+++ b/src/Input/Textarea/Textarea.test.tsx
@@ -1,0 +1,117 @@
+import * as React from 'react';
+
+import * as renderer from 'react-test-renderer';
+import '@testing-library/jest-dom/extend-expect';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import Textarea from './Textarea';
+import { SecondaryColor, PrimaryColor } from '../../Style/Colors';
+
+interface Props {
+  label: string;
+  onChange: jest.Mock;
+}
+
+const props: Props = {
+  label: 'Description',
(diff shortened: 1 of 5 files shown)


src/Input/Textarea/Textarea.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const TextareaSnapshot = renderer
+    .create(<Textarea label={props.label} />, options)
+    .toJSON();
+  expect(TextareaSnapshot).toMatchSnapshot();
+});
+
+describe('when it is rendered', () => {
+  it('should display the correct input label', () => {
+    const { textareaInput } = setupTextarea();
+    expect(textareaInput).toBeTruthy();
```
aaa59891: Hi @franciscrispin do you follow any practice for this kind of test?

look at this test case, I don't know which input is correct in this textarea because the value is in setupTextarea function
and this value should never return false value right 🤔 ?

could we have something like:
1. define arguments
2. get the component based on the arguments
3. check the component fields

franciscrispin: Oh in this case `const textareaInput = getByLabelText(props.label) as HTMLTextAreaElement;` will error if the `value` is not found so `expect(textareaInput).toBeTruthy();` will be `falsy` and the test will fail. 

I don't really have a testing practice like you described haha. Can you share an example of the 3 steps you described? @aaa59891

franciscrispin: Okay thank you! Will look into it!

aaa59891:
LGTM

PR #157, opened 2019-11-14
Tag: Add unit and snapshot tests

franciscrispin:


src/General/Tag/Tag.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+import * as renderer from 'react-test-renderer';
+import '@testing-library/jest-dom/extend-expect';
+import { render } from '@testing-library/react';
+
+import Tag from './Tag';
+
+const children = 'Glints Aries';
+
+it(`<Tag> should render a label with the text "Glints Aries"`, () => {
```
no13000:
```suggestion
it(`should render a label with the text ${children}`, () => {
```

franciscrispin: Oh the `<ComponentName>` is intentional cause I wanted to differentiate my snapshot test from the other test so I'm using it as a sort of convention haha 

E.g. https://github.com/glints-dev/glints-aries/blob/v4/src/Input/TextField/TextField.test.tsx#L38

PR #163, opened 2019-11-25
Add icon buttons

### LINKED ISSUE #122: Add IconButton component
We manually maintain different `IconButton` components in our existing codebases. There is a need to implement one on aries that will cover developers' common use-cases. Will have to consult our designers for this as well.

image
image

ritikamotwani:
### Part 1
Adding a button with only an icon. (Shapes -  default, square)
Usage: 
```
<Button
  variant="icon"
>
  <CloseIcon />
</Button>
```
Square button

```
<Button
  variant="icon"
  shape="square"
>
  <CloseIcon />
</Button>
```
image

### Part 2
Adding icon with the current variant buttons.

Usage:
```
<Button variant="primary"> // the variant shouldn't matter for this to work
  GO TO PROFILE <RightArrowIcon position="right" />
</Button>
```
image

Fixes: #122

### DIFF
--- a/src/General/Button/Button.tsx
+++ b/src/General/Button/Button.tsx
@@ -7,7 +7,7 @@ import GhostButton from './GhostButton';
 import LinkButton from './LinkButton';
 import IconButton from './IconButton';
 
-import { Variant } from '../../Utils/StyleConfig';
+import { Position, Variant } from '../../Utils/StyleConfig';
 
 const renderButton: React.FunctionComponent<Props> = ({
   children,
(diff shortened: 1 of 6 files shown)


ritikamotwani:
Hi @franciscrispin 
I had a doubt in the icon added to the current buttons.
With the current mentioned usage the position of the icon is specified in the `children` so the code has become very messy as I have to iterate and amend the children being rendered. 
I couldn't think of any better idea :thinking: 
Is there something better that exists to do this or should we change the usage to something like:
```
<Button variant="primary" iconPosition="right">
```
so that the position is extracted from the button property only?

franciscrispin:
@ritikamotwani ahh good point. Hmm I think we can do away with the `position` property altogether actually to simplify things. We can just use css to determine where to add the margin. Can refer to the antd buttons as an example :smile: 

image

Issue #108, opened 2019-09-02
<Tooltip> component not working on mobile


franciscrispin:
Currently we use `onTouchEnd` and `onTouchStart` event handlers as a quick fix to make the tooltip component usable on mobile. https://github.com/glints-dev/glints-aries/blob/v4/src/Display/Tooltip/Tooltip.tsx#L31

However this means the user has to click and hold on to the tooltip for it to appear which is porr UX. We need to implement something similar to the `hover` tooltip used by [material UI](https://material-ui.com/components/tooltips/#triggers)

Expected behaviour:
Desktop: Open on mouseover, Close on mouseleave
Mobile: Open on click, Close on click outside

Current behaviour: 
Desktop: Open on mouseover, Close on mouseleave :heavy_check_mark: 
Mobile: Open on touch, Closes on touch leave :x: 

Comments: 
- The `onMouseLeave` event is not supported by iOS https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/HandlingEvents/HandlingEvents.html
- Another potential issue is on Safari because it handles click events differently https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event#Safari_Mobile

franciscrispin:
Fixed in https://github.com/glints-dev/glints-aries/pull/117

franciscrispin:
The tootip still cannot be closed on iOS for versions < 13.1.3

franciscrispin:
@rosslo will help with this :smile:

rosslo:
Fixed in https://github.com/glints-dev/glints-aries/pull/164

PR #165, opened 2019-11-27
Fix: should close tooltip when click trigger button

rosslo:
Relevant Issue: https://github.com/glints-dev/glints-aries/issues/108

### DIFF
--- a/src/Display/Tooltip/Tooltip.tsx
+++ b/src/Display/Tooltip/Tooltip.tsx
@@ -20,29 +20,26 @@ const Tooltip: React.FunctionComponent<Props> = ({
   const showTooltip = () => setIsShow(true);
   const hideTooltip = () => setIsShow(false);
 
-  const hideTooltipIfTouchOutside = (element: HTMLElement) => {
-    const hasTouchedOutsideOfTooltipContent = !contentRef.current.contains(
-      element
-    );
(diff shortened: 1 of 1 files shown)


src/Display/Tooltip/Tooltip.tsx
```diff
@@ -15,30 +15,38 @@ const Tooltip: React.FunctionComponent<Props> = ({
...
+  const [isShow, setIsShow] = React.useState(false);
+  const showTooltip = () => setIsShow(true);
+  const hideTooltip = () => setIsShow(false);
 
-  const touchOutside = (event: TouchEvent) => {
-    const element = event.target as HTMLElement;
-    const hasTouchedOutside = !ref.current.contains(element);
+  const hideTooltipIfTouchOutside = React.useCallback(
+    (event: TouchEvent | React.TouchEvent) => {
+      const hasTouchedOutsideOfTooltipContent = !contentRef.current.contains(
```
aaa59891: in order to prevent null pointer, suggest to add check for contentRef.current

rosslo: Since `hideTooltipIfTouchOutside` muse be triggered after `showTooltip`, don't think that check is necessary.

aaa59891:
LGTM

Issue #166, opened 2019-11-28
Make pop over component reusable for filters. (Step 1: Add a number badge)


ritikamotwani:
Currently the filters have a number badge showing the number of options selected from the drop down.
![image](https://user-images.githubusercontent.com/22127980/69780412-3566fb80-11d1-11ea-9952-d79a8f23f868.png)
Add a property to the `Pop over` called showNumberBadge with the number to be shown beside the pop over label.

PR #167, opened 2019-11-28
Popover: Add a property to show the number of options selected as a badge

ritikamotwani:
Fixes: #166 
Add a property to the pop over called showBadgeNumber which if greater than 0 shows a badge with the number specified.

### DIFF
--- a/src/Display/Popover/Popover.tsx
+++ b/src/Display/Popover/Popover.tsx
@@ -12,6 +12,7 @@ import {
   PopOverContent,
   PopoverItemWrapper,
 } from '../../Style/Display/PopoverStyle';
+import Badge from '../../General/Badge';
 
 class Popover extends React.Component<Props, State> {
   constructor(props: Props) {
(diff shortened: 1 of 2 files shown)


franciscrispin:
Hi @ritikamotwani ,

@aaa59891 and I had some discussions about this.

They key question is whether this behaviour to show the badge should belong in aries or should be customized by the developer in the codebase.

To help us think about it, we categorized components into 3 groups. (Note `codebase` here refers to `dst`/`employers`)
1. cross-project component - should exist in glints-aries
2. global shared component - should exist in the shared components folder in the codebase 
3. feature component - should exists in the module in the codebase

I'm not sure whether this behaviour is required in `employers`. If it is, we can add it in aries. If it isn't, then this should be done on `dst` itself by extending the aries `Popover` component.

ritikamotwani:
@franciscrispin I guess then this behaviour shouldn't be added in aries.
The reason being a Popover component in any of the public UI libraries rarely has this badge thing. This can be included in the shared component of the `dst` repo, as it would make more sense there, and as the filter being used in `dst` is similar everywhere, if some dev finds more things to be changed for the Popover to reused as a filter, they can simply refactor the shared component so that it's similar in the project.

franciscrispin:
@ritikamotwani Yeap agree with your assessment! Feel free to close the issue if there are no more issues :smile:

ritikamotwani:
Closing the PR and issue in respect to the discussion here: https://github.com/glints-dev/glints-aries/pull/167#issuecomment-564005868

PR #172, opened 2019-12-09
Gallery: The number of images displayed at once to be dynamic

ritikamotwani:
For the number of images shown at once while using the Gallery component, the user should be able to pass that as a value. And the default one is set to 8 (as per the current implementation).

When the number is set to 3:
image

### DIFF
--- a/src/Application/Gallery/Gallery.tsx
+++ b/src/Application/Gallery/Gallery.tsx
@@ -13,6 +13,8 @@ import {
   GalleryThumbnailWrapper,
 } from '../../Style/Application/GalleryStyle';
 
+const defaultImagesDisplayed = 8;
+
 class Gallery extends React.Component<Props, State> {
   sliderRef: React.RefObject<Slider>;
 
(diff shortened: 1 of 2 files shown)


stories/Application/GalleryStory.tsx
```diff
@@ -4,10 +4,25 @@ import StorybookComponent from '../StorybookComponent';
 
 import Gallery from '../../src/Application/Gallery';
 
+const props = {
+  Gallery: [
+    {
+      name: 'imagesDisplayed',
+      type: 'number',
+      defaultValue: '8',
+      possibleValue: 'integer',
```
franciscrispin: I think we can change it to `number` since there is no `integer` type in js

src/Application/Gallery/Gallery.tsx
```diff
@@ -54,9 +54,12 @@ class Gallery extends React.Component<Props, State> {
 
   componentDidMount() {
     const { children } = this.props;
-
-    if (React.Children.count(children) > 8) {
-      this.setState({ imageLeft: React.Children.count(children) - 8 });
+    let { imagesDisplayed } = this.props;
```
franciscrispin: Suggestion:
Replace the following:
```
let { imagesDisplayed } = this.props;
    if (!imagesDisplayed) {
      this.setState({ imageLeft: React.Children.count(children) - 8 });	      imagesDisplayed = 8;
    }
```
with: 
```
const { imagesDisplayed = 8 } = this.props;
```

src/Application/Gallery/Gallery.tsx
```diff
@@ -72,13 +75,17 @@ class Gallery extends React.Component<Props, State> {
 
   render() {
     const { children } = this.props;
+    let { imagesDisplayed } = this.props;
```
franciscrispin: Same comment as above

ritikamotwani:
Thanks @franciscrispin !
Updated the PR.

PR #173, opened 2019-12-10
Hotfix vn poppins font

franciscrispin:
Add the SVN Poppins font to allow Vietnamese characters to display correctly.

- Add `SVN Poppins` font to `GlobalStyles`
- Add the `unicode-range` property for `SVN Poppins` to tell the browser to load the font only when the characters fall within the range of vn characters
- Update `font-family` in `GlintsContainer`

Note: The unicode-range was taken from http://vietunicode.sourceforge.net/charset/


src/Style/GlobalStyle.css
```diff
@@ -1,49 +1,138 @@
...
+
 @font-face {
-    font-family: 'Poppins';
-    src: url('../../assets/fonts/Poppins-Light.ttf');
-    font-weight: 300;
-    font-style: normal;
-    font-stretch: normal;
-    line-height: normal;
-    letter-spacing: normal;
+  font-family: 'SVN Poppins';
```
yjwong: Just wondering, does this work if you use the same `font-family` name? If it does, then it might be the case that you don't have to specify multiple font families within `GlintsContainerStyle`.

franciscrispin: @yjwong Yeap it works! Just had to declare the original poppins font-face after declaring the SVN Poppins font-face so the browser will load the original font by default.

yjwong: Awesome!

Issue #183, opened 2020-01-11
Accordion: Fix UI mismatches 


franciscrispin:
Glints UI: https://app.zeplin.io/project/5ba343bf0c399c33db483653/screen/5ba358620049d012fcb8b234 
Glints Aries Status: https://docs.google.com/spreadsheets/d/1O9WY_x9MmzRIbskPE-2_TsKgOz3N7C61iEqcx6zWln8/

1. Identify the UI Mismatches
2. Fix the UI Mismatches
3. Implement Unit Tests for the component

- In addition to the design on Glints UI, the icon position should be customizable to be on the left or right

franciscrispin:
Minor UI mismatches remaining: 
- animate the change from plus to minus
- start the divider only after after the icon

westwood846:
Closing, can make new issue when designs are available.

PR #184, opened 2020-01-14
Feature/refactor accordion

franciscrispin:
https://github.com/glints-dev/glints-aries/issues/183

- Add `iconPosition` prop to display the icon on the `right` and `left`. According to our designers, the icons used will be fixed as they do not need to be customizable. 
- Currently, the minus icon jumps when it switches from the plus icon. This will be fixed when the minus icon is replaced with a vertically centered minus icon. 

Note: DO NOT MERGE until the Accordion has been tested to be backwards compatible.

### DIFF
--- a/src/Style/Display/AccordionStyle.ts
+++ b/src/Style/Display/AccordionStyle.ts
@@ -1,59 +1,103 @@
-import styled from 'styled-components';
-import { SecondaryColor } from '../Colors';
+import styled, { css } from 'styled-components';
 
-export const AccordionContainer = styled.div`
-  position: relative;
+import { SecondaryColor, Greyscale } from '../Colors';
+import { IconPosition } from '../../Display/Accordion/Accordion';
+
+const iconSize = 15;
+const iconMargin = 30;
+const offsetIcon = iconSize + iconMargin;
+const transitionTiming = '0.24s';
+
+export const PanelWrapper = styled.div`
+  border-top: solid 1px ${Greyscale.grey};
+  outline: none;
 `;
 
-export const AccordionContentWrapper = styled.div`
+export const Container = styled.div`
+  ${PanelWrapper}:first-child {
+    border-top: none;
+  }
+`;
+
+export const IconLabelWrapper = styled.div<IconLabelWrapper>`
   position: relative;
-  display: grid;
-  grid-template-columns: max-content 1fr;
-  grid-gap: 0.6em;
-  font-size: 1em;
-  line-height: 1.5;
-  padding: 1.5em 0;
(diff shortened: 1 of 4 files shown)


src/Style/Display/AccordionStyle.ts
```diff
@@ -1,59 +1,106 @@
...
+import { SecondaryColor, Greyscale } from '../Colors';
+import { IconPosition } from '../../Display/Accordion/Accordion';
+
+const iconSize = 15;
+const iconMargin = 30;
+const offsetIcon = iconSize + iconMargin;
+const transitionTiming = '0.24s';
+
+export const PanelWrapper = styled.div`
   position: relative;
```
rosslo: `position: relative;`  should be unused now.

src/Style/Display/AccordionStyle.ts
```diff
@@ -1,59 +1,106 @@
...
+  border-top: solid 1px ${Greyscale.grey};
+  outline: none;
 `;
 
-export const AccordionContentWrapper = styled.div`
+export const Container = styled.div`
   position: relative;
```
rosslo: same as above

src/Style/Display/AccordionStyle.ts
```diff
@@ -1,59 +1,106 @@
...
+    border-top: none;
+  }
+`;
+
+export const IconLabelWrapper = styled.div<IconLabelWrapper>`
+  position: relative;
+  display: flex;
+  justify-content: ${({ position }) =>
+    position === 'left' ? 'flex-start' : 'space-between'};
+  margin-top: 30px;
```
rosslo: I think we should use `padding-top` to prevent margin-collapsing.

The height of < Accordion /> doens't include the margin-top of first < Accordion.Panel />
image

franciscrispin: Oh actually this is intentional because according to the design, the hover size should not include the margin-top :pray: https://app.zeplin.io/project/5ba343bf0c399c33db483653/screen/5ba358620049d012fcb8b234

src/Style/Display/AccordionStyle.ts
```diff
@@ -1,59 +1,103 @@
...
+  svg {
+    width: ${`${iconSize}px`};
+    height: ${`${iconSize}px`};
+  }
+
+  ${({ position, active }) => {
+    if (position === 'right') {
+      return css`
+        transform: ${active ? 'rotate(180deg)' : 'none'};
+        transition: transform 0.24s;
```
rosslo: can use `${transitionTiming}` here

PR #188, opened 2020-01-14
Select Errors

westwood846:
This adds properties `error` and `CustomError` to Select. When `error` is set to `true`, the Select will behave just like when setting `status` to `error`. When set to a string, the message will be displayed below the Select. By default, the message will be shown in a `span` with red text color, but supplying a `CustomError` component will render the message inside the `CustomError` instead.

This will be useful for unifying the behaviour in the frontend and will allow us to revert the Selects in the profile page modals back to the Glints Aries Select.

### DIFF
--- a/src/Input/Select/Select.test.tsx
+++ b/src/Input/Select/Select.test.tsx
@@ -110,6 +110,39 @@ describe('when it is rendered', () => {
     const selectInput = getByRole('combobox');
     expect(selectInput).toHaveTextContent('');
   });
+
+  it('should not render the default error if error is falsy', () => {
+    const { getByRole } = render(<Select>{SelectChildren}</Select>);
+    const selectContainer = getByRole('combobox').parentElement.parentElement;
+    expect(selectContainer.lastChild).not.toHaveClass('select-error-default');
+  });
+
+  it('should not render the default error if error is true', () => {
+    const { getByRole } = render(
+      <Select error={true}>{SelectChildren}</Select>
+    );
+    const selectContainer = getByRole('combobox').parentElement.parentElement;
+    expect(selectContainer.lastChild).not.toHaveClass('select-error-default');
+  });
+
+  it('should render the error string in the default error', () => {
+    const { getByRole } = render(
+      <Select error={'test'}>{SelectChildren}</Select>
+    );
+    const selectContainer = getByRole('combobox').parentElement.parentElement;
+    expect(selectContainer.lastChild).toHaveClass('select-error-default');
+    expect(selectContainer.lastChild).toHaveTextContent('test');
+  });
+
+  it('should render the error component in the custom error', () => {
+    const errorComponent = <span className="errorComponent" />;
+    const { getByRole } = render(
+      <Select error={errorComponent}>{SelectChildren}</Select>
+    );
+    const selectContainer = getByRole('combobox').parentElement.parentElement;
+    expect(selectContainer.lastChild).toHaveClass('select-error-default');
+    expect(selectContainer.lastChild.firstChild).toHaveClass('errorComponent');
+  });
 });
 
 describe('when it is clicked', () => {
@@ -213,41 +246,6 @@ describe('when error is a component', () => {
     );
     const selectInput = getByRole('combobox');
     expect(selectInput).toHaveStyle(`border-color: ${PrimaryColor.glintsred}`);
-  });
-});
-
-describe('when renderError is not given', () => {
-  it('should not render the default error if error is falsy', () => {
-    const { getByRole } = render(<Select>{SelectChildren}</Select>);
-    const selectContainer = getByRole('combobox').parentElement.parentElement;
-    expect(selectContainer.lastChild).not.toHaveClass('select-error-default');
-  });
-
-  it('should not render the default error if error is true', () => {
-    const { getByRole } = render(
-      <Select error={true}>{SelectChildren}</Select>
-    );
-    const selectContainer = getByRole('combobox').parentElement.parentElement;
-    expect(selectContainer.lastChild).not.toHaveClass('select-error-default');
-  });
-
-  it('should render the error string in the default error', () => {
-    const { getByRole } = render(
-      <Select error={'test'}>{SelectChildren}</Select>
-    );
-    const selectContainer = getByRole('combobox').parentElement.parentElement;
-    expect(selectContainer.lastChild).toHaveClass('select-error-default');
-    expect(selectContainer.lastChild).toHaveTextContent('test');
-  });
-
-  it('should render the error component in the custom error', () => {
-    const errorComponent = <span className="errorComponent" />;
-    const { getByRole } = render(
-      <Select error={errorComponent}>{SelectChildren}</Select>
-    );
-    const selectContainer = getByRole('combobox').parentElement.parentElement;
-    expect(selectContainer.lastChild).toHaveClass('select-error-default');
-    expect(selectContainer.lastChild.firstChild).toHaveClass('errorComponent');
   });
 });
 
--- a/src/Input/Select/Select.test.tsx
+++ b/src/Input/Select/Select.test.tsx
@@ -186,6 +186,118 @@ describe("when status = 'error'", () => {
   });
 });
 
+describe('when error is true', () => {
+  it('should show a red border', () => {
+    const { getByRole } = render(
+      <Select error={true}>{SelectChildren}</Select>
+    );
+    const selectInput = getByRole('combobox');
+    expect(selectInput).toHaveStyle(`border-color: ${PrimaryColor.glintsred}`);
+  });
+});
+
+describe('when error is a string', () => {
+  it('should show a red border', () => {
+    const { getByRole } = render(
+      <Select error="error">{SelectChildren}</Select>
+    );
+    const selectInput = getByRole('combobox');
+    expect(selectInput).toHaveStyle(`border-color: ${PrimaryColor.glintsred}`);
+  });
+});
+
+describe('when error is a component', () => {
+  it('should show a red border', () => {
+    const { getByRole } = render(
+      <Select error={<span>error</span>}>{SelectChildren}</Select>
+    );
+    const selectInput = getByRole('combobox');
+    expect(selectInput).toHaveStyle(`border-color: ${PrimaryColor.glintsred}`);
+  });
+});
+
+describe('when renderError is not given', () => {
+  it('should not render the default error if error is falsy', () => {
+    const { getByRole } = render(<Select>{SelectChildren}</Select>);
+    const selectContainer = getByRole('combobox').parentElement.parentElement;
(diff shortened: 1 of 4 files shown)


src/Input/Select/Select.tsx
```diff
@@ -317,6 +326,8 @@ interface Props extends React.ComponentPropsWithoutRef<typeof SelectInput> {
   noOptionResult?: string;
   removeDropIcon?: boolean;
   removeFloatingLabel?: boolean;
+  error?: string | boolean;
```
franciscrispin: Same comment - error could be an object as well if we use `FormattedMessage`

stories/Input/SelectStory.tsx
```diff
@@ -26,9 +26,28 @@ const props = {
...
+      defaultValue: '',
+      possibleValue: 'any',
+      require: 'no',
+      description:
+        'Sets error state on component. If string, the error will be shown below the select',
+    },
+    {
+      name: 'ErrorComponent',
+      type: 'Component',
+      defaultValue: '<DefaultError />',
```
franciscrispin: I think we can just show that it is a `<span color="#hexcode" />`

stories/Input/SelectStory.tsx
```diff
@@ -145,6 +164,45 @@ const SelectStory = () => (
...
+        <Select.Option value="software engineer">
+          Software Engineer
+        </Select.Option>
+      </Select>
+    </div>
+    <div style={{ width: '300px' }}>
+      <Select
+        label="Jobs"
+        error="🎂 I am a custom error message"
+        ErrorComponent={({ children }) => (
```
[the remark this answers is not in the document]
franciscrispin: Oh I used `error` instead of `children` because that seems more semantic. No need for the destructuring props here as well because it is just a function returning a component, rather than a component itself.

src/Input/Select/Select.tsx
```diff
@@ -305,6 +308,12 @@ class Select extends React.Component<Props, State> {
           handleClick={this.handleClick}
           handleMouseEnter={this.handleMouseEnter}
         />
+        {typeof error === 'string' &&
```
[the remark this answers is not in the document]
westwood846: I think we should not cover that case and only allow strings to be passed in here. So the translation would have to happen in the component controlling this error. That way the select can remain agnostic of any i8n happening on the strings it's displaying.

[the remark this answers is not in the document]
franciscrispin: How does the translation happen in the component? From a user POV it is much easier to pass error={<FormattedMessage id="" defaultMessage="" />}. Also I recall one of the feedback from our glints-aries discussion is that the props we take in shouldn't be too restrictive. I believe it is related to this https://docs.google.com/spreadsheets/d/1jjW5oP9atI9FciA_-gnZQsqgsLEvF2zi-iHX9x8mM88/edit#gid=8065191&range=B28

stories/Input/SelectStory.tsx
westwood846: Updated!

src/Input/Select/Select.tsx
westwood846: See comment above.

src/Input/Select/Select.tsx
```diff
@@ -266,7 +269,7 @@ class Select extends React.Component<Props, State> {
             role="combobox"
             aria-expanded={isFocus}
             aria-autocomplete="list"
-            status={status}
+            status={status || (error && 'error')}
```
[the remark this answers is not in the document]
westwood846: Updated!

stories/Input/SelectStory.tsx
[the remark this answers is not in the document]
westwood846: Updated!

src/Input/Select/Select.tsx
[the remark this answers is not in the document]
westwood846: `error={<FormattedMessage id="" defaultMessage="" />}` this is a thing that I think we should not be doing, even though it's convenient.

But anyway, I guess in light of that KR we should be lenient here. I added the option to pass a component.

westwood846:
@franciscrispin any thoughts what the timeline for the deprecation of the status prop should look like? Have you done this kind of thing before?

I believe the status prop is rarely used for the success state.

src/Input/Select/Select.test.tsx
```diff
@@ -126,12 +131,126 @@ describe('when no results are found', () => {
...
+  it('should show a red border', () => {
+    const { getByRole } = render(
+      <Select error={<span>error</span>}>{SelectChildren}</Select>
+    );
+    const selectInput = getByRole('combobox');
+    expect(selectInput).toHaveStyle(`border-color: ${PrimaryColor.glintsred}`);
+  });
+});
+
+describe('when renderError is not given', () => {
```
franciscrispin: Do you think the test cases for both `when renderError is not given` and `when renderError is given` is necessary? My initial thought is that we can just test for `when renderError is given`. If not we risk duplicating test for every single prop passed into the component if we follow this convention. If we really want to test it, maybe we can put it in the `when it is rendered` block on line 74. Wdyt?

stories/Input/SelectStory.tsx
```diff
@@ -27,9 +27,28 @@ const props = {
...
+      type: 'ReactNode | boolean | string',
+      defaultValue: '',
+      possibleValue: 'any',
+      require: 'no',
+      description:
+        'Sets error state on component. If string, the error will be shown below the select',
+    },
+    {
+      name: 'renderError',
+      type: 'ReactNode',
```
franciscrispin: I think this should be ``type: function` instead?

src/Input/Select/Select.tsx
```diff
@@ -259,6 +270,17 @@ class Select extends React.Component<Props, State> {
       isLoading,
     } = this.state;
 
+    const deprecatedStatus = status || (error && 'error');
+
+    const shouldRenderError = !error || typeof error === 'boolean';
```
franciscrispin: Oh lines 275-282 seems a bit hard to grok for me :sweat_smile: 

I was confused by `shouldRenderError` because I assumed it was a flag to check if the user passed a function to `renderError`. But then we go on to check `renderError ? ` again. Maybe there is a more readable way to write this without the nested tenary? :pray:

src/Input/Select/Select.test.tsx
```diff
@@ -126,12 +131,126 @@ describe('when no results are found', () => {
...
+  });
+
+  it('should render the error component in the custom error', () => {
+    const errorComponent = <span className="errorComponent" />;
+    const { getByRole } = render(
+      <Select error={errorComponent}>{SelectChildren}</Select>
+    );
+    const selectContainer = getByRole('combobox').parentElement.parentElement;
+    expect(selectContainer.lastChild).toHaveClass('select-error-default');
+    expect(selectContainer.lastChild.firstChild).toHaveClass('errorComponent');
```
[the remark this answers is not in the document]
westwood846: Screenshot 2020-02-11 at 15 52 32
Screenshot 2020-02-11 at 16 05 59

Oh man I can't make this work... I agree, `toContainElement` would be nicer but I can't figure this out. Not familiar with that testing lib at all. Do you know how to do this?

src/Input/Select/Select.test.tsx
westwood846: I moved the tests, I think it's less confusing this way. I really want these tests because I want to test the rather complex interaction between `error` and `renderError`. No need to test combinations with other props.

stories/Input/SelectStory.tsx
westwood846: I was following this: https://github.com/typescript-cheatsheets/react-typescript-cheatsheet#useful-react-prop-type-examples Do you think people won't understand `ReactNode`?

src/Input/Select/Select.tsx
westwood846: Yeah I see that this is confusing. I rewrote the logic to have a flag `shouldShowError` which actually determines whether we're going to render an error or not. And then I've replaced the nested ternary with a nested if/else, not beautiful but more readable. What do you think?

src/Input/Select/Select.test.tsx
franciscrispin: Okay got it!

stories/Input/SelectStory.tsx
franciscrispin: Ahh I see. I didn't understand at first but after reading it makes sense :+1:

src/Input/Select/Select.tsx
franciscrispin: Yeap, more readable. Thanks!

src/Input/Select/Select.test.tsx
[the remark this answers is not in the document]
franciscrispin: This should work :smile: 

```js
it('should render the error string in the custom error', () => {
    const customError = (e: string) => (
      <div data-testid="custom-error">{e}</div>
    );
    const { container, getByTestId } = render(
      <Select renderError={customError} error={'nasty error'}>
        {SelectChildren}
      </Select>
    );
    const errorHTML = getByTestId('custom-error').outerHTML;
    expect(container).toContainHTML(errorHTML);
    expect(container).toHaveTextContent('nasty error');
  });
```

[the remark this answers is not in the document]
franciscrispin: This should work :nerd_face: 
```
  it('should render the error string in the custom error', () => {
    const customError = (e: string) => (
      <div data-testid="custom-error">{e}</div>
    );
    const { container, getByTestId } = render(
      <Select renderError={customError} error="nasty error">
        {SelectChildren}
      </Select>
    );
    const errorHTML = getByTestId('custom-error');
    expect(container).toContainElement(errorHTML);
    expect(container).toHaveTextContent('nasty error');
  });
```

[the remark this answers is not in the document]
franciscrispin: Opps bad naming, `errorHTML` should be `errorElement` instead.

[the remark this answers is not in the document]
westwood846: Updated, please take a look.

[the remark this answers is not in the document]
franciscrispin: LGTM

franciscrispin:
@westwood846 LGTM. Sorry this took 1 month to merge hahaha. You can merge it after you resolve the conflict.

franciscrispin:
> @franciscrispin any thoughts what the timeline for the deprecation of the status prop should look like? Have you done this kind of thing before?
> 
> I believe the status prop is rarely used for the success state. Grepping through employers and dst (`grep -R 'status="success"' dst/app`) shows only one match:
> 
> ```
> dst/app/components/OpportunityPsychFlatModal/CoverLetter.js:        status="success"
> ```

@westwood846 turns out status prop is used in dst https://gitlab.glints.com/glints/glints-dst/-/blob/staging/app/modules/LiveDiscussion/Components/CustomInputComponents.js

Seems like the grep was a bit too simplistic haha

PR #190, opened 2020-01-16
Select: Enable scrolling with up and down arrow keys

franciscrispin:


stories/Input/SelectStory.tsx
```diff
@@ -143,6 +143,9 @@ const SelectStory = () => (
         <Select.Option value="software engineer">
           Software Engineer
         </Select.Option>
+        <Select.Option value="finance">Finance</Select.Option>
+        <Select.Option value="design">Design</Select.Option>
+        <Select.Option value="human resources">Human Resources</Select.Option>
```
rosslo: I just noticed we have package `jsx-to-string`.

So I think we can use it to prevent duplication of `usage`

```
// Select.tsx
const Option: React.FunctionComponent<SelectItemProps> = () => null;
Option.displayName = 'Select.Option';

class Select extends React.Component<Props, State> {
  static Option = Option;
  ...
};

// SelectStory.tsx
const Usage = (
  <Select label="Jobs">
    <Select.Option value="accountant">Accountant</Select.Option>
    <Select.Option value="business development">
      Business Development
    </Select.Option>
    <Select.Option value="software engineer">Software Engineer</Select.Option>
    <Select.Option value="finance">Finance</Select.Option>
    <Select.Option value="design">Design</Select.Option>
    <Select.Option value="human resources">Human Resources</Select.Option>
  </Select>
);

const SelectStory = () => (
  <StorybookComponent
    ...
    usage={Usage}
  >
    <div style={{ width: '300px' }}>{Usage}</div>
  </StorybookComponent>
);
```

franciscrispin: Wow didn't know this existed haha. Thanks @rosslo

PR #219, opened 2020-01-20
Icon: Add more props to increase customizability

franciscrispin:
Previously, devs would have to create wrapper components to style the Icon. E.g.:
```
<IconWrapper>
  <GlintsAriesIcon />
</IconWrapper>
```

Added props to fix this issue and allow for better customization.
- width, height, className, onClick

### DIFF
--- a/src/General/Icon/Icon.tsx
+++ b/src/General/Icon/Icon.tsx
@@ -1,18 +1,38 @@
 import * as React from 'react';
 
 const Icon: React.FunctionComponent<Props> = props => {
-  const { children, color } = props;
+  const {
+    className,
+    children,
+    color,
+    width = '1em',
+    height = '1em',
+    onClick,
+    ...restProps
+  } = props;
 
   return (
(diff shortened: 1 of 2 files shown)


src/General/Icon/Icon.tsx
```diff
@@ -1,18 +1,38 @@
 import * as React from 'react';
 
 const Icon: React.FunctionComponent<Props> = props => {
-  const { children, color } = props;
+  const {
+    className,
+    children,
+    color,
```
rosslo: What do you think we set `currentColor` as default color prop here? That's what I mentioned that icon should be able to inherit color. https://caniuse.com/#search=currentColor

franciscrispin: @rosslo I think we can use currentColor if we find that most of the time, the icon color should be the same as the inherited color (from the parent container). I'm not sure if this is the case though. In any case, I think this change might lead to some breaking changes for our UI so maybe we can explore this in another issue?

PR #220, opened 2020-01-21
Modal: Refactor to Handle onClose Properly

westwood846:
This MR refactors the Modal component to remove the internal state, which before has been completely controlled by the `isVisible` prop, which made it redundant.

This MR also refactors the way `onClose` is called. Previously, the `componentWillUnmount` was supposed to unregister the escape key event handler, so that the event would not trigger while the modal was closed. This did not work because the modal's visibility is toggled through the `display` CSS prop, so the component would never actually unmount.

### DIFF
--- a/src/Display/Modal/Modal.tsx
+++ b/src/Display/Modal/Modal.tsx
@@ -23,8 +23,7 @@ class Modal extends React.Component<Props> {
   }
 
   componentDidMount() {
-    const { onClose } = this.props;
-    document.addEventListener('keydown', escEvent(onClose), false);
+    this.onOpen();
   }
 
   componentDidUpdate(prevProps: Props) {
@@ -43,6 +42,15 @@ class Modal extends React.Component<Props> {
   }
 
   componentWillUnmount() {
+    this.onClose();
+  }
+
+  onOpen() {
+    const { onClose } = this.props;
+    document.addEventListener('keydown', escEvent(onClose), false);
+  }
+
+  onClose() {
     const { onClose } = this.props;
     document.removeEventListener('keydown', escEvent(onClose), false);
     document.body.removeAttribute('style');
--- a/src/Display/Modal/Modal.tsx
+++ b/src/Display/Modal/Modal.tsx
@@ -13,7 +13,7 @@ import {
   ModalFooter,
 } from '../../Style/Display/ModalStyle';
 
-class Modal extends React.Component<Props, State> {
+class Modal extends React.Component<Props> {
   modalContentAreaRef: React.RefObject<HTMLDivElement>;
   mouseDownTarget: HTMLElement;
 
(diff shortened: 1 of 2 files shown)


src/Display/Modal/Modal.tsx
```diff
@@ -13,9 +13,10 @@ import {
   ModalFooter,
 } from '../../Style/Display/ModalStyle';
 
-class Modal extends React.Component<Props, State> {
+class Modal extends React.Component<Props> {
   modalContentAreaRef: React.RefObject<HTMLDivElement>;
   mouseDownTarget: HTMLElement;
+  escEvent: (e: KeyboardEvent) => any;
 
   state = {
```
tingtingdisco: state can be removed

src/Display/Modal/Modal.tsx
```diff
@@ -26,34 +27,35 @@ class Modal extends React.Component<Props, State> {
...
-    document.addEventListener('keydown', escEvent(onClose), false);
-  }
+  componentDidUpdate(prevProps: Props) {
+    const { isVisible } = this.props;
 
-  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
-    if (nextProps.isVisible && !prevState.isOpen) {
+    // closed -> open
+    if (!prevProps.isVisible && isVisible) {
+      this.modalContentAreaRef.current.focus();
```
tingtingdisco: can we remove the `modalContentAreaRef`?

src/Display/Modal/Modal.tsx
westwood846: Removed it!

src/Display/Modal/Modal.tsx
westwood846: I did not touch that logic in this PR. No idea what that's about in the first place haha.

tingtingdisco: yeah, there seems to be some dead code and unnecessary code, like the `escEvent` which is essentially an alias of `onClose`, maybe can be removed as well.

src/Display/Modal/Modal.tsx
```diff
@@ -13,47 +13,49 @@ import {
...
   }
 
   componentWillUnmount() {
+    this.handleClose();
+  }
+
+  handleOpen() {
     const { onClose } = this.props;
-    document.removeEventListener('keydown', escEvent(onClose), false);
+    this.escEvent = escEvent(onClose);
```
tingtingdisco: can do `this.escEvent = escEvent(onClose);` in constructor?

src/Display/Modal/Modal.tsx
```diff
@@ -13,47 +13,49 @@ import {
...
+      this.modalContentAreaRef.current.focus();
       document.body.style.overflow = 'hidden';
-      return { isOpen: true };
+      this.handleOpen();
     }
 
-    if (!nextProps.isVisible && prevState.isOpen) {
+    // open -> closed
+    if (prevProps.isVisible && !isVisible) {
       document.body.removeAttribute('style');
```
tingtingdisco: this line is in `handleClose` already, can be removed

src/Display/Modal/Modal.tsx
```diff
@@ -13,47 +13,49 @@ import {
...
-  }
+  componentDidUpdate(prevProps: Props) {
+    const { isVisible } = this.props;
 
-  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
-    if (nextProps.isVisible && !prevState.isOpen) {
+    // closed -> open
+    if (!prevProps.isVisible && isVisible) {
+      this.modalContentAreaRef.current.focus();
       document.body.style.overflow = 'hidden';
```
tingtingdisco: can move `document.body.style.overflow = 'hidden';` to `handleOpen`

PR #221, opened 2020-01-21
Modal: Don't Call onClose When Closed

westwood846:
In some cases the onClose function is being called even when the modal is already closed.

For instance, this can happen because the escape event listener does not get unregistered when the modal closes. This is a bug but I've decided to treat the symptom rather than the root cause because this component will likely get rewritten soon anyway.

I tried fixing the root cause before in https://github.com/glints-dev/glints-aries/pull/220 but the state management of Modal is more complex than I realized at first. A more thorough redesign of this component's API will be necessary to define the desired behavior and implement it, with or without the internal state.

### DIFF
--- a/src/Display/Modal/Modal.tsx
+++ b/src/Display/Modal/Modal.tsx
@@ -32,9 +32,13 @@ class Modal extends React.Component<Props, State> {
     if (this.state.isOpen) onClose();
   }
 
-  componentDidMount() {
+  registerEventEscapeListener() {
     this.escEvent = escEvent(this.guardedOnClose.bind(this));
     document.addEventListener('keydown', this.escEvent, false);
+  }
+
+  componentDidMount() {
+    this.registerEventEscapeListener();
   }
 
   static getDerivedStateFromProps(nextProps: Props, prevState: State) {
--- a/src/Display/Modal/Modal.tsx
+++ b/src/Display/Modal/Modal.tsx
@@ -26,9 +26,17 @@ class Modal extends React.Component<Props, State> {
     this.modalContentAreaRef = React.createRef();
   }
 
-  componentDidMount() {
+  guardedOnClose() {
     const { onClose } = this.props;
-    document.addEventListener('keydown', escEvent(onClose), false);
+    if (this.state.isOpen) onClose();
+  }
+
+  componentDidMount() {
+    document.addEventListener(
+      'keydown',
+      escEvent(this.guardedOnClose.bind(this)),
+      false
+    );
   }
 
   static getDerivedStateFromProps(nextProps: Props, prevState: State) {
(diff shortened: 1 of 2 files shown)


src/Display/Modal/Modal.tsx
```diff
@@ -52,8 +60,11 @@ class Modal extends React.Component<Props, State> {
   }
 
   componentWillUnmount() {
-    const { onClose } = this.props;
-    document.removeEventListener('keydown', escEvent(onClose), false);
+    document.removeEventListener(
+      'keydown',
+      escEvent(this.guardedOnClose),
```
tingtingdisco: this event listener will not be removed, because `escEvent(this.guardedOnClose)` creates another function. We need to pass the same event handler function we passed to `addEventListener` here.

westwood846: Yeah I'm not trying to fix that with this PR.

With this the event handler will still be called but it won't call `onClose` because of the check.

westwood846: @tingtingdisco so is this good?

tingtingdisco: @westwood846 sorry for the late reply, I think the event handler should be correctly removed, it's related to your change.

westwood846: Updated!

src/Display/Modal/Modal.tsx
```diff
@@ -26,9 +27,18 @@ class Modal extends React.Component<Props, State> {
...
+    if (this.state.isOpen) onClose();
+  }
+
+  registerEventEscapeListener() {
+    this.escEvent = escEvent(this.guardedOnClose.bind(this));
+    document.addEventListener('keydown', this.escEvent, false);
+  }
+
+  componentDidMount() {
+    this.registerEventEscapeListener();
```
tingtingdisco: maybe we can `addEventListener` only when the modal is visible, then we don't need the `guardedOnClose`
 ```
componentDidMount() {
  if (!this.props.isVisible) return;
  this.registerEventEscapeListener();
}
```

Comment (tingtingdisco) [reply src/Display/Modal/Modal.tsx]: and suggested changes on registerEventEscapeListener and removeEscapeEventListener:
```
onModalOpen() {
  this.escEvent = escEvent(this.guardedOnClose.bind(this));
  document.addEventListener('keydown', this.escEvent, false);
  document.body.style.overflow = 'hidden';
}

onModalClose() {
  document.removeEventListener('keydown', this.escEvent, false);
}
```

tingtingdisco: the state `isOpen` seems to be a duplicate of the prop `isVisible`, can we remove it?

westwood846: > the state isOpen seems to be a duplicate of the prop isVisible, can we remove it?
I tried this but it increased the scope of this so I decided not to do it. I kinda need to merge this one soon.

> so we can remove `getDerivedStateFromProps`
Same as above. `getDerivedStateFromProps` is needed for the state.

> maybe we can `addEventListener` only when the modal is visible
Thanks for the suggestion, updated! Much more readable this way.

tingtingdisco:
LGTM

tingtingdisco:
@westwood846 do you know how to release a new version?

westwood846:
Thanks @tingtingdisco ! I will try to follow the instructions here: https://github.com/glints-dev/glints-aries/wiki/Contributing

PR #226, opened 2020-02-06
Add `NumberInput` Component

westwood846:
I need a number input for a feature on employers and I think it should be in Glints Aries. The `TextField `component is easily extended for this purpose. In case we want to go for a more sophisticated solution in the future this PR creates a wrapper around `TextField` and exports a new `NumberInput` component. This way we won't have to change any imports when we change the `NumberInput`.

There's also a new story for the new component which reuses parts of the `TextFieldStory` to ensure that changes made to `TextField` will also be documented in `NumberInputStory`, since `NumberInput` it is basically a copy of the `TextField`. 

The test is a bit minimalistic because the component is basically the same as `TextField`.


tingtingdisco:
I cannot delete this zero, and do you think we should hide the browser UI up and down arrows?
Screen Shot 2020-02-10 at 1 20 37 PM

src/Input/NumberInput/NumberInput.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return { numberInput, utils };
+}
+
+function setupNumberInputWithChangeEvent(otherProps?: any) {
+  const { numberInput } = setupNumberInput(otherProps);
+  userEvent.type(numberInput, inputValue);
+  return { numberInput };
+}
+
+it('<NumberInput> should render with a label of Username', () => {
```
tingtingdisco: what is `a label of Username`? can we call this test `it('renders correctly to match snapshot')`

src/Input/NumberInput/NumberInput.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const TextFieldSnapshot = renderer
+    .create(<NumberInput label={props.label} />)
+    .toJSON();
+  expect(TextFieldSnapshot).toMatchSnapshot();
+});
+
+describe('when it is rendered', () => {
+  it('should have an empty string as its value', () => {
```
tingtingdisco: I guess we can add more context to the test case like `describe('when it is rendered without value passed to it' `

src/Input/NumberInput/NumberInput.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+function setupNumberInput(otherProps?: any) {
+  const { getByLabelText, ...utils } = render(
+    <NumberInput
+      label={props.label}
+      onChange={props.onChange}
+      {...otherProps}
+    />
+  );
+  const numberInput = getByLabelText(props.label) as HTMLInputElement;
+  return { numberInput, utils };
```
tingtingdisco: `utils` is not being used, can we remove it?

src/Input/NumberInput/index.ts
```diff
@@ -0,0 +1,5 @@
+import NumberInput from './NumberInput';
+
+export { NumberInput };
+
+export default NumberInput;
```
tingtingdisco: why we need both named export and default export?

src/Input/TextField/TextField.tsx
```diff
@@ -135,19 +135,23 @@ class TextField extends React.Component<Props, State> {
...
+export interface Props
+  extends React.ComponentPropsWithoutRef<typeof TextFieldInput> {
   type: textFieldType;
   label: string;
   disabled?: boolean;
   className?: string;
   removeFloatingLabel?: boolean;
+  min?: number;
+  max?: number;
+  step?: number;
```
tingtingdisco: I am not sure whether we need `step`,  do you need it in your use case?

src/Input/NumberInput/index.ts
westwood846: Yes. I saw it done in other components here in the repo.

src/Input/TextField/TextField.tsx
westwood846: I don't need it but I added it because this is supposed to be a replacement for `<input type="number />` which has it. Since this is a component library I figured it would be good to cover as many potential use cases as possible.

westwood846:
> I cannot delete this zero, and do you think we should hide the browser UI up and down arrows?
> Screen Shot 2020-02-10 at 1 20 37 PM

I will look into the zero.

I left the up and down arrows on purpose because I think they belong on a number input. This component should eventually have them in proper Glints style but for now the default ones work.

src/Input/TextField/TextField.tsx
tingtingdisco: got it

tingtingdisco:
@westwood846 regarding to the arrows UI, could you confirm it with designers, if they are ok with it then we can have it.

Issue #222, opened 2020-02-05
<Select /> with disableTyping shouldn't have cursor and open mobile keyboard when on focus.


rosslo:
Failing case: https://gitlab.glints.com/glints/glints-dst/merge_requests/1107#note_29589

rosslo:
PR: https://github.com/glints-dev/glints-aries/pull/228

PR #182, opened 2020-01-03
Deprecate/job card

franciscrispin:
https://github.com/glints-dev/glints-aries/issues/181

PR #228, opened 2020-02-12
disable typing Select shouldn't have cursor and open mobile keyboard

rosslo:
Related issue: https://github.com/glints-dev/glints-aries/issues/222

### DIFF
--- a/src/Input/Select/Select.tsx
+++ b/src/Input/Select/Select.tsx
@@ -281,6 +281,7 @@ class Select extends React.Component<Props, State> {
             value={selectedValue}
             small={small}
             disableTyping={disableTyping}
+            readOnly={disableTyping}
             {...defaultProps}
           />
           {!removeFloatingLabel && (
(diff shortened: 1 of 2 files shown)


westwood846:
Can you please remove the `yarn.lock` commit? Just delete it with interactive rebase and force-push to this branch.

rosslo:
> Can you please remove the `yarn.lock` commit? Just delete it with interactive rebase and force-push to this branch. IMO `yarn.lock` updates should only go with dependency changes or deliberate dependency upgrades.

rosslo:
@westwood846 Yeah, agree with your point to yarn.lock updates. But I think it's okay to update `yarn.lock` within this MR, those updates of it related to https://github.com/glints-dev/glints-aries/pull/182/files. Which the dev forgot to update `yarn.lock`.

westwood846:
> Or do you think we should have another for it?

Usually I would prefer another MR but I don't see the point of being so pedantic here. Should be fine :D

westwood846:
LGTM

Issue #229, opened 2020-02-17
Select: Pressing Enter key on Select.Option not firing `onOptionClick`


franciscrispin:
Steps to reproduce:
1. Add `onOptionClick={() => console.log('clicked')}` to any `Select.Option`
2. Click that option and see that "clicked" is logged
3. Use the arrow keys to navigate to the option and press the enter key

Expected: Console should log "clicked"
Actual: Console does not log "clicked"

https://www.loom.com/share/4e8e7bfcd3314a25b4189d1a40a2070e

PR #230, opened 2020-02-17
Select: Call onOptionClick when enter key is pressed on an option

franciscrispin:
https://github.com/glints-dev/glints-aries/issues/229

### DIFF
--- a/src/Input/Select/Select.tsx
+++ b/src/Input/Select/Select.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
-import { isEqual } from 'lodash';
+import { get, isEqual } from 'lodash';
 
 import classNames from 'classnames';
 
@@ -197,6 +197,14 @@ class Select extends React.Component<Props, State> {
     } else if (e.keyCode === 13) {
       inputElement.blur();
       const activeElement = this.getActiveElement();
+      if (!activeElement) {
+        return;
+      }
+
+      const activeElementIndex = Number(get(activeElement, 'dataset.id'));
+      const activeChild = React.Children.toArray(children)[activeElementIndex];
+      const onOptionClick = get(activeChild, 'props.onOptionClick');
+
       this.setState({
         selectedValue: activeElement.innerText,
         filterValue: React.Children.map(children, data => data),
@@ -206,6 +214,9 @@ class Select extends React.Component<Props, State> {
       if (onChange !== undefined) {
         const itemValue = activeElement.dataset.value;
         onChange(itemValue);
+      }
+      if (onOptionClick !== undefined) {
+        onOptionClick(e);
       }
     } else if (e.keyCode === 27) {
       inputElement.blur();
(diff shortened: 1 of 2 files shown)


src/Input/Select/Select.tsx
```diff
@@ -197,6 +197,14 @@ class Select extends React.Component<Props, State> {
     } else if (e.keyCode === 13) {
       inputElement.blur();
       const activeElement = this.getActiveElement();
+      const activeElementIndex = Number(get(activeElement, 'dataset.id'));
+      const activeChild =
+        isInteger(activeElementIndex) &&
```
rosslo: is `isInteger` necessary here? since `dataset.id` is assigned by `index` which must be an integer.

src/Input/Select/Select.tsx
```diff
@@ -197,6 +197,14 @@ class Select extends React.Component<Props, State> {
     } else if (e.keyCode === 13) {
       inputElement.blur();
       const activeElement = this.getActiveElement();
+      const activeElementIndex = Number(get(activeElement, 'dataset.id'));
```
rosslo: Since `Number(null) === 0`, `parseInt` might be safer. But might need `isNaN` check as well.

src/Input/Select/Select.tsx
```diff
@@ -197,6 +197,14 @@ class Select extends React.Component<Props, State> {
...
+      const activeChild =
+        isInteger(activeElementIndex) &&
+        React.Children.toArray(children).find(
+          (child, i) => i === activeElementIndex
```
rosslo: I think we can simply use `React.Children.toArray(children)[activeElementIndex]` instead.

src/Input/Select/Select.tsx
```diff
@@ -207,6 +215,9 @@ class Select extends React.Component<Props, State> {
         const itemValue = activeElement.dataset.value;
         onChange(itemValue);
       }
+      if (onOptionClick !== undefined) {
+        onOptionClick();
```
rosslo: Maybe we can pass `event` as the parameter as we do in `handleClick`. Just come up with a scenario is devs can know the `onOptionClick` is triggered by keyboard or mouse via the `event` parameter.

src/Input/Select/Select.tsx
franciscrispin: Oh maybe i'll do `Number(get(activeElement, 'dataset.id', undefined));` instead. Will that work?

src/Input/Select/Select.tsx
franciscrispin: Nope not necessary, added it to be more defensive in case `dataset.id`` gets overwritten somehow.

src/Input/Select/Select.tsx
rosslo: There also shouldn't be a way for `dataset.id` to be `undefined`. 🤔 
I think while `dataset.id` is `undefined`, `activeElement` should be `undefined` as well.
Therefore, we might need the condition `if(!activeElement) return;` after line 199.

Issue #218, opened 2020-01-17
Make text input components controllable


westwood846:
Please specify here which components this means before working on this.

tingtingdisco:
#236 implemented this issue.
We can pass a ref prop to Textarea and TextField to control their underlying input element.

PR #236, opened 2020-02-24
Feature/add ref to textarea and textfield

tingtingdisco:
this PR implements issue #218 

wrap Textarea and TextField in `forwardRef` to let a `ref` prop can be directly passed to the underlying `textarea` and `input` element.

### DIFF
--- a/src/Input/TextField/TextField.test.tsx
+++ b/src/Input/TextField/TextField.test.tsx
@@ -146,3 +146,31 @@ describe('when removeFloatingLabel is true', () => {
     expect(textFieldInput).toBeTruthy();
   });
 });
+
+describe('<TextField /> forwards ref to underlying input element', () => {
+  const label = 'forward-ref';
+
+  test('ref is being forwarded correctly', () => {
+    const ref = React.createRef<HTMLInputElement>();
+    const { queryByLabelText } = render(
+      <TextField ref={ref} label={label} type="text" />
+    );
+    const textInput = queryByLabelText(label);
+    expect(textInput).toEqual(ref.current);
+  });
+
+  test('underlying textarea can be focused/blurred through ref', () => {
+    const ref = React.createRef<HTMLInputElement>();
+    const { queryByLabelText } = render(
+      <TextField ref={ref} label={label} type="text" />
+    );
+    const textInput = queryByLabelText(label);
+    expect(textInput).not.toHaveFocus();
+
+    ref.current.focus();
+    expect(textInput).toHaveFocus();
+
+    ref.current.blur();
+    expect(textInput).not.toHaveFocus();
+  });
+});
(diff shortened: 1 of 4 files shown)


src/Input/Textarea/Textarea.test.tsx
```diff
@@ -115,3 +115,27 @@ describe('when removeFloatingLabel is true', () => {
...
   });
 });
+
+describe('<Textarea /> forwards ref to underlying textarea element', () => {
+  const label = 'forward-ref';
+
+  test('ref is being forwarded correctly', () => {
+    const ref = React.createRef<HTMLTextAreaElement>();
+    const { queryByLabelText } = render(<Textarea ref={ref} label={label} />);
+    const textarea = queryByLabelText(label);
```
rosslo: Same as the comment above.

src/Input/TextField/TextField.test.tsx
```diff
@@ -146,3 +146,31 @@ describe('when removeFloatingLabel is true', () => {
...
+
+describe('<TextField /> forwards ref to underlying input element', () => {
+  const label = 'forward-ref';
+
+  test('ref is being forwarded correctly', () => {
+    const ref = React.createRef<HTMLInputElement>();
+    const { queryByLabelText } = render(
+      <TextField ref={ref} label={label} type="text" />
+    );
+    const textInput = queryByLabelText(label);
```
rosslo: Can reuse the function `setupTextField`  at line 19 as below snippet

```
describe('<TextField /> forwards ref to underlying input element', () => {
  const ref = React.createRef<HTMLInputElement>();

  test('ref is being forwarded correctly', () => {
    const { textFieldInput } = setupTextField({ ref });
    expect(textFieldInput).toEqual(ref.current);
  });
 // ...
});
```

tingtingdisco: hmm, I think we should not expect those functions will be reused. Because it's not reasonable to review the whole test file before adding new test cases. What do you think?

tingtingdisco: another point is that writing test code and writing production code are different. In general, we want to be DRY when writing production code. But when writing test code, DRY is not that necessary in my opinion.

rosslo: Yeah, I agree with you, DRY is not that necessary for writing testing.
But I think it is also good to apply for readability, I think we can consider limiting the reusable function in `describe` scope and depend on how many repetitive codes are there.

tingtingdisco: I am not sure if I understand you correctly, which way has better readability?

I think `const { textFieldInput } = setupTextField({ ref });` is less readable, especially when we are reviewing diffs, we cannot know what `setupTextField` does unless we review the whole file to find its implementation.

Issue #211, opened 2020-01-17
Icons Can Inherit Their Parents Color


westwood846:
OKR 2.b

rosslo:
PR: https://github.com/glints-dev/glints-aries/pull/240

PR #240, opened 2020-02-26
Feature/icon inherit color

rosslo:
Issue: https://github.com/glints-dev/glints-aries/issues/211 
Reference: https://caniuse.com/#search=currentColor

### DIFF
--- a/src/General/Icon/Icon.test.tsx
+++ b/src/General/Icon/Icon.test.tsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 import * as renderer from 'react-test-renderer';
 import { fireEvent, render } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
+import styled from 'styled-components';
 
 import Icon from './Icon';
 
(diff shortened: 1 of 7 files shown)


src/General/Icon/Icon.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+import '@testing-library/jest-dom/extend-expect';
+
+import Icon from './Icon';
+
+const testId = 'icon-svg';
+const children = (
+  <path d="M100 56.563H56.562V100H43.438V56.562H0V43.438h43.438V0h13.124v43.438H100z"></path>
+);
+
+it('<Icon> should render a svg with color, width, height and a path as children', () => {
```
tingtingdisco: suggestion: 
1. replace all the `it` with `test`
2. change the description of this case to `it matches snapshot`

tingtingdisco: can you render the snapshot without `children`?

tingtingdisco: hmm, can `<Icon />` be rendered without children in all the test cases?

src/General/Icon/Icon.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const IconSnapshot = renderer.create(<Icon>{children}</Icon>).toJSON();
+  expect(IconSnapshot).toMatchSnapshot();
+});
+
+it('<Icon> should accept a className to apply the style from styled-component', () => {
```
tingtingdisco: why is `<Icon />` related to `styled-component`?

stories/General/IconStory.tsx
```diff
@@ -26,29 +25,75 @@ const props = {
     {
       name: 'color',
       type: 'string',
-      defaultValue: <code>black</code>,
+      defaultValue: <code>inherit parent's color</code>,
```
tingtingdisco: maybe `defaultValue: <code>currentColor (inherit parent's color)</code>`

src/General/Icon/Icon.test.tsx
tingtingdisco: ah, if you want to make sure it can be wrapped by `styled`, maybe you can do it in the test, otherwise I think the description can be `<Icon /> should accept a className prop`

src/General/Icon/Icon.test.tsx
rosslo: Nope, because `children` is required prop for `Icon`.
Got an error `Property 'children' is missing in type '{}' but required in type 'Props'`

tingtingdisco: ok

rosslo:
@tingtingdisco Updated, please review again 🙏

tingtingdisco:
LGTM

PR #243, opened 2020-02-28
Render Button as Configurable Tag

### LINKED ISSUE #83: <Button> components need to be configurable as HTML anchors
Requested by @yjwong 

Component: Button

Feature Request: 
Configure button component to use <a> as the base component e.g. `<Button component="a" />`

westwood846:
Adds a property 'tag' to all Button components. When specified, the button will render as the specified tag. I chose the name 'tag" because react-intl uses the same name.

Closes #83

### DIFF
--- a/src/General/Button/Button.test.tsx
+++ b/src/General/Button/Button.test.tsx
@@ -15,6 +15,25 @@ it('<Button> should render with text "click me" and an onClick handler', () => {
   expect(ButtonSnapshot).toMatchSnapshot();
 });
 
+it('<Button> should render as button by default', () => {
+  const tag = 'a';
+  const { container } = render(<Button tag={tag}>test</Button>);
+  expect(container.querySelector(tag)).toBeTruthy();
+});
+
+describe('<Button> should render as the given tag', () => {
+  const tags = ['a', 'h1', 'div'];
+  tags.forEach((tag: string) => {
+    it(`<Button> should render as ${tag}`, () => {
+      const { container } = render(
+        <Button tag={tag as React.ElementType}>button</Button>
+      );
+      const tagRegexp = new RegExp(tag, 'i');
+      expect(container.querySelector(tag).tagName).toMatch(tagRegexp);
+    });
+  });
+});
+
 it('should call onClick once when it is clicked', () => {
   const onClick = jest.fn();
   const { getByText } = render(<Button onClick={onClick}>click me</Button>);
--- a/src/General/Button/Button.test.tsx
+++ b/src/General/Button/Button.test.tsx
@@ -16,9 +16,10 @@ it('<Button> should render with text "click me" and an onClick handler', () => {
 });
 
 it('<Button> should render as button by default', () => {
-  const tag = 'a';
-  const { container } = render(<Button tag={tag}>test</Button>);
-  expect(container.querySelector(tag)).toBeTruthy();
+  const { getByTestId } = render(
+    <Button data-testid="test-button">test</Button>
+  );
+  expect(getByTestId('test-button').tagName).toBe('BUTTON');
 });
 
 describe('<Button> should render as the given tag', () => {
--- a/src/General/Button/LinkButton.tsx
+++ b/src/General/Button/LinkButton.tsx
@@ -6,9 +6,14 @@ import { LinkBtn } from '../../Style/General/ButtonStyle';
 const LinkButton: React.FunctionComponent<Props> = ({
   children,
   className,
+  tag,
   ...defaultProps
 }) => (
-  <LinkBtn className={classNames('aries-linkbtn', className)} {...defaultProps}>
+  <LinkBtn
+    className={classNames('aries-linkbtn', className)}
+    as={(tag as React.ElementType) || 'button'}
+    {...defaultProps}
+  >
     {children}
   </LinkBtn>
 );
@@ -17,6 +22,7 @@ interface Props extends React.ComponentPropsWithoutRef<typeof LinkBtn> {
   children: React.ReactNode;
   className: string;
   block?: boolean;
+  tag?: React.ElementType;
 }
 
 export default LinkButton;
--- a/src/General/Button/PrimaryButton.tsx
+++ b/src/General/Button/PrimaryButton.tsx
@@ -9,6 +9,7 @@ const PrimaryButton: React.FunctionComponent<Props> = ({
   disabled,
   block,
   small,
+  tag,
   ...defaultProps
 }) => (
   <PrimaryContainer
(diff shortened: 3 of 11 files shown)


src/General/Button/Button.test.tsx
```diff
@@ -15,6 +15,17 @@ it('<Button> should render with text "click me" and an onClick handler', () => {
   expect(ButtonSnapshot).toMatchSnapshot();
 });
 
+it('<Button> should render as ther given tag', () => {
```
franciscrispin: This test works fine but just some minor points:
- You don't have to use `data-test-id` here. You only use it when you want to use `getByTestId`
- Instead of `.not.toBeEmpty()` you could use `toBeTruthy` https://jestjs.io/docs/en/expect#tobetruthy

That said I think this test should 1. test more tags and 2. be more explicit that we are checking for the tag name. I suggest something like this instead.
```
describe('contains the correct className for each variant:', () => {
  const tags = ['a', 'h1', 'div'];
  tags.forEach((tag: string & Element) => {
    it(`<Button> should render as the given tag: ${tag}`, () => {
      const { container } = render(
        <Button tag={tag as React.ElementType}>anchor button</Button>
      );
      const tagRegexp = new RegExp(tag, 'i');
      expect(container.querySelector(tag).tagName).toMatch(tagRegexp);
    });
  });
});
```

Also, I think we should add one test for the default case where no tag is passed and it should render a `button` element.

PS: Not sure if `tag: string & Element` is bad typescript haha. You could just change it to `any` as well and remove the `React.ElementType`.

stories/General/ButtonStory.tsx
```diff
@@ -44,6 +44,21 @@ const buttonProps = [
     require: 'no',
     description: 'Sets Button to small version.',
   },
+  {
```
franciscrispin: Could we also show an example component that uses `tag="a"` since that will likely be a common use case.

stories/General/ButtonStory.tsx
```diff
@@ -44,6 +44,21 @@ const buttonProps = [
...
+    name: 'tag',
+    type: 'string',
```
franciscrispin: Maybe `React.ElementType` is more suitable

stories/General/ButtonStory.tsx
```diff
@@ -44,6 +44,21 @@ const buttonProps = [
...
+    defaultValue: <code>button</code>,
+    possibleValue: (
```
franciscrispin: I believe the  possible value should just be any valid html tag. Ideally we just want devs to pass in `tag="a"` so we can change this to `possibleValue: any valid html tag e.g. "a"`

src/General/Button/LinkButton.tsx
```diff
@@ -6,9 +6,14 @@ import { LinkBtn } from '../../Style/General/ButtonStyle';
 const LinkButton: React.FunctionComponent<Props> = ({
   children,
   className,
+  tag,
   ...defaultProps
 }) => (
-  <LinkBtn className={classNames('aries-linkbtn', className)} {...defaultProps}>
+  <LinkBtn
```
franciscrispin: I'm wondering if we should allow users to change the tag for this LinkButton in particular. Since it's name specifically states it is a `LinkButton` maybe it should stay as an anchor? Only concern is that it won't be consistent with the rest of the buttons. WDYT?

src/General/Button/PrimaryButton.tsx
```diff
@@ -38,6 +40,7 @@ interface Props extends React.ComponentPropsWithoutRef<typeof PrimaryBtn> {
   block?: boolean;
   small?: boolean;
   removeHoverEffect?: boolean;
+  tag?: React.ElementType;
```
franciscrispin: On an unrelated note, we should really abstract all these repeated interfaces to a common interface :stuck_out_tongue: Don't have to do it in this PR though.

src/General/Button/LinkButton.tsx
westwood846: I would like to leave it as it is ^^ If they just want a link, they could have just used a link instead of the button.

src/General/Button/PrimaryButton.tsx
westwood846: That's true! Well, I guess the storybook is still changing, somebody might improve this eventually ^^

src/General/Button/Button.test.tsx
westwood846: > Not sure if tag: string & Element
I just changed it to `string` since that's what is being iterated.

> add one test for the default case
Done!

> be more explicit that we are checking for the tag name
Done!

stories/General/ButtonStory.tsx
westwood846: Added!

stories/General/ButtonStory.tsx
westwood846: Changed it.

stories/General/ButtonStory.tsx
westwood846: Changed it.

src/General/Button/Button.test.tsx
```diff
@@ -15,6 +15,25 @@ it('<Button> should render with text "click me" and an onClick handler', () => {
...
+  const { container } = render(<Button tag={tag}>test</Button>);
+  expect(container.querySelector(tag)).toBeTruthy();
+});
+
+describe('<Button> should render as the given tag', () => {
+  const tags = ['a', 'h1', 'div'];
+  tags.forEach((tag: string) => {
+    it(`<Button> should render as ${tag}`, () => {
+      const { container } = render(
+        <Button tag={tag as React.ElementType}>anchor button</Button>
```
franciscrispin: Opps I realied `anchor button` is misleading haha my bad. Can just be `button` instead

src/General/Button/Button.test.tsx
```diff
@@ -15,6 +15,25 @@ it('<Button> should render with text "click me" and an onClick handler', () => {
   expect(ButtonSnapshot).toMatchSnapshot();
 });
 
+it('<Button> should render as button by default', () => {
+  const tag = 'a';
+  const { container } = render(<Button tag={tag}>test</Button>);
```
franciscrispin: We shouldn't pass any tag to `Button` and the assertion should be:

`expect(container.querySelector(tag).tagName).toBe('BUTTON');`

franciscrispin:
Oh sorry one more, there should be no underline below the text on hover/active/focus. This might affect the other button variants too. 

image

franciscrispin:
@westwood846 LGTM!

PR #245, opened 2020-02-28
Modal: Convert to functional component

### LINKED ISSUE #244: Convert Modal to Functional Component

westwood846:
This MR converts the `Modal` component into a functional component.

I also simplifies the internal state management somewhat.

Closes #244.

### DIFF
--- a/src/Display/Modal/Modal.tsx
+++ b/src/Display/Modal/Modal.tsx
@@ -32,7 +32,14 @@ const Modal = (props: Props) => {
   const modalContentAreaRef = React.useRef(null);
 
   React.useEffect(() => {
-    if (isVisible) modalContentAreaRef.current.focus();
+    if (isVisible) {
+      // On modal open
+      modalContentAreaRef.current.focus();
+      document.body.style.overflow = 'hidden';
+    } else {
+      // On modal close
+      document.body.removeAttribute('style');
+    }
   }, [isVisible]);
 
   React.useEffect(() => {
--- a/src/Display/Modal/Modal.tsx
+++ b/src/Display/Modal/Modal.tsx
@@ -52,14 +52,12 @@ const Modal = (props: Props) => {
   // To prevent the modal from closing
   // when a mousedown event occurs inside the ModalContentArea
   // but the subsequent mouseup event occurs outside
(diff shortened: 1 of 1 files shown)


src/Display/Modal/Modal.tsx
```diff
@@ -13,144 +13,115 @@ import {
...
-            {footer !== undefined && (
-              <ModalFooter className="modal-footer">{footer}</ModalFooter>
-            )}
-          </ModalContentArea>
-        </ModalDialog>
-      </ModalContainer>
-    );
-  }
-}
+  const [mouseDownTarget, setMouseDownTarget] = React.useState<HTMLElement>(
```
tingtingdisco: I suggest using `useRef` to keep the `mouseDownTarget`, because `useState` triggers re-render. The original implementation using `this.mouseDownTarget` to keep the value was equivalent to using `useRef` in a functional component.

src/Display/Modal/Modal.tsx
```diff
@@ -13,144 +13,115 @@ import {
...
-    document.addEventListener('keydown', this.escEvent, false);
-  }
-
-  componentDidMount() {
-    this.registerEventEscapeListener();
-  }
-
-  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
-    if (nextProps.isVisible && !prevState.isOpen) {
-      document.body.style.overflow = 'hidden';
```
tingtingdisco: do we still need this `hidden` style?

src/Display/Modal/Modal.tsx
westwood846: Updated!

src/Display/Modal/Modal.tsx
westwood846: Thanks for catching this, should not have deleted it.

westwood846: Thanks for catching this, should not have deleted it!

src/Display/Modal/Modal.tsx
westwood846: Updated!

tingtingdisco:
LGTM thank u @westwood846 !

Issue #242, opened 2020-02-27
Test Automatically on NPM Publish


westwood846:
To make sure we don’t publish bad code, we should check if tests pass before publishing.

rosslo:
Should also include **Formatting** and **Linting**.

rosslo:
Automate bumping version and adding git tag

rosslo:
PR: https://github.com/glints-dev/glints-aries/pull/246

franciscrispin:
Oh could you also update the docs once the PR is merged :smile: https://github.com/glints-dev/glints-aries/wiki/Contributing#how-to-publish-a-new-version

cc @rosslo @westwood846

PR #246, opened 2020-03-01
Feature/automate publishing npm

rosslo:
# What changed
**Issue**: https://github.com/glints-dev/glints-aries/issues/242

This PR is to enhance `npm version` command to automate the publish process of `glints-aries` as follows:
1. Building
   - `npm run build`
2. Linting
   - `npm run lint`
   - converted `babel-eslint` to `typescript-eslint-parser` to support TypeScript
3. Testing
   - `npm run test`
4. Check if you are logged in to avoid publishing failed
   - `npm whoami`
5. Bump minor version and commit message
   -  By default, `npm version` automatically adds a git tag and commits a message for a new version.
   -  change the version format from `4.0.0-beta.66` to `4.0.66` to support automatic version bump
6. Publish to npm
   - trigger `npm postpublish`
7. Add git tag
   - `git push --tags`

Btw, `npm version` accepts three parameters `patch` | `minor` | `major`, we can rely on this rule https://docs.npmjs.com/about-semantic-versioning#incrementing-semantic-versions-in-published-packages to determine an appropriate version number.
And, If there is a publishment without bumping version, we can just use `npm publish` instead of `npm version`.


rosslo:
Can review this by commits.

tingtingdisco:
@rosslo is it possible to break this PR into 2 PRs, one is about the publishing and the other is about fix errors?

rosslo:
> @rosslo is it possible to break this PR into 2 PRs, one is about the publishing and the other is about fix errors?
> 
> So, the first PR is to setup and fix lint errors, and the second one is publishing automation.

Yeah, it's possible, but I don't think it's needed. Since the separated PR for publishing automation will only have >10 code changes (last two commits in this PR) and it will also need to base on  the first PR is to setup and fix lint errors.

tingtingdisco:
ok got it

package.json
```diff
@@ -11,6 +11,7 @@
     "build:es": "rm -rf es && cross-env BABEL_ENV=es babel src -d es --extensions .ts,.tsx --ignore src/**/*.test.tsx",
     "build:ts": "tsc --rootDir src -d --declarationDir lib && tsc --rootDir src -d --declarationDir es",
     "build:icon": "babel-node --extensions .ts,.tsx src/Utils/Icon/generateIcons.ts",
+    "format": "prettier --write src/**/*.{ts,tsx} stories/**/*.{ts,tsx}",
```
tingtingdisco: in what situation should we run `format`?

.stylelintrc
```diff
@@ -3,5 +3,13 @@
   "extends": [
     "stylelint-config-recommended",
     "stylelint-config-styled-components"
-  ]
+  ],
+  "rules": {
+    "no-descending-specificity": null,
+    "property-no-vendor-prefix": [
```
tingtingdisco: why add these two rules?

.eslintrc.json
```diff
@@ -1,63 +1,27 @@
...
-    "react/jsx-filename-extension": 0,
-    "react/jsx-no-target-blank": 0,
-    "react/require-default-props": 0,
-    "react/require-extension": 0,
-    "react/self-closing-comp": 0,
-    "react/sort-comp": 0,
-    "react/jsx-no-undef": ["error", { "allowGlobals": true }],
-    "jsx-a11y/no-static-element-interactions": 0,
-    "jsx-a11y/click-events-have-key-events": 0,
-    "prettier/prettier": "error"
```
tingtingdisco: how these rules being removed and added?

tingtingdisco: are the removed rules included in the `extends`?

src/Input/AutoComplete/AutoComplete.tsx
```diff
@@ -219,6 +217,8 @@ class AutoComplete extends React.Component<Props, State> {
       onChange,
       small,
       removeFloatingLabel,
+      children, // eslint-disable-line @typescript-eslint/no-unused-vars
+      value, // eslint-disable-line @typescript-eslint/no-unused-vars
```
tingtingdisco: 🤷‍♂

.eslintrc.json
tingtingdisco: never mind, as long as the code passed lint I think it's ok to update rules

src/Input/AutoComplete/AutoComplete.tsx
rosslo: I was feeling 🤷‍♂too when I did this haha.

If we just delete this too variable, it will be included in `defaultProps` below them which will be passed to `AutoCompleteInput`, and thus it causes the bugs like overwrite the `value={selectedValue}`, `input tag element should not have a children` something like this.

I was thinking another soultion as below, but I just think the current way is simpler.
```
const autoCompleteInputProps = Object.assign({}, defaultProps);
delete autoCompleteInputProps.value;
delete autoCompleteInputProps.children;
<AutoCompleteInput
 // ...
 {...autoCompleteInputProps}
/>
```

Wdyt?

package.json
```diff
@@ -11,14 +11,15 @@
     "build:es": "rm -rf es && cross-env BABEL_ENV=es babel src -d es --extensions .ts,.tsx --ignore src/**/*.test.tsx",
     "build:ts": "tsc --rootDir src -d --declarationDir lib && tsc --rootDir src -d --declarationDir es",
     "build:icon": "rm -rf src/General/Icon/components && babel-node --extensions .ts,.tsx src/Utils/Icon/generateIcons.ts && npm run format:icon",
-    "format": "prettier --write src/**/*.{ts,tsx} stories/**/*.{ts,tsx}",
     "format:icon": "prettier --write src/General/Icon/components/*.tsx",
-    "prepublishOnly": "npm run build",
+    "prepublishOnly": "npm run lint && npm run test && npm run build && npm whoami && npm version patch -m 'Bump to %s'",
```
tingtingdisco: can we make bumping version an interactive command like this, let it default to bump patch number, but we can still type an arbitrary version?
```
// after lint, test and build
current version: x.x.x
bump version to (x.x.x+1):
```

package.json
```diff
@@ -11,14 +11,15 @@
...
+    "postpublish": "git push --tags",
```
tingtingdisco: how does this work? what tag it pushes?

src/Input/AutoComplete/AutoComplete.tsx
tingtingdisco: This is not related to this PR so just leave it as is.

.stylelintrc
rosslo: 1. We have ~30 `no-descending-specificity`'s errors, and I don't think it's a really important rule for me. For example, it will have an error for below:
```
div {
  &:disable {
     // ...
     &:hover {
       // ...
      }
  }
  &:hover {
    // ...
  }
}

```
There is no need to move `&:hover  before `&:disable` for me, I don't think it has a big impact on maintenance.
2. `property-no-vendor-prefix` is used to ignore the error for `-webkit-appearance` css attribute.

package.json
rosslo: `npm version` will create a version commit and tag automatically, so this script is to push 
 the version tag created by `npm version` to the origin.

package.json
rosslo: It was deleted in a later commit, I was using it to fix the prettier's errors.

src/Input/AutoComplete/AutoComplete.tsx
tingtingdisco: I am thinking do we really need to extend the type `ComponentPropsWithoutRef` for `Props`

.eslintrc.json
rosslo: Yeah, I guess the most important rules of them should be included in the `extends` now. We can add other rules progressively in the future.

package.json
tingtingdisco: why you keep `format:icon`?

package.json
rosslo: I was thinking use `npm version <patch | minor | major>` to bump version and automatically trigger publishing, and we still can just use `npm publish` directly to publish without  bumping version.

.stylelintrc
tingtingdisco: for 2, why we need `-webkit-appearance` in our css code? I thought `styled-components` can prefix it for us?

package.json
tingtingdisco: thank you @rosslo 🥇

package.json
rosslo: Because the Icon Components (under `src/General/Icon/components`) are automatically generated without formatting, and thus them will cause `prettier/prettier`'s error. I'd like to fomat it by script rather than setting ignore them in `prettier/prettier` rule.

tingtingdisco:
@rosslo I run `yarn lint:ts` but failed, what should I do before running it?
Screen Shot 2020-03-03 at 3 00 25 PM

.stylelintrc
rosslo: Oh yes `styled-components` should do automatic vendor prefixing for us. Will update it, thanks!

rosslo:
> @rosslo I run `yarn lint:ts` but failed, what should I do before running it?
> Screen Shot 2020-03-03 at 3 00 25 PM

This is relevant to https://github.com/glints-dev/glints-aries/pull/246#discussion_r386829675.
You can just run `format:icon` or `build:icon`(which will clean the deprecated Icon component files now) .

tingtingdisco:
> > @rosslo I run `yarn lint:ts` but failed, what should I do before running it?
> > Screen Shot 2020-03-03 at 3 00 25 PM
> 
> This is relevant to [#246 (comment)](https://github.com/glints-dev/glints-aries/pull/246#discussion_r386829675).
> You can just run `format:icon` or `build:icon`(which will clean the deprecated Icon component files now) .

OK it works, so do you think `prepublishOnly` should have `build` runs before `lint` and `test`?

rosslo:
> OK it works, so do you think `prepublishOnly` should have `build` runs before `lint` and `test`?

In practice, there is another script `prepare` might be suitable for building.
https://docs.npmjs.com/misc/scripts
> prepare: Run both BEFORE the package is packed and published, on local npm install without any arguments, and when installing git dependencies (See below). This is run AFTER prepublish, but BEFORE prepublishOnly.

But I'm not sure if there is any concern about building package after `yarn install`. 🤔

tingtingdisco:
> > OK it works, so do you think `prepublishOnly` should have `build` runs before `lint` and `test`?
> 
> Oh yeah, according to this situation, should move `build` before `lint`.
> In practice, there is another script `prepare` might be suitable for `build`.
> https://docs.npmjs.com/misc/scripts
> 
> > prepare: Run both BEFORE the package is packed and published, on local npm install without any arguments, and when installing git dependencies (See below). This is run AFTER prepublish, but BEFORE prepublishOnly.
> 
> But I'm not sure if there is any concern about building package after `yarn install`. 🤔

I think it's ok to run `build` after `install`. It's also ok to just change the order in `prepublishOnly`.

package.json
tingtingdisco: ok, just provide a way to let us bump version manually, and add instructions in wiki.

westwood846:
Fixes #249

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "glints-aries",
-  "version": "4.0.0-beta.66",
+  "version": "4.0.66",
```
westwood846: I had a conversation with @franciscrispin about this a while ago: We kinda agreed to keep the pre-releases for the time being:

```
marco 6:17
What’s up with the version numbers anyway?
6:18
How long is this pre-release thing going to go? ^^

francis  6:19 PM
I was planning to officially release and do away with the pre-releases once testing was completed haha. Don't feel confident releasing it without solid test coverage :sweat_smile:

marco  6:19 PM
At this point, what’s the difference?^^

francis  6:19 PM
As for the version numbers we can just keep incrementing it while it is in beta
6:22
Hmm no difference now that I think about it. I think at that point I still considered it a "public" library so I wanted to wait for the test coverage to be 100% before releasing haha
```

Though I still think it doesn't matter too much. If the pre-release thing is a problem for the script let's switch to regular semvers.

PR #247, opened 2020-03-02
Tabs: Update the css to fix styling of the tabs button

ritikamotwani:
Fix the styling of the tabs.
The padding and the width are not showing effect due to the missing `;`

image


westwood846:
LGTM

Issue #232, opened 2020-02-18
Components that don't have className property


aaa59891:
this issue records components that don't have className property, that lacks the freedom to add className and loses the ability to use styled lib to customize the css

1. Tabs.Pane
2. Dropdown.Item

PR #248, opened 2020-03-08
Tabs.Pane and Dropdown.Item: Add the classname property to these components

ritikamotwani:
Fixes: #232 

For the following components, on adding className="class-name" the property wasn't set and the style couldn't be manipulated through it:
1. Tabs.Pane
2. Dropdown.Item

### DIFF
--- a/src/Display/Tabs/Tabs.tsx
+++ b/src/Display/Tabs/Tabs.tsx
@@ -53,13 +53,15 @@ const Tabs: Tabs = ({
             children,
             (data: React.ReactElement<TabPaneProps>, index) => {
               const tabLabel = data.props.label || index;
+              const tabClassName = data.props.tabClassName || '';
               return (
                 <li
                   className={classNames(
                     `tab-${tabLabel}`,
                     { active: activeTabOrIndex === tabLabel },
                     `${alignment}-tab`,
-                    `${variant}`
+                    `${variant}`,
+                    `${tabClassName}`
                   )}
                   key={data.props.tab}
                   role="tab"
--- a/src/Display/Tabs/TabPane.tsx
+++ b/src/Display/Tabs/TabPane.tsx
@@ -17,6 +17,7 @@ export interface Props
   children: React.ReactNode;
   tab?: string;
   label?: string | number;
+  tabClassName?: string;
 }
 
 export default TabPane;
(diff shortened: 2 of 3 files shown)


src/Display/Tabs/Tabs.tsx
```diff
@@ -53,13 +53,15 @@ const Tabs: Tabs = ({
             children,
             (data: React.ReactElement<TabPaneProps>, index) => {
               const tabLabel = data.props.label || index;
+              const tabClassName = data.props.className || '';
```
tingtingdisco: @ritikamotwani the `data.props.className` is the `className` passed to `<Tabs.Pane />` right? So it will also apply to panel, could you confirm this? Seems like you want to apply the `className` to tab, so maybe add a `tabClassName` prop to `TabPane`?

tingtingdisco: hmm, I think add a `tabClassName` prop to this `<Tabs />` component is better.

tingtingdisco: what do you think?

ritikamotwani: @tingtingdisco the structure for the Tabs is
```
<Tabs alignment={} className="optional">
<Tabs.Pane tab="name" className="optional-pane"> Body </Tabs.Pane>
<Tabs.Pane tab="Without classname" > Body </Tabs.Pane>
</Tab>
```

So currently the classNames to Tabs do get reflected to the Tabs component, but the classNames to Tabs.Pane are not getting reflected. They shouldn't be reflected to the entire Tab but only to that particular pane.
We'll have to send from here to the Tab.Pane Component just like the tab name is being sent. The thing is the children are getting iterated in this file and then individual ones are getting sent to Tab.Pane component.

Why would I add the tabClassName to Tabs component, they are not related. Tabs has it's own className :thinking:

tingtingdisco: Tabs' className is used to style `TabsContainer`
my concern is when we pass className like this `<Tabs.Pane className="optional-pane">`, I would expect the className is applied to `TabsContent` not the `Tab` where we click on.

ritikamotwani: I think this className is for the Pane only not the content. Because for the content you have to already pass a component inside, you can easily add a className for your content. But currently we cannot set our own className for the Pane we're clicking on.

tingtingdisco: yes, the className is actually being passed to TabsContent in the TabPane component, but the Tabs component is not doing it, it's quite confusing.

So I think we should not make it more confusing by using this className to style `tab`.

tingtingdisco:
LGTM, thank u @ritikamotwani !

PR #252, opened 2020-03-13
Select: Convert to functional component

### LINKED ISSUE #251: Convert Select to Functional Component

westwood846:
Fixes #251.

### DIFF
--- a/src/Input/Select/Select.test.tsx
+++ b/src/Input/Select/Select.test.tsx
@@ -36,9 +36,10 @@ const SelectComponent = (
 );
 
 function setupOpenSelectMenu() {
-  const { getByRole, getAllByRole } = render(SelectComponent);
+  const { getByRole, getAllByRole, getByTestId } = render(SelectComponent);
   const selectInput = getByRole('combobox') as HTMLInputElement;
   const selectList = getByRole('listbox');
+  const selectLabel = getByTestId('select-label');
   const displayedOptions = getAllByRole('option');
 
   fireEvent.focus(selectInput);
@@ -46,6 +47,7 @@ function setupOpenSelectMenu() {
   return {
     selectInput,
     selectList,
+    selectLabel,
     displayedOptions,
   };
 }
@@ -68,7 +70,12 @@ function setupSelectOptionFromMenu() {
 }
 
 function setupSelectOptionFromMenuWithKeyHandling() {
-  const { selectInput, selectList, displayedOptions } = setupOpenSelectMenu();
+  const {
+    selectInput,
+    selectList,
+    selectLabel,
+    displayedOptions,
+  } = setupOpenSelectMenu();
   const randomIndex = Math.floor(Math.random() * displayedOptions.length);
   const randomOption = displayedOptions[randomIndex];
   const arrowDownEvent = { key: 'ArrowDown', keyCode: 40 };
@@ -85,6 +92,7 @@ function setupSelectOptionFromMenuWithKeyHandling() {
   return {
     selectInput,
     selectList,
+    selectLabel,
     randomOption,
   };
 }
@@ -210,6 +218,12 @@ describe('when the enter key is pressed on an option', () => {
   it('should close the option menu', () => {
     const { selectList } = setupSelectOptionFromMenuWithKeyHandling();
     expect(selectList.hasAttribute('open')).toBe(false);
+  });
+
+  it('should float the label ', () => {
+    const { selectLabel } = setupSelectOptionFromMenuWithKeyHandling();
+    const style = window.getComputedStyle(selectLabel);
+    expect(style.transform).not.toEqual('');
   });
 });
 
--- a/src/Input/Select/Select.test.tsx
+++ b/src/Input/Select/Select.test.tsx
@@ -222,8 +222,7 @@ describe('when the enter key is pressed on an option', () => {
 
   it('should float the label ', () => {
     const { selectLabel } = setupSelectOptionFromMenuWithKeyHandling();
-    const style = window.getComputedStyle(selectLabel);
-    expect(style.transform).not.toEqual('');
+    expect(selectLabel).toHaveStyle('transform: translate3d(-15px,-20px,0);');
   });
 });
 
--- a/src/Input/Select/Select.tsx
+++ b/src/Input/Select/Select.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
-import { get, isEqual } from 'lodash';
+import { get } from 'lodash';
 
 import classNames from 'classnames';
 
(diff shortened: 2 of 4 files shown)


src/Input/Select/Select.tsx
```diff
@@ -15,346 +15,271 @@ import {
...
       }
+    },
+    [children, onChange]
+  );
+
+  // Should be called when the user selects an option
+  const handleSelect = React.useCallback(() => {
+    const activeElement = getActiveElement();
+    const selectedValue = activeElement.innerText;
+    setSelectedValue(selectedValue || '');
```
rosslo: Should we have an empty string as a fallback?

src/Input/Select/Select.tsx
```diff
@@ -15,346 +15,271 @@ import {
...
-          <SelectErrorDefault className="select-error-default">
-            {error}
-          </SelectErrorDefault>
-        );
+      if (e.keyCode === 38 && cursor > 0) {
+        setCursor(cursor - 1);
+        scrollToActiveElement();
+      } else if (e.keyCode === 40 && cursor < filterValue.length - 1) {
+        setCursor(cursor + 1);
+        scrollToActiveElement();
```
rosslo: Saw the original usage is to call `scrollToActiveElement` after `cursor` was updated.

But the current one seems to call them concurrently.

rosslo:
@westwood846 Could you rebase this PR onto the latest staging? Since I saw this PR uses some hooks which might cause the errors of `react-hooks/rules-of-hooks`. 🙏

src/Input/Select/Select.tsx
```diff
@@ -15,346 +15,271 @@ import {
...
-    const { cursor, filterValue } = this.state;
-    const inputElement = e.target as HTMLInputElement;
-
-    if (disableTyping) {
-      e.preventDefault();
+    const activeElementIndex = Number(get(activeElement, 'dataset.id'));
+    const activeChild = filterValue[activeElementIndex];
+    const onOptionClick = get(activeChild, 'props.onOptionClick');
+    if (onOptionClick !== undefined) {
+      onOptionClick(selectedValue);
```
rosslo: I saw we were passing `event` as the parameter of `onOptionClick`. 

Did you check the usage of it in the `marketplace` and `employers` projects?

src/Input/Select/Select.tsx
```diff
@@ -15,346 +15,271 @@ import {
...
+        childrenArray.filter(data =>
           data.props.children
             .toLowerCase()
             .includes(e.target.value.toLowerCase())
-        ),
-        cursor: 0,
-      });
-
-      if (onChange !== undefined) {
-        return onChange(e);
```
rosslo: same as https://github.com/glints-dev/glints-aries/pull/252#discussion_r393438409

src/Input/Select/Select.tsx
westwood846: Yeah this seems to have no effect, I'm removing it.

src/Input/Select/Select.tsx
westwood846: Good catch, I wasn't familiar with the second `setState` argument and didn't realise that this should be sequential. I added an `effect` that triggers the scroll _after_ `cursor` updates.

src/Input/Select/Select.tsx
westwood846: Re-added it, shouldn't have been removed.

src/Input/Select/Select.tsx
westwood846: The funny thing is, implementations seem to depend on the changed value being returned here. For example here: 
image (app/modules/UgcQnA/Components/QuestionForm/TopicField.js in DST)

westwood846:
> @westwood846 Could you rebase this PR onto the latest staging? Since I saw this PR uses some hooks which might cause the errors of `react-hooks/rules-of-hooks`. 🙏

Rebased and fixed!

src/Input/Select/Select.tsx
rosslo: Yes, it will be handy if onChange and onOptionClick can return the changed value only.

Do you think we should refactor the use of Select on dst and employers for ^?

rosslo:
Bug: Use the keyboard to select option, the floating label doesn't work well.
image

src/Input/Select/Select.tsx
```diff
@@ -15,346 +15,287 @@ import {
...
+      } else if (e.keyCode === 27) {
+        inputElement.blur();
+        setIsFocus(false);
+      }
+    },
+    [disableTyping, filterValue, cursor, handleSelect]
+  );
+
+  React.useEffect(() => {
+    if (cursor && shouldScrollToCursor) {
```
rosslo: I think there must be a condition is `cursor` is zero and `shouldScrollToCursor ` is true. You can test with long options, you won't be able to scroll to first options.

src/Input/Select/Select.tsx
```diff
@@ -15,346 +15,287 @@ import {
...
+  const getActiveElement = React.useCallback(() => {
+    return node.current.querySelector('.active') as HTMLLIElement;
+  }, []);
+
+  // Should be called when the user selects an option
+  const handleSelect = React.useCallback(
+    (e?: void) => {
+      const activeElement = getActiveElement();
+      const selectedValue = activeElement.innerText;
+      // Need  || '' here to ensure that the select input is never uncontrolled,
```
rosslo: Hmm, when will `selectedValue` be nil?

If it is defensive programming, I think we can just make `<SelectInput value={selectedValue || ''} // ... />` instead, then it might be safer.

src/Input/Select/Select.tsx
westwood846: I tested agains some sample DST usages. I think the current state of the PR preserves the previous behaviour. As evidenced by the screenshot above, the `onChange` component did return the changed value.

src/Input/Select/Select.tsx
westwood846: No this is some weird issue with React. Because you cannot change from uncontrolled to controlled component. `selectedValue` is undefined at some point before the user types, so the underlying select input would be interpreted as uncontrolled first. This simply makes sure that that never happens.

rosslo: Yeah, I knew the point for controlled component, but after I played around with `Select` I still cannot understand how `selectedValue` can be `undefined` 🤔

westwood846: It happens in the tests. There are seven test cases were I could find this issue. They are the ones who use either `setupSelectOptionFromMenuWithKeyHandling` or `setupSelectOptionFromMenu`. Not quite sure why this is.

westwood846: Okay I found out that the `innerText` is sometimes undefined. `textContent` works better here. That's what's used by the tests too.

westwood846: > Bug: Use the keyboard to select option, the floating label doesn't work well.
> Btw, would you mind adding a test case to catch this haha?
> image

Fixed

src/Input/Select/Select.tsx
westwood846: Removed!

westwood846:
>Bug: Use the keyboard to select option, the floating label doesn't work well.
>Btw, would you mind adding a test case to catch this haha?

Fixed the bug and added the test. I couldn't figure out how to test a component with snapshots after interaction so I'm just manually checking the `transform` style here. Not very elegant but gets the job done since the test will fail if any inadvertent changes are made to the floating behaviour.

src/Input/Select/Select.test.tsx
```diff
@@ -211,6 +219,12 @@ describe('when the enter key is pressed on an option', () => {
     const { selectList } = setupSelectOptionFromMenuWithKeyHandling();
     expect(selectList.hasAttribute('open')).toBe(false);
   });
+
+  it('should float the label ', () => {
+    const { selectLabel } = setupSelectOptionFromMenuWithKeyHandling();
+    const style = window.getComputedStyle(selectLabel);
+    expect(style.transform).not.toEqual('');
```
rosslo: I think we can use `toHaveStyle` instead.
```
expect(selectLabel).toHaveStyle('transform: translate3d(-15px,-20px,0);');
```

src/Input/Select/Select.tsx
```diff
@@ -15,346 +15,283 @@ import {
...
+
+  const getActiveElement = React.useCallback(() => {
+    return node.current.querySelector('.active') as HTMLLIElement;
+  }, []);
+
+  // Should be called when the user selects an option
+  const handleSelect = React.useCallback(
+    (e?: React.ChangeEvent<HTMLInputElement>) => {
+      const activeElement = getActiveElement();
+      // const selectedValue = get(e, 'target.value', activeElement.textContent);
```
rosslo: It seems you forgot to remove it.

src/Input/Select/Select.test.tsx
westwood846: Thanks for the tip! Implemented this :)

src/Input/Select/Select.tsx
westwood846: Yep! ^^

Issue #239, opened 2020-02-25
Convert TextField to Functional Component


rosslo:
PR: https://github.com/glints-dev/glints-aries/pull/254

PR #254, opened 2020-03-15
Convert TextField to functional component

rosslo:
Issue: https://github.com/glints-dev/glints-aries/issues/239

This conversion also refactored the TextField to be a pure controlled component by removing the defaultValue prop. So before publishing this change, we should update all defaultValue to value of TextField in Marketplace and Employers projects.

For example:
```
// in app/components/GlintsJobSearchFilter/FilterOption/FilterSalary.js
<SalaryInput
    type="number"
    onChange={this.makeSalaryHandler(
    DATA_TYPE.maxSalary,
        this.props.setMaxSalary
    )}
    label="Maximum Salary"
    defaultValue={salaryFilter.maximum} // should update to value={salaryFilter.maximum}
/>
```

### DIFF
--- a/src/Input/TextField/TextField.tsx
+++ b/src/Input/TextField/TextField.tsx
@@ -10,132 +10,93 @@ import {
   TextFieldLabel,
 } from '../../Style/Input/TextFieldStyle';
 
-class TextField extends React.Component<Props, State> {
-  constructor(props: Props) {
-    super(props);
-    const { type } = this.props;
+export const isEmpty = (type: textFieldType, value: any) => {
+  if (value === undefined || value === null) return true;
+  if (type === 'number' && isNaN(value)) return true;
+  return value === '';
+};
 
-    this.state = {
-      floating: false,
-      inputType: type,
-    };
-  }
+const TextField: React.FunctionComponent<Props> = props => {
+  const {
+    value,
+    defaultValue,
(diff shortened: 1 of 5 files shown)


src/Input/TextField/TextField.tsx
```diff
@@ -10,132 +10,93 @@ import {
...
-    onBlur: (e: React.FocusEvent<HTMLInputElement>) => void
-  ) => {
-    const listener = (e: React.FocusEvent<HTMLInputElement>) => {
-      this.setState({
-        floating: e.target.value.length > 0,
-      });
-
-      if (onBlur !== undefined) {
-        return onBlur(e);
+export const isEmpty = (type: textFieldType, value: any) => {
```
tingtingdisco: can we have a `isFilled` instead of `isEmpty`? 
So this will be easier to read `setFloating(isFilled(inputType, checkedValue));`

tingtingdisco: this is minor.

tingtingdisco:
is there a way we can have backward compatibility of `defaultValue`? LGTM basically.

rosslo:
@tingtingdisco updated, please take a look again. 🙏

tingtingdisco:
LGTM, THANK U @rosslo

Issue #214, opened 2020-01-17
Modal children should only render if modal is open


westwood846:
OKR 2.e

westwood846:
Closed by #345

PR #256, opened 2020-03-16
Feature/render visible modal only

tingtingdisco:
#214 Modal children should only render if modal is open

### DIFF
--- a/src/Display/Modal/Modal.test.tsx
+++ b/src/Display/Modal/Modal.test.tsx
@@ -37,17 +37,17 @@ function setupModal(isVisible: boolean) {
       <p>{props.content}</p>
     </Modal>
   );
-  const { getByTestId, getByRole } = render(ModalComponent);
+  const { queryByTestId, queryByRole } = render(ModalComponent);
 
   return {
-    modalContainer: getByTestId('modal-container') as Element,
-    modalDialog: getByRole('dialog') as Element,
-    closeButton: getByTestId('close-button') as Element,
+    modalContainer: queryByTestId('modal-container') as Element,
+    modalDialog: queryByRole('dialog') as Element,
+    closeButton: queryByTestId('close-button') as Element,
     onClose,
   };
 }
 
-it('<Modal> should render with a title, content, footer and an onClick handler', () => {
+test('<Modal> should match snapshot', () => {
   const ModalSnapshot = renderer.create(OpenedModal).toJSON();
   expect(ModalSnapshot).toMatchSnapshot();
 });
@@ -84,9 +84,9 @@ describe('when modal is opened', () => {
 });
 
 describe('when modal is closed', () => {
-  it('should not be visible', () => {
+  it('should not be in the dom', () => {
     const { modalContainer } = setupModal(false);
-    expect(modalContainer).toHaveStyle('visibility: hidden');
+    expect(modalContainer).not.toBeInTheDocument();
   });
 });
 
@@ -121,8 +121,8 @@ describe('onClose should be called once when:', () => {
 
 describe('onClose should not have been called when:', () => {
   it('escape key is pressed on closed modal', () => {
-    const { modalContainer, onClose } = setupModal(false);
-    fireEvent.keyDown(modalContainer, escapeEvent);
+    const { onClose } = setupModal(false);
+    fireEvent.keyDown(document, escapeEvent);
     expect(onClose).toHaveBeenCalledTimes(0);
   });
 });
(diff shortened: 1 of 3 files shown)


src/Display/Modal/Modal.tsx
```diff
@@ -79,15 +126,17 @@ const Modal = (props: Props) => {
...
     <ModalContainer
       data-testid="modal-container"
       className={classNames('aries-modal', className)}
       centering={centering}
-      isOpen={isVisible}
+      isOpen={isOpen}
       removeAnimation={removeAnimation}
       onMouseDown={handleMouseDown}
       onClick={handleClick}
+      ref={modalContainerRef}
```
rosslo: This seems to be useless?

src/Display/Modal/Modal.tsx
```diff
@@ -29,22 +37,61 @@ const Modal = (props: Props) => {
...
+    }
+  }, [isVisible, mountingState]);
+
+  // UNMOUNT_START set isOpen to false, triggers css transition
   React.useEffect(() => {
-    if (isVisible) {
+    if (!isVisible && mountingState === MountingState.VISIBLE) {
+      setMountingState(MountingState.UNMOUNT_START);
+    }
+  }, [isVisible, mountingState]);
```
rosslo: Should we combine those 3 `useEffect` above? Since they have the same dependency.

src/Display/Modal/Modal.test.tsx
```diff
@@ -168,3 +168,44 @@ it('should center the Modal vertically when centering is true', () => {
...
+      onClose={props.onClose}
+    >
+      <p>{props.content}</p>
+    </Modal>
+  );
+
+  const { queryByTestId, rerender } = render(
+    <ModalComponent isVisible={false} />
+  );
+  rerender(<ModalComponent isVisible={true} />);
```
rosslo: Should check `expect(queryByTestId('modal-container')).not.toBeInTheDocument();` before this line?

src/Display/Modal/Modal.tsx
```diff
@@ -29,22 +37,61 @@ const Modal = (props: Props) => {
...
+
+  // unmount Modal after fade out transition finished
+  const onTransitionEnd = React.useCallback(() => {
+    if (!isVisible) {
```
rosslo: Maybe `mountingState  === MountingState.UNMOUNT_START`  is more suitable for this condition?

src/Display/Modal/Modal.tsx
tingtingdisco: yes, it's useless

src/Display/Modal/Modal.tsx
tingtingdisco: Do you think make them as one effect will be easier to read?

src/Display/Modal/Modal.test.tsx
tingtingdisco: the test description is `<Modal> should be visible when isVisible changes from false to true`, so I didn't check it, and I think there is already a test checked it. What do you think?

src/Display/Modal/Modal.tsx
tingtingdisco: let me try

tingtingdisco: ok

src/Display/Modal/Modal.test.tsx
tingtingdisco: but I can add

src/Display/Modal/Modal.tsx
tingtingdisco: ok

tingtingdisco:
updated @rosslo thank u~~

rosslo:
@tingtingdisco LGTM!

rosslo:
Hi @tingtingdisco, it appears that we can just use the following to instead `MountingState`. 🤔 

```
// ...
    <ModalContentArea // ... >
          {isVisible && (
            <>
              {!hideHeader && (
                <ModalHeader className="modal-header">
                  <h3>{title}</h3>
                  <button
                    data-testid="close-button"
                    type="button"
                    onClick={onClose}
                  >
                    <CloseIcon color={hideContentArea ? 'white' : 'grey'} />
                  </button>
                </ModalHeader>
              )}
              <ModalBody
                className="modal-body"
                hideContentArea={hideContentArea}
              >
                {children}
              </ModalBody>
              {footer !== undefined && (
                <ModalFooter className="modal-footer">{footer}</ModalFooter>
              )}
            </>
          )}
    </ModalContentArea>
// ...
```

tingtingdisco:
I don't understand. We also don't want to render modal container.

tingtingdisco:
To unmount modal after css transition, we need extra states

src/Display/Modal/Modal.tsx
```diff
@@ -81,6 +81,10 @@ const Modal = (props: Props) => {
       // On modal close
       document.body.removeAttribute('style');
     }
+
+    return () => {
+      document.body.removeAttribute('style');
```
rosslo: According to your commit message, I think we should add `if (mountingState === MountingState.UNMOUNTED)` here.

Hmm, but why, shouldn't line 82 do the same thing?

tingtingdisco: hmm, in our feature, we wrap Modal in a Component, the Modal is used to delete the Component. When the component is unmounted, the Modal's css transition is not finished yet, but it is being unmounted together with the Component.

Because of this, I want to close this PR, I think it's better to let this Modal component have minimum logic.

PR #264, opened 2020-03-23
NPM: push version commit & check branch is v4

rosslo:
Before `glints-aries` is published, we should:
- ensure the branch is `v4`
- pull the latest commit

After `glints-aries` is  published, we should:
- push the commit which changes the version field of package.json to GitHub.


package.json
```diff
@@ -11,14 +11,15 @@
...
     "format:icon": "prettier --write src/General/Icon/components/*.tsx",
     "lint": "npm run lint:css && npm run lint:ts",
     "lint:css": "stylelint src/Style/**/*.ts",
     "lint:ts": "eslint src/**/*.{ts,tsx}",
-    "preversion": "npm whoami && npm run validate",
+    "preversion": "npm run check:branch && git pull && npm whoami && npm run validate",
     "postversion": "npm publish",
-    "prepublishOnly": "npm whoami && npm run validate",
-    "postpublish": "git push origin --tags",
+    "prepublishOnly": "npm run check:branch && git pull && npm whoami && npm run validate",
```
tingtingdisco: why preversion and  prepublishOnly do similar things?

rosslo: Prevent someone who doesn't use `npm version` script to do publish from skipping the validation process.

tingtingdisco: how do you prevent `npm publish` without an updated version in package.json?

tingtingdisco: ah, then publish will not work because version already exist.

rosslo: `npm` will deny the publish without version change.

tingtingdisco: please add instruction to wiki

package.json
```diff
@@ -11,14 +11,15 @@
...
+    "postpublish": "git push && git push --tags",
```
tingtingdisco: why we push `postpublish`

rosslo: you mean should do in `postversion`?

`postpublish` will can help us to ensure the publish is done, then we can do git push without worries about the git tag is in a wrong commit.

tingtingdisco: I think we should push right after `npm version` so maybe postversion?

rosslo: Do you remember if the git tag confit occurs in `git push --tags`? If so, we should do it in `postversion`.

tingtingdisco: maybe we should just have one standard way to do publish:
step 1. `npm version`
step 2. `npm publish`

rosslo: Yes `git push --tags` should not do in `postpublish` becuase those git change is made by `npm version`

PR #268, opened 2020-04-01
Scripts: Add count-imports scripts

westwood846:
This script takes the path to a repo as its only argument and then runs it through an AST parser to count the number of Aries components that are being imported by the repo. 

This is a better alternative to just counting the number of import statements with a regex because that fails to count new imports that are being made in files which already had an Aries import.


package.json
```diff
@@ -98,9 +101,12 @@
     "file-loader": "^3.0.1",
     "fork-ts-checker-webpack-plugin": "^1.3.4",
     "generate-changelog": "^1.7.1",
+    "glob": "^7.1.6",
     "jest": "^24.8.0",
     "jest-styled-components": "^6.3.1",
     "jsx-to-string": "^1.4.0",
+    "lodash": "^4.17.15",
+    "mocha": "^7.1.1",
```
franciscrispin: Can we do the test with just `jest`?

westwood846: Oh yes, can, this was not supposed to be there anymore ^^

scripts/count-imports/index.js
```diff
@@ -0,0 +1,50 @@
...
+
+const { ariesImportIdentifiers } = require('./filters');
+
+const cwd = process.argv[2];
+if (!cwd) throw 'Please pass the repo to count in as an argument';
+
+// Generates a list of all .js files in the given directory
+const globFiles = (cwd, callback) => {
+  return glob(
+    `**/*.js`,
```
rosslo: Should exclude test files and include TS files since @franciscrispin is migrating TypeScript for dst.
For example:
```
'**/!(*.spec|*.d).{ts,tsx,js,jsx}'
```

scripts/count-imports/parser.js
```diff
@@ -0,0 +1,27 @@
...
+const options = {
+  sourceType: 'module',
+  allowImportExportEverywhere: false,
+  allowReturnOutsideFunction: false,
+  ranges: false,
+  tokens: false,
+  plugins: [
+    'asyncGenerators',
+    'classProperties',
+    // "decorators",
```
rosslo: would you like to remove or keep it?

scripts/count-imports/parser.js
```diff
@@ -0,0 +1,27 @@
...
+    'doExpressions',
+    'exportExtensions',
+    'flow',
```
rosslo: should use `typescript` instead.

scripts/count-imports/parser.js
```diff
@@ -0,0 +1,27 @@
...
+    'functionSent',
+    'functionBind',
+    'jsx',
+    'objectRestSpread',
+    'dynamicImport',
+    'numericSeparator',
+    'optionalChaining',
+    'optionalCatchBinding',
```
rosslo: can consider removing the followings unneeded plugins
```
decorators,
doExpressions,
functionSent,
numericSeparator,
optionalChaining
```

rosslo:
Could we have those new files in TypeScript? I just want to keep consistency. 😅

westwood846:
> Could we have those new files in TypeScript? I just want to keep consistency. 😅

Thought about that myself but then I decided that we might wanna keep scripts like this one in JS so users don't have to transpile before running. There's [ts-node](https://github.com/TypeStrong/ts-node) for this but then this is yet another dependency.

rosslo:
> ts-node

Can refer to script `build:icon` which use `babel-node` to transpile the `generateIcons.ts`

scripts/count-imports/index.js
westwood846: Done!

scripts/count-imports/parser.js
westwood846: Can remove.

scripts/count-imports/parser.js
westwood846: Good point, thanks.

scripts/count-imports/parser.js
westwood846: Removed!

westwood846:
> > ts-node
> 
> Can refer to script `build:icon` which use `babel-node` to transpile the `generateIcons.ts`

Converted! Can you please help me check? I'm not very familiar with converting to TS.

scripts/count-imports/index.ts
```diff
@@ -0,0 +1,52 @@
+// Usage: node index.js path/to/directory
+// Counts the number of imported Aries components in a given directory
```
rosslo, line 2: Might need to update comment here. 🙏

rosslo:
Can rename `filters.spec.ts` to `filters.test.ts` to have the same suffix as the others'. 
Sorry for being picky. 😅

scripts/count-imports/index.ts
westwood846: Updated!

westwood846:
> Can rename `filters.spec.ts` to `filters.test.ts` to have the same suffix as the others'.

Renamed!


> Sorry for being picky. 😅

Actually I'm glad for your suggestions, they have all been very sensible :)

rosslo (approved):

PR #269, opened 2020-04-02
Upgrade Packages

westwood846:
This PR upgrades all packages (via `yarn upgrade`). It also fixes the version of [`kind-of` ](https://www.npmjs.com/package/kind-of) to `^6.0.3` because of a [security vulnerability](https://github.com/glints-dev/glints-aries/network/alert/yarn.lock/kind-of/open).

There where some subtle API changes in some dependencies which made it necessary to fix some tests and update their snapshots. Also the storybooks knobs were breaking so I decided to remove them, since they are not being used anyway. Additionally, I'm removing storybook actions which have not been utilised much as well. This gets rid of the annoying knobs storybook panel.


package.json
```diff
@@ -101,6 +97,7 @@
     "jest": "^24.8.0",
     "jest-styled-components": "^6.3.1",
     "jsx-to-string": "^1.4.0",
+    "kind-of": "^6.0.3",
```
rosslo: Hmm, I'm not sure if we need this to resolve the alert. Since after `yarn upgrade`, there is no error when `yarn audit`.

src/Input/Select/Select.test.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from 'react';
 import * as lodash from 'lodash';
 import * as renderer from 'react-test-renderer';
 import '@testing-library/jest-dom/extend-expect';
-import { fireEvent, render } from '@testing-library/react';
+import { fireEvent, render, getAllByTestId } from '@testing-library/react';
```
rosslo: eslint warning: 'getAllByTestId' is defined but never used

rosslo: Btw, would you mind to resolve this eslint error within this PR? 😅 
I think it occurred due to the upgrade of dependencies.

```
/src/Display/Slider/Slider.tsx
  151:3  error  componentWillReceiveProps is deprecated since React 16.9.0, use UNSAFE_componentWillReceiveProps instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops. Use https://github.com/reactjs/react-codemod#rename-unsafe-lifecycles to automatically update your components  react/no-deprecated
```

westwood846: >eslint warning: 'getAllByTestId' is defined but never used

Removed.

> Btw, would you mind to resolve this eslint error within this PR?

Changed to `UNSAFE_componentWillReceiveProps`. The component will eventually be converted to a functional component, which will replace this deprecated lifecycle method.

package.json
westwood846: Removed, thanks!

rosslo (approved):

PR #279, opened 2020-04-06
Dropdown: Add test

### LINKED ISSUE #271: Write Tests for Dropdown

tingtingdisco:
Closes #271

### DIFF
--- a/src/Navigation/Dropdown/Dropdown.test.tsx
+++ b/src/Navigation/Dropdown/Dropdown.test.tsx
@@ -308,6 +308,23 @@ describe('<Dropdown/> keydown event', () => {
     fireEvent.keyDown(document, { key: 'Esc', keyCode: 27 });
     expect(dropdownContent).not.toBeVisible();
   });
+
+  test('handleEscKeydown listener should be removed before component unmount', async () => {
+    const removeEventListenerSpy = jest.spyOn(document, 'removeEventListener');
+    const ref = React.createRef<Dropdown>();
+    const { unmount } = render(
+      <Dropdown label="Career" ref={ref}>
+        <DropdownItem value="Product Manager">Product Manager</DropdownItem>
+        <DropdownItem value="Software Engineer">Software Engineer</DropdownItem>
+      </Dropdown>
+    );
+
+    const handleEscKeydown = ref.current.handleEscKeydown;
+    unmount();
+    expect(removeEventListenerSpy).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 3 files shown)


src/Navigation/Dropdown/Dropdown.test.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+  test('showHoverLine', () => {
+    render(
+      <Dropdown label="Career" showHoverLine>
+        <DropdownItem value="Product Manager">Product Manager</DropdownItem>
+        <DropdownItem value="Software Engineer">Software Engineer</DropdownItem>
+      </Dropdown>
+    );
+    expect(queryByRole(document.body, 'listbox')).toHaveStyle(
+      'margin-top: 13px;'
```
ritikamotwani, line 120: Is having toHaveStyle better than checking if the property of some css value exists or not? Because the styles can change according to the design. 
Say if I make margin-top 12px, I'll have to change the test too.
But then again it would be better in the sense to check if the style set is correct :thinking:

src/Navigation/Dropdown/Dropdown.test.tsx
```diff
@@ -0,0 +1,330 @@
...
+        <DropdownItem value="Software Engineer">Software Engineer</DropdownItem>
+      </Dropdown>
+    );
+
+    const dropdown = queryByRole(document.body, 'menuitem');
+    fireEvent.click(dropdown);
+    fireEvent.mouseDown(queryByText(document.body, 'Product Manager'));
+
+    expect(document.querySelector('.dropdown-listbox')).not.toBeVisible();
+    expect(
```
ritikamotwani, line 212: Could you please explain this part :P

src/Navigation/Dropdown/Dropdown.test.tsx
tingtingdisco: hmm, test cases are spec of a module, and 'margin-top: 13px;' is the current spec. When spec changes, test needs to be updated, because they are the same thing. So I think specifying `13px` is the correct way. If spec of a module changes and we don't need to update its test file, then I think the test file is not be written correctly haha.

src/Navigation/Dropdown/Dropdown.test.tsx
tingtingdisco: 1. click dropdown button, and then options is shown
2. fire mouseDown on the option `Product Manager`, which close dropdown (`.dropdown-listbox`)
3. validate `.dropdown-listbox` is not visible
4. validate the text on dropdown button is `Product Manager`.

ritikamotwani (approved):

PR #293, opened 2020-04-17
RadioButton: Add tests

### LINKED ISSUE #292: Tests: RadioButton

westwood846:
Closes #292.

### DIFF
--- a/src/Input/RadioButton/RadioButton.test.tsx
+++ b/src/Input/RadioButton/RadioButton.test.tsx
@@ -1,0 +1,105 @@
+import * as React from 'react';
+import * as renderer from 'react-test-renderer';
+import { fireEvent, render } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+
+import RadioButton from './RadioButton';
+
+const singleRadioButton = (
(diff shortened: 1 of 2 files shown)


src/Input/RadioButton/RadioButton.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+        name="job-type"
+        value="internship"
+        data-testid="radio-2"
+        onChange={onChange}
+      />
+    </React.Fragment>
+  );
+  return {
+    onChange,
+    container,
```
rosslo: `container` isn't used, can consider removing it.

rosslo:
@westwood846 Do you think we should add the test cases for `checked` and `theme` props?

src/Input/RadioButton/RadioButton.test.tsx
westwood846: Removed!

westwood846:
> @westwood846 Do you think we should add the test cases for `checked` and `theme` props?

Added a snapshot test for theme. `checked` is actually being tested already (`twoRadioButtonsOneChecked`).

rosslo:
@westwood846 LGTM

PR #296, opened 2020-04-17
Collapsible: Convert to functional component

### LINKED ISSUE #295: Convert Collapsible to Functional Component

westwood846:
Closes #295

### DIFF
--- a/src/Display/Collapsible/Collapsible.tsx
+++ b/src/Display/Collapsible/Collapsible.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import classNames from 'classnames';
+import { isUndefined } from 'lodash';
 import { ArrowDownIcon } from '../../General/Icon/components';
 
 import {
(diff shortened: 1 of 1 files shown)


src/Display/Collapsible/Collapsible.tsx
```diff
@@ -9,72 +10,63 @@ import {
...
-    const { isOpen } = this.props;
-
-    if (isOpen === false && isOpen !== undefined) {
-      this.setState({
-        isOpen,
-      });
+  // If controlled, don't change the state on click
+  // If uncontrolled, toggle change
+  const toggle = React.useCallback(() => {
+    if (isUndefined(isOpen)) {
```
rosslo: I think it will be good for readability if we have a variable `const isUncontrolled = isUndefined(isOpen);`

src/Display/Collapsible/Collapsible.tsx
```diff
@@ -9,72 +10,63 @@ import {
...
-  componentDidMount() {
-    const { isOpen } = this.props;
-
-    if (isOpen === false && isOpen !== undefined) {
-      this.setState({
-        isOpen,
-      });
+  // If controlled, don't change the state on click
+  // If uncontrolled, toggle change
+  const toggle = React.useCallback(() => {
```
rosslo: I think `useCallback` is not needed here since it is passed to a div.

westwood846: What's the difference? I'm using `useCallback` here mostly because I'm defaulting to that instead of creating a plain function (which would get recreated on every render). But tbh while I don't know which has better performance, I'd guess it will never make a difference.

rosslo: I usually use `useCallback` for `PureComponent` or the dependency of hooks to prevent unnecessary render. 

But in this case, I think `CollapsibleContainer` will re-render every time because it's a node rather than a component. Therefore, the `useCallback` will be the extra cost.

rosslo:
@westwood846 We cannot open the `usage` on storybook haha.

image

westwood846:
> @westwood846 We cannot open the `usage` on storybook haha.

Oh bugger! Seems like it hasn't been clear in the past whether `isOpen` will determine the initial state or if it can be used to make the component controlled.

In the storybook component it could be both:
```
<Collapsible label="Usage" isOpen={false}>
  {usage
    ? this.renderUsage(usage)
    : this.renderDefaultUsage(children)}
</Collapsible>
```

While in DST's `AddPhotoField` it looks like a controlled component:
```
<Collapsible
  label={<PhotoLabel numOfImages={photos.length} />}
  isOpen={photos.length > 0}
>
```
(dst/app/modules/UgcQnA/Components/QuestionForm/AddPhotoField.js)

In DST's `AddPollField` it's definitely just an initial state.
```
<Collapsible isOpen={false} label={<Label isAttached={isAllFilled} />}>
```
(dst/app/modules/UgcQnA/Components/QuestionForm/AddPollField/AddPollField.js)

I'd suggest we clear this up by adding a `isInitiallyOpen` prop and migrating cases like the last one above. What do you think?

rosslo:
> I'd suggest we clear this up by adding a `isInitiallyOpen` prop and migrating cases like the last one above. What do you think?

Oh I think we were refactoring `Collapsible` wrong, `Collapsible` is never be implemented for controlled component, `isOpen` is just set for the initial / default `isOpen` status. As it is described `isOpen: Collapsing as default.` in the storybook.

So we will need to remove the code related to `isControlled`, then just pass `props.isOpen` to `useState` as default status.

westwood846:
> Oh I think we were refactoring `Collapsible` wrong

Yeah I see what you mean. Kinda sad but okay. Changed it so that `isOpen` just determines the initial state.

src/Display/Collapsible/Collapsible.tsx
```diff
@@ -9,72 +10,58 @@ import {
...
-  };
+const Collapsible: React.FunctionComponent<Props> = ({
+  label,
+  children,
+  className,
+  isOpen,
+  ...defaultProps
+}) => {
+  const [isOpenInternal, setIsOpenInternal] = React.useState<boolean>(
+    isUndefined(isOpen) ? true : isOpen
```
rosslo, line 21: Can simplify to `Boolean(isOpen)`.

src/Display/Collapsible/Collapsible.tsx
```diff
@@ -9,72 +10,58 @@ import {
...
 
 interface Props
   extends React.ComponentPropsWithoutRef<typeof CollapsibleContainer> {
   label: React.ReactNode;
   isOpen?: boolean;
-}
-
-interface State {
-  isOpen: boolean;
+  isInitiallyOpen?: boolean;
```
rosslo: Should remove this haha.

rosslo (approved):

PR #299, opened 2020-04-20
Tests/tooltip divider grid colors

### LINKED ISSUE #285: Tests: Colors

### LINKED ISSUE #287: Tests: Grid

### LINKED ISSUE #297: Tests: Tooltip

### LINKED ISSUE #298: Tests: Divider

tingtingdisco:
Add tests for Tooltip Divider Grid Colors
Closes #297 
Closes #298 
Closes #287 
Closes #285


src/Display/Tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,92 @@
+import * as React from 'react';
+import 'jest-styled-components';
```
tingtingdisco, line 2: I think we can include `jest-styled-components` in all tests, so css can be included in snapshot.

westwood846: I thought that the classnames generated by styled-components are hashes of the CSS? In that case the classname would change when the CSS changes, and thus the snapshot comparison would fail.

src/Layout/Grid/Grid.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+          </Grid.Col>
+          <Grid.Col xs={12} sm={3} md={9}>
+            mock content
+          </Grid.Col>
+        </Grid.Row>
+      </Grid>
+    );
+    expect(asFragment()).toMatchSnapshot();
+  });
+});
```
westwood846, line 40: I wonder if there is a way to test that the grid is wrapping it's columns properly on low width viewports?

src/Style/Colors.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  test('Greyscale', () => {
+    expect(Greyscale).toEqual({
+      black: '#000000',
+      grey: '#777777',
+      lightgrey: '#C6C6C6',
+      softgrey: '#F3F3F3',
+      white: '#FFFFFF',
+    });
+  });
+});
```
westwood846, line 57: Maybe it would make more sense to go with a snapshot test here? I think you could just do something like `expect(Greyscale).toMatchSnapshot()`. That would have the advantage that you wouldn't have to manually adjust colors when they change, just have to update the snapshot. And you'd still accomplish the goal of having failing tests when the Colors objects change.

src/Display/Tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+import '@testing-library/jest-dom/extend-expect';
+import { fireEvent, render } from '@testing-library/react';
+
+import Tooltip from './Tooltip';
+
+describe('<Tooltip/> snapshots', () => {
+  const TooltipComponent = () => (
+    <Tooltip
+      text="mock text"
+      position="right"
```
westwood846: Can we test the other positions as well? It would raise the components line-coverage to 100%.

src/General/Divider/Divider.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+describe('<Divider/> snapshots', () => {
+  test('should match snapshot when no props is being passed', () => {
+    const { asFragment } = render(<Divider />);
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  test('should match snapshot when all props is being passed', () => {
+    const { asFragment } = render(
+      <Divider theme="blue" className="name" style={{ top: '10px' }} />
```
westwood846: Can we test the other themes too? Needed for 100% coverage. May be done programatically.

westwood846 (changes requested):
Requesting some minor changes/additions.

src/Display/Tooltip/Tooltip.test.tsx
tingtingdisco: yes, my idea is that maybe we should check css in the first generated snapshot to see whether the css is correct, otherwise we can only detect css changes.

src/Layout/Grid/Grid.test.tsx
tingtingdisco: you mean the css should be different in different screen width? if so, it is the reason I included `jest-styled-components`, the css is in the snapshot, so we can check it.

src/Style/Colors.test.ts
tingtingdisco: thank u! let me try it, didn't know we can do this `expect(Greyscale).toMatchSnapshot()`

src/Display/Tooltip/Tooltip.test.tsx
tingtingdisco: ok

src/General/Divider/Divider.test.tsx
tingtingdisco: ok

tingtingdisco:
thank u @westwood846 updated the PR, please review again~

src/Display/Tooltip/Tooltip.test.tsx
westwood846: Yeah but we can't check for correctness of the CSS. We can only make sure that it does not change unexpectedly.

src/Style/Colors.test.ts
tingtingdisco: Screen Shot 2020-04-23 at 11 09 22 AM

@rosslo do you know why this color snapshot failed, does it run test against complied js file?

westwood846 (approved):
LGTM!

src/Style/Colors.test.ts
tingtingdisco: oh yes it runs the `lib/Style/Colors.test.js`

tingtingdisco: @westwood846 I will leave the color test as is, we need to change our test setting.

rosslo: Haha, I think we should not include test files in packages (lib/ es).

PR #323, opened 2020-05-05
Tag: align with GDS

### LINKED ISSUE #196: Tag: Fix UI Mismatches
GDS: https://share.goabstract.com/f57059a4-edfc-4634-a1e6-c2b586b2eacf?sha=

rosslo:
GDS Tag:
https://app.abstract.com/share/f57059a4-edfc-4634-a1e6-c2b586b2eacf?sha=

# What changed
- remove useless `theme` prop
- add `variant` prop to set the Tag's type (Basic Tag, Button Tag, Action Tag)
- add `action` prop to set the Action Tag's type
- add `icon` prop to set the Button Tag's right icon
- adjust the clickable area of`Tag`
- adjust the layout of `Tag` story

### DIFF
--- a/src/General/Tag/Tag.tsx
+++ b/src/General/Tag/Tag.tsx
@@ -11,7 +11,6 @@ const Tag: React.FunctionComponent<Props> = props => {
   const {
     className,
     children,
-    theme,
     block,
     outline,
     borderStyle,
@@ -35,7 +34,6 @@ const Tag: React.FunctionComponent<Props> = props => {
         className={classNames('aries-tag', className)}
         role="presentation"
         tabIndex={0}
-        theme={theme}
         block={block}
         outline={outline}
         borderStyle={borderStyle}
@@ -55,7 +53,6 @@ const Tag: React.FunctionComponent<Props> = props => {
 };
 
 Tag.defaultProps = {
-  theme: 'grey',
   block: false,
   outline: false,
   borderStyle: 'solid',
@@ -63,7 +60,6 @@ Tag.defaultProps = {
 
 interface Props extends React.ComponentPropsWithoutRef<typeof TagContainer> {
   children: React.ReactNode;
-  theme?: string;
   onClick?(e: React.MouseEvent<HTMLLabelElement, MouseEvent>): void;
   startIcon?: React.ReactNode;
   endIcon?: React.ReactNode;
--- a/src/General/Tag/Tag.tsx
+++ b/src/General/Tag/Tag.tsx
@@ -13,7 +13,7 @@ const Tag: React.FunctionComponent<Props> = props => {
     children,
     block,
     outline,
-    borderStyle,
+    border,
     startIcon,
     endIcon,
     onClick,
@@ -36,7 +36,7 @@ const Tag: React.FunctionComponent<Props> = props => {
         tabIndex={0}
         block={block}
         outline={outline}
-        borderStyle={borderStyle}
+        border={border}
         {...restProps}
       >
         <TagContent
@@ -55,7 +55,7 @@ const Tag: React.FunctionComponent<Props> = props => {
 Tag.defaultProps = {
   block: false,
   outline: false,
-  borderStyle: 'solid',
+  border: 'solid',
 };
 
 interface Props extends React.ComponentPropsWithoutRef<typeof TagContainer> {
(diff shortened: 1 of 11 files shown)


src/General/Tag/Tag.tsx
```diff
@@ -1,38 +1,65 @@
...
   outline: false,
+  border: 'solid',
 };
 
 interface Props extends React.ComponentPropsWithoutRef<typeof TagContainer> {
   children: React.ReactNode;
-  theme?: string;
+  onClick?(e: React.MouseEvent<HTMLDivElement, MouseEvent>): void;
+  startIcon?: React.ReactNode;
+  endIcon?: React.ReactNode;
```
tingtingdisco, line 62: why need startIcon and endIcon? they can be in the `children`?

tingtingdisco: ah I see the spec

tingtingdisco:
I don't see this is being implemented
Screen Shot 2020-05-06 at 1 43 43 PM

stories/General/TagStory.tsx
```diff
@@ -1,81 +1,294 @@
...
+    fill: ${Greyscale.black};
+  }
+\`;
+<Tag endIcon={<ClickableCloseIcon onClick={() => alert('Clicked!')} />}/>}>Add</Tag>
+`}
+      >
+        <Heading>Customized Tags</Heading>
+        <TagContainer>
+          <Tag
+            endIcon={<ClickableCloseIcon onClick={() => alert('Clicked!')} />}
```
tingtingdisco: is it possible to make the tag have `onStartIconClick` and `onEndIconClick`, so we don't need to style `ClickableCloseIcon`

src/General/Tag/Tag.tsx
tingtingdisco: shall we let users pass any startIcon  and endIcon? It seems like only 3 icons are in the design system
Screen Shot 2020-05-06 at 1 53 58 PM

stories/General/TagStory.tsx
```diff
@@ -1,81 +1,294 @@
...
+      </TagContainer>
+    </StorybookComponent>
+  );
+};
+
+const ClickableTagStory = () => {
+  const usage = `<Tag>Non-clickable tag</Tag>
+<Tag startIcon={<AddIcon onClick={() => alert('Clicked!')} />}>Clickable icon</Tag>
+<Tag endIcon={<CloseIcon onClick={() => alert('Clicked!')} />}>Clickable icon</Tag>
+<Tag onClick={() => alert('Clicked!')}>Clickable tag</Tag>`;
```
tingtingdisco: I don't see this type of clickable tag in the spec, only cursor becomes pointer when mouse is hovering?

stories/General/TagStory.tsx
```diff
@@ -1,81 +1,294 @@
...
+      <TagContainer>
+        <Tag border="none">None border</Tag>
+      </TagContainer>
+    </StorybookComponent>
+  );
+};
+
+const ClickableTagStory = () => {
+  const usage = `<Tag>Non-clickable tag</Tag>
+<Tag startIcon={<AddIcon onClick={() => alert('Clicked!')} />}>Clickable icon</Tag>
```
tingtingdisco: I don't see this type of only startIcon can be clicked in the spec.

stories/General/TagStory.tsx
rosslo: The `ClickableCloseIcon` is used to customize the icon color, it's not related to the click event.

tingtingdisco: yes I mean we should not let user customize icon color

stories/General/TagStory.tsx
rosslo: Those are the example of how to make the Tag be clickable. 
1. add onClick event on the icon you pass, then only the icon will be clickable.
2.add onClick event on the Tag directly, then the whole tag will be clickable.

rosslo:
> I don't see this is being implemented
> Screen Shot 2020-05-06 at 1 43 43 PM

Do you think we should have a `variant` prop for them?

stories/General/TagStory.tsx
rosslo: I just want to make Tag be flexible. Do you think we should limit this?

src/General/Tag/Tag.tsx
rosslo: same as https://github.com/glints-dev/glints-aries/pull/323#issuecomment-624464171

stories/General/TagStory.tsx
rosslo: If we agree that we should  a variant prop, we can define it in aries.

tingtingdisco (approved):
LGTM, nice storybook refactoring, thank u @rosslo!

PR #326, opened 2020-05-06
Fix Select Changed Behaviour

### LINKED ISSUE #313: Select Changed Behaviour in 4.0.72
It seems that I inadvertently changed the behaviour of `Select` to call `onSelect` with the label, not the value, of the selected element.

See https://glints.slack.com/archives/CBM88R7U1/p1588055556273400 for details.

As a workaround dependent projects can fix the glints-aries version to 4.0.71 but this will block them from updating until this issue is resolved.

westwood846:
Fixes #313 

When I converted `Select` into a functional component I made some mistakes which this PR reverts.

* First of all, this PR fixes the test that should have caught the changed behaviour.
* Reverts a change that made `handleChange` call `onChange` with the user input when the user types. Previously it called `onChange` with the whole change event. This prevents a new, unreported bug in the front-end project where typing into the select would trigger any `onChange` callback with the typed input. For example in employers there's a Select for changing a candidate status. Typing "Inter" (Interviewed) into the Select would send an API request for changing the status to "Inter").
* In handleSelect, use the triggering event as data source for selectedValue if passed. This function gets called with an event when the user clicks an option, and then there is no active element, so we have to get the data from the event. Again, this is reverting a mistake I made when converting the component.
* In handleSelect, call onChange with the value, not the text content of the selected option. This fixes #313. Again, reverting to before-conversion behaviour.

### DIFF
--- a/src/Input/Select/Select.test.tsx
+++ b/src/Input/Select/Select.test.tsx
@@ -186,8 +186,9 @@ describe('when it is clicked', () => {
 
 describe('when an option is clicked', () => {
   it('should call the onOptionClick function', () => {
-    setupSelectOptionFromMenu();
+    const { randomOption } = setupSelectOptionFromMenu();
     expect(props.onOptionClick).toHaveBeenCalledTimes(1);
+    expect(props.onChange).toHaveBeenCalledWith(randomOption.dataset.value);
   });
 
   it('should show the value of the option on the select input', () => {
@@ -203,8 +204,10 @@ describe('when an option is clicked', () => {
 
 describe('when the enter key is pressed on an option', () => {
   it('should call the onOptionClick function', () => {
-    setupSelectOptionFromMenuWithKeyHandling();
-    expect(props.onOptionClick).toHaveBeenCalledTimes(1);
+    const { randomOption } = setupSelectOptionFromMenuWithKeyHandling();
+    expect(props.onChange).toHaveBeenCalledWith(
+      randomOption.getAttribute('data-value')
+    );
   });
 
   it('should show the value of the option on the select input', () => {
--- a/src/Input/Select/Select.tsx
+++ b/src/Input/Select/Select.tsx
@@ -148,7 +148,7 @@ const Select: ISelect = (props: Props) => {
   // Should be called when the user selects an option
   const handleSelect = React.useCallback(
     (e?: React.ChangeEvent<HTMLInputElement>) => {
-      const activeElement = getActiveElement();
+      const activeElement = e ? e.target : getActiveElement();
       const selectedValue = activeElement.textContent;
       setSelectedValue(selectedValue);
       setFilterValue(React.Children.map(children, data => data));
--- a/src/Input/Select/Select.tsx
+++ b/src/Input/Select/Select.tsx
@@ -156,7 +156,7 @@ const Select: ISelect = (props: Props) => {
       setFloating(true);
 
       if (onChange !== undefined) {
-        onChange(selectedValue);
+        onChange(activeElement.getAttribute('data-value'));
       }
 
       const activeElementIndex = Number(get(activeElement, 'dataset.id'));
--- a/src/Input/Select/Select.tsx
+++ b/src/Input/Select/Select.tsx
@@ -135,7 +135,7 @@ const Select: ISelect = (props: Props) => {
       setCursor(0);
 
       if (onChange !== undefined) {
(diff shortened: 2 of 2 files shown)


src/Input/Select/Select.test.tsx
```diff
@@ -186,8 +186,11 @@ describe('when it is clicked', () => {
 
 describe('when an option is clicked', () => {
   it('should call the onOptionClick function', () => {
-    setupSelectOptionFromMenu();
+    const { randomOption } = setupSelectOptionFromMenu();
     expect(props.onOptionClick).toHaveBeenCalledTimes(1);
+    expect(props.onChange).toHaveBeenCalledWith(
+      randomOption.getAttribute('data-value')
```
tingtingdisco: can read this from `randomOption.dataset.value`

src/Input/Select/Select.tsx
```diff
@@ -135,7 +135,7 @@ const Select: ISelect = (props: Props) => {
       setCursor(0);
 
       if (onChange !== undefined) {
-        return onChange(e.target.value);
+        return onChange(e);
```
tingtingdisco, line 138: this is a breaking change, please update other projects

tingtingdisco: why here passing `e` to onChange, and handleSelect is passing `activeElement.getAttribute('data-value')` to onChange?

tingtingdisco: hmm, I think it's better to keep passing `e.target.value` to onChange, since the `handleSelect` can be called without a `event` object passed.

westwood846: ...so do you approve of this? If you go back in the history you'll see that the above was the behaviour before the conversion into a functional component. Not sure why they did it like that, but I thought it better to restore it.

src/Input/Select/Select.test.tsx
westwood846:
```suggestion
      randomOption.dataset.value
```

src/Input/Select/Select.tsx
tingtingdisco: please make sure you understand this change

tingtingdisco (commented):
please explain the changes and check how the changes will affect other projects.

src/Input/Select/Select.test.tsx
westwood846: Done

src/Input/Select/Select.tsx
westwood846: https://github.com/glints-dev/glints-aries/blob/2d402a4872e579a447ec274c67ce9a9053a1ea7c/src/Input/Select/Select.tsx#L162

```
  handleChange = (
    onChange: (e: React.ChangeEvent<HTMLInputElement>) => void
  ) => {
    const listener = (e: React.ChangeEvent<HTMLInputElement>) => {
      const { children } = this.props;
      const childrenArray = React.Children.toArray(children) as Array<
        React.ReactElement<SelectItemProps>
      >;

      this.setState({
        selectedValue: e.target.value,
        filterValue: childrenArray.filter(data =>
          data.props.children
            .toLowerCase()
            .includes(e.target.value.toLowerCase())
        ),
        cursor: 0,
      });

      if (onChange !== undefined) {
        return onChange(e);
      }
    };
```

This is the old onChange function. It is calling the `onChange` callback with `e`. That's why I want to revert this. In my opinion this `onChange` should not be called here at all. What's the point of calling it with what the user has typed? Ow with the change event? But for some reason it was being done like that in the past, and it did not interfere with `onChange` being called when an option is called. So to be honest, I don't understand why it was like that, but I want to revert it to that behaviour since the new one does not work, apparently.

westwood846:
> please explain the changes and check how the changes will affect other projects.

Updated the description. I tested against Employers, that's how I found the two other bugs.

@tingtingdisco what's the bug on Marketplace? I Can test against Marketplace too to confirm that this PR fixes the bug as expected.

src/Input/Select/Select.tsx
tingtingdisco: thank u @westwood846 got it

tingtingdisco:
@westwood846 I only know community team has a feature is related to this change, so I have written a test for it, but I am not sure whether we have other features rely on th `onChange`.

tingtingdisco (approved):
LGTM

PR #328, opened 2020-05-08
Swipeable: make scrollbar visible on desktop

rosslo:


src/Style/Display/SwipeableStyle.ts
```diff
@@ -1,15 +1,31 @@
...
   &::-webkit-scrollbar {
-    display: none;
+    -webkit-appearance: none;
+    height: 7px;
+  }
+
+  /* make the horizontal scrollbar visible on desktop */
+  &::-webkit-scrollbar-thumb {
+    border-radius: 4px;
+    background-color: rgba(0, 0, 0, 0.5);
```
franciscrispin: @rosslo btw the color is different from `dst` but that might not matter 
https://gitlab.glints.com/glints/glints-dst/-/blob/staging/next/pages/_app.sc.ts#L78

franciscrispin (approved):

src/Style/Display/SwipeableStyle.ts
rosslo: Nice catch, I wasn't aware of this. Just updated it, thanks!

PR #344, opened 2020-05-22
Select: Refactor and fix bugs

### LINKED ISSUE #224: Select: input value should be controlled by the value prop only
Related issue: https://glints.slack.com/archives/CS09D0LJX/p1580881897001900

### LINKED ISSUE #341: Select.Option: Allow component to accept props passed to it
- Use-case: I want to be able to pass some data-attributes to Select.Option
- Currently Select.Option is transformed using React.Children and no props are getting passed through to the `SelectItem` component

### LINKED ISSUE #342: Select: children and isLoading uses a stale value when rendering async data
- The children and isLoading prop do not get updated when their values change

`array` is fetched asynchronously 

```
<Select isLoading={isLoading}>
  {array.map((value) => {
    <Select.Option value={value} key={value}>
      {value}
    </Select.Option>;
  })}
</Select>;
```

Example: https://www.loom.com/share/cc6e672bc8d140f2af39e817fccd3443

In the video here, the search term "hello" returns 0 results as seen in the console (the empty array right at the bottom). However, the previous array with  30 values is still rendered.

### LINKED ISSUE #343: Select: Inconsistent behaviour of `onChange` function
- The onChange function returns the `event` when typing into the input but returns the `value` of the `Select.Option` when clicking an item from the dropdown. This inconsistent behaviour results in code that looks like this when I want to handle the onChange differently:

```
onChange={(
  valueOrEvent: string | React.ChangeEvent<HTMLInputElement>
) => {
  if (typeof valueOrEvent === 'string') {
    handleValue(valueOrEvent); 
  } else {
    handleEvent(valueOrEvent)
  }
}}
```

Ideally the onChange should always return the `event`. Returning the `value` directly while seemingly convenient may not be the best practice as it leads to inconsistent behaviour.

tingtingdisco:
Added prop `onInputChange` to resolve #343 
Fixes #342 
Fixes #224 
Resolves #341

### DIFF
--- a/src/Input/Select/Select.test.tsx
+++ b/src/Input/Select/Select.test.tsx
@@ -360,3 +360,196 @@ it('up and down arrow keys can be used to navigate the options menu', () => {
   fireEvent.keyDown(selectInput, arrowUpEvent);
   expect(firstOption).toHaveClass('active');
 });
+
+describe('<Select/> on click outside', () => {
+  it('should lose focus', async () => {
+    const Component = () => (
+      <React.Fragment>
+        <Select label="ClickOutside">
+          <Select.Option value="ClickOutside">ClickOutside</Select.Option>
+        </Select>
+        <div>outside</div>
+      </React.Fragment>
+    );
+
+    const { queryByTestId, queryByRole, queryByText } = render(<Component />);
+    const selectInput = queryByRole('combobox') as HTMLInputElement;
+    const selectList = queryByTestId('listbox');
+    const outside = queryByText('outside');
+
+    fireEvent.focus(selectInput);
+    expect(selectList.hasAttribute('open')).toEqual(true);
+    fireEvent.click(outside);
+    expect(selectList.hasAttribute('open')).toEqual(false);
+    expect(selectInput).not.toHaveFocus();
+  });
+
+  it('should remove click outside listener when unmount', () => {
+    const addEventListenerSpy = jest.spyOn(document, 'addEventListener');
+    const removeEventListenerSpy = jest.spyOn(document, 'removeEventListener');
+
+    const Component = () => (
+      <React.Fragment>
+        <Select label="ClickOutside">
+          <Select.Option value="ClickOutside">ClickOutside</Select.Option>
+        </Select>
+      </React.Fragment>
+    );
+
+    const { unmount } = render(<Component />);
+    unmount();
+    expect(removeEventListenerSpy.mock.calls[0][1]).toEqual(
+      addEventListenerSpy.mock.calls[0][1]
+    );
+    addEventListenerSpy.mockRestore();
+    removeEventListenerSpy.mockRestore();
+  });
+});
+
+describe('<Select/> onBlur onFocus', () => {
+  it('should call onBlur and onFocus', () => {
+    const onBlurSpy = jest.fn();
+    const onFocusSpy = jest.fn();
+    const Component = () => (
+      <React.Fragment>
+        <Select label="ClickOutside" onBlur={onBlurSpy} onFocus={onFocusSpy}>
+          <Select.Option value="ClickOutside">ClickOutside</Select.Option>
+        </Select>
+      </React.Fragment>
+    );
+
+    const { queryByRole, asFragment } = render(<Component />);
+    const selectInput = queryByRole('combobox') as HTMLInputElement;
+
+    fireEvent.focus(selectInput);
+    expect(onFocusSpy).toHaveBeenCalledTimes(1);
+    // use snapshot to validate isFocus
+    expect(asFragment()).toMatchSnapshot();
+
+    fireEvent.blur(selectInput);
+    expect(onBlurSpy).toHaveBeenCalledTimes(1);
+    // use snapshot to validate floating
+    expect(asFragment()).toMatchSnapshot();
+  });
+});
+
+describe('<Select/> handleValueChange', () => {
+  it('input value should be the same as passed prop value', async () => {
+    const Component = ({ value }: { value: string }) => (
+      <React.Fragment>
+        <Select label="handleValueChange" value={value}>
+          <Select.Option value="handleValueChange">
+            handleValueChange
+          </Select.Option>
+        </Select>
+      </React.Fragment>
+    );
+
+    const mockValue = 'mock';
+    const { queryByRole, rerender } = render(<Component value={mockValue} />);
+    const selectInput = queryByRole('combobox');
+    expect(selectInput.getAttribute('value')).toEqual(mockValue);
+    rerender(<Component value="" />);
+    expect(selectInput.getAttribute('value')).toEqual('');
+  });
+});
+
+describe('<Select/> prop onInputChange', () => {
+  it('should call onInputChange on input change', async () => {
+    const onInputChangeSpy = jest.fn();
+    const Component = () => (
+      <React.Fragment>
+        <Select label="onInputChange" onInputChange={onInputChangeSpy}>
+          <Select.Option value="onInputChange">onInputChange</Select.Option>
+        </Select>
+      </React.Fragment>
+    );
+
+    const mockValue = 'mock value';
+    const { queryByRole } = render(<Component />);
+    const selectInput = queryByRole('combobox');
+    fireEvent.change(selectInput, {
+      target: {
+        value: mockValue,
+      },
+    });
+    expect(onInputChangeSpy).toHaveBeenCalledTimes(1);
+    // not able to validate SyntheticEvent
+  });
+});
+
+describe('<Select/> disableTyping', () => {
+  it('esc key should have no effect', async () => {
+    const Component = () => (
+      <React.Fragment>
+        <Select label="disableTyping" disableTyping={true}>
+          <Select.Option value="disableTyping">disableTyping</Select.Option>
+        </Select>
+      </React.Fragment>
+    );
+
+    const { queryByRole, queryByTestId } = render(<Component />);
+    const selectInput = queryByRole('combobox');
+    const selectList = queryByTestId('listbox');
+
+    fireEvent.focus(selectInput);
+    expect(selectList.hasAttribute('open')).toEqual(true);
+    fireEvent.keyDown(selectInput, { key: 'Esc', keyCode: 27 });
+    expect(selectList.hasAttribute('open')).toEqual(true);
+  });
+});
+
+describe('<Select/> esc key', () => {
+  it('esk keydown event should close the option list', async () => {
+    const Component = () => (
+      <React.Fragment>
+        <Select label="disableTyping">
+          <Select.Option value="disableTyping">disableTyping</Select.Option>
+        </Select>
+      </React.Fragment>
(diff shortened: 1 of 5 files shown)


src/Input/Select/Select.test.tsx
```diff
@@ -360,3 +360,196 @@ it('up and down arrow keys can be used to navigate the options menu', () => {
...
+      </React.Fragment>
+    );
+
+    const { queryByTestId, queryByRole, queryByText } = render(<Component />);
+    const selectInput = queryByRole('combobox') as HTMLInputElement;
+    const selectList = queryByTestId('listbox');
+    const outside = queryByText('outside');
+
+    fireEvent.focus(selectInput);
+    expect(selectList.hasAttribute('open')).toEqual(true);
```
rosslo, line 382: should we add `expect(selectInput).toHaveFocus();` to correspond to line 384?

src/Input/Select/Select.test.tsx
```diff
@@ -360,3 +360,196 @@ it('up and down arrow keys can be used to navigate the options menu', () => {
...
+
+    const { queryByRole, queryByTestId, asFragment } = render(<Component />);
+    const selectInput = queryByRole('combobox');
+    const option = queryByTestId('option');
+    const selectList = queryByTestId('listbox');
+
+    fireEvent.focus(selectInput);
+    expect(selectList.hasAttribute('open')).toEqual(true);
+    fireEvent.mouseEnter(option);
+    expect(asFragment()).toMatchSnapshot();
```
rosslo, line 575: I think we should have `expect(option).toHaveClass('active');`

src/Input/Select/Select.test.tsx
```diff
@@ -360,3 +360,196 @@ it('up and down arrow keys can be used to navigate the options menu', () => {
...
+
+    fireEvent.focus(selectInput);
+    expect(selectList.hasAttribute('open')).toEqual(true);
+    fireEvent.keyDown(selectInput, { key: 'Esc', keyCode: 27 });
+    expect(selectList.hasAttribute('open')).toEqual(true);
+  });
+});
+
+describe('<Select/> esc key', () => {
+  it('esk keydown event should close the option list', async () => {
```
rosslo: typo `esk`

src/Input/Select/Select.test.tsx
```diff
@@ -360,3 +360,196 @@ it('up and down arrow keys can be used to navigate the options menu', () => {
...
+    const Component = () => (
+      <React.Fragment>
+        <Select label="disableTyping">
```
rosslo, line 506: I think the label value `disableTyping` is not suitable for this case.

src/Input/Select/Select.test.tsx
```diff
@@ -360,3 +360,196 @@ it('up and down arrow keys can be used to navigate the options menu', () => {
...
+      </React.Fragment>
+    );
+
+    const { queryByRole, asFragment } = render(<Component />);
+    const selectInput = queryByRole('combobox') as HTMLInputElement;
+
+    fireEvent.focus(selectInput);
+    expect(onFocusSpy).toHaveBeenCalledTimes(1);
+    // use snapshot to validate isFocus
+    expect(asFragment()).toMatchSnapshot();
```
rosslo, line 427: May I know the reason why you use `toMatchSnapshot` rather than `toHaveFocus`?

src/Input/Select/Select.test.tsx
```diff
@@ -360,3 +360,196 @@ it('up and down arrow keys can be used to navigate the options menu', () => {
...
+        </Select>
+      </React.Fragment>
+    );
+
+    const mockValue = 'mock';
+    const { queryByRole, rerender } = render(<Component value={mockValue} />);
+    const selectInput = queryByRole('combobox');
+    expect(selectInput.getAttribute('value')).toEqual(mockValue);
+    rerender(<Component value="" />);
+    expect(selectInput.getAttribute('value')).toEqual('');
```
rosslo, line 455: Could we test the `value` change start from an empty string?  So the change flow might be "" -> "mock" -> "".
Because I have a workaround for the case that the `value` prop doesn't apply because the initial value is an empty string.
https://gitlab.glints.com/glints/glints-dst/-/merge_requests/1485/diffs#78d78e4d84be8f56d70d434e1b9e2859becb3501_0_77

src/Input/Select/Select.test.tsx
```diff
@@ -360,3 +360,196 @@ it('up and down arrow keys can be used to navigate the options menu', () => {
...
+    const mockValue = 'mock value';
+    const { queryByRole } = render(<Component />);
+    const selectInput = queryByRole('combobox');
+    fireEvent.change(selectInput, {
+      target: {
+        value: mockValue,
+      },
+    });
+    expect(onInputChangeSpy).toHaveBeenCalledTimes(1);
+    // not able to validate SyntheticEvent
```
rosslo, line 479: What does this comment mean?

src/Input/Select/Select.test.tsx
```diff
@@ -360,3 +360,196 @@ it('up and down arrow keys can be used to navigate the options menu', () => {
...
+  });
+});
+
+describe('<Select/> disableTyping', () => {
+  it('esc key should have no effect', async () => {
```
rosslo: Shouldn't the users be able to select an option via keyboard?

src/Input/Select/Select.tsx
```diff
@@ -24,82 +24,118 @@ const Select: ISelect = (props: Props) => {
...
+    if (!inputValue) {
+      return childrenOptions;
     }
-  }, []);
 
-  React.useLayoutEffect(() => {
-    // Checking if children data is exist or not.
-    if (React.Children.count(children) !== 0) {
-      setFilterValue(React.Children.map(children, data => data));
+    const isInputValueOneOfOptions = childrenOptions.find(
```
rosslo: I think `.some` is more suitable than `.find` because the `isInputValueOneOfOptions` should be a boolean.

src/Input/Select/Select.tsx
```diff
@@ -24,82 +24,118 @@ const Select: ISelect = (props: Props) => {
...
+    function updateOptions() {
+      const availableOptions = getAvailableOptions();
+      setOptions(availableOptions);
+    },
+    [getAvailableOptions]
+  );
 
-    return () => document.removeEventListener('click', handleOnBlur, false);
-  }, [children, handleOnBlur, status, value]);
+  // handle click outside
```
rosslo: Can remove this comment because the function name is self-explanatory.

rosslo:
Could you comment on the https://github.com/glints-dev/glints-aries/issues/343 for the solution 'onInputChange'?

src/Input/Select/Select.tsx
```diff
@@ -24,82 +24,118 @@ const Select: ISelect = (props: Props) => {
...
+  const [activeOptionIndex, setActiveOptionIndex] = React.useState<number>(0);
+  const [
+    shouldScrollToActiveOption,
+    setShouldScrollToActiveOption,
+  ] = React.useState<boolean>(false);
+
+  const selectContainerRef: React.RefObject<HTMLDivElement> = React.useRef();
+
+  // set options based on children and inputValue
+  const getAvailableOptions = React.useCallback(() => {
```
rosslo: Do you think it is possible to use `const availabeOptions = useMemo(// ...)` to instead of `getAvailableOptions` and `const [options, setOptions] = React.useState(//...)`?

src/Input/Select/Select.tsx
```diff
@@ -24,82 +24,118 @@ const Select: ISelect = (props: Props) => {
...
     small,
     disableTyping,
     removeFloatingLabel,
     removeDropIcon,
     error,
     renderError,
-    value, // eslint-disable-line @typescript-eslint/no-unused-vars
-    children, // eslint-disable-line @typescript-eslint/no-unused-vars
+    value,
+    defaultValue,
```
rosslo, line 36: `defaultValue` should be passed to `SelectInput`

stories/Input/SelectStory.tsx
```diff
@@ -38,15 +38,15 @@ const props = {
...
+      description: `WARNING: Don't customize error message style when passing ReactNode, pass true to display error state without error message`,
     },
     {
       name: 'renderError',
       type: 'ReactNode',
       defaultValue: '<span color="#EC272B" />',
       possibleValue: 'any',
       require: 'no',
-      description: `WARNING: Will only allow string type in the future, don't customize error message style. Replaces the default error component. Receives 'error' as children.`,
+      description: `WARNING: Don't customize error message style. Replaces the default error component. Receives 'error' as children.`,
```
rosslo: I think this can be simplified to "(DEPRECATED) please use `error` instead"

src/Input/Select/Select.tsx
```diff
@@ -24,82 +24,118 @@ const Select: ISelect = (props: Props) => {
...
 
-  React.useLayoutEffect(() => {
-    // Checking if children data is exist or not.
-    if (React.Children.count(children) !== 0) {
-      setFilterValue(React.Children.map(children, data => data));
+    const isInputValueOneOfOptions = childrenOptions.find(
+      data => data.props.children === inputValue
+    );
+    if (isInputValueOneOfOptions) {
+      return childrenOptions;
```
rosslo, line 80: If the user searches the correct full name of one of the options, the option won't be just one.  
Maybe it can be resolved by adding a state `isTyping` or `isSearching` with condition `isInputValueOneOfOptions` 🤔

image

src/Input/Select/Select.test.tsx
tingtingdisco: will remove `toHaveFocus`, fireEvent.focus doesn't update jsdom.

src/Input/Select/Select.test.tsx
tingtingdisco: please check snapshot the element has class `active`

src/Input/Select/Select.test.tsx
tingtingdisco: because we can validate it from the snapshot.

src/Input/Select/Select.test.tsx
tingtingdisco: we cannot validate the argument of `onInputChangeSpy` when it's being called, because the event object is a SyntheticEvent.

src/Input/Select/Select.test.tsx
tingtingdisco: will update

tingtingdisco:
> Could you comment on the #343 for the solution 'onInputChange'?

the issue will be linked to this PR.

src/Input/Select/Select.tsx
tingtingdisco: it's being used to set `inputValue` already.

src/Input/Select/Select.tsx
tingtingdisco: I will not do this improvement in this PR

rosslo (approved):

PR #345, opened 2020-05-25
Modal: won't mount children when visible

### LINKED ISSUE #214: Modal children should only render if modal is open

rosslo:
## Why
- To avoid unmounting the children manually, e.g. https://gitlab.glints.com/glints/glints-dst/-/blob/staging/app/modules/Login/Components/LoginModal.js#L18

## What changed
- Modal won't mount children when it is not visible
- Gallery add test case for the failed focus event
- Already tested the `QuestionCardGallery` component on my local

### DIFF
--- a/src/Application/Gallery/Gallery.tsx
+++ b/src/Application/Gallery/Gallery.tsx
@@ -64,7 +64,12 @@ class Gallery extends React.Component<Props, State> {
   }
 
   componentDidUpdate(prevProps: Props, prevState: State) {
-    if (!prevState.visible && this.state.visible) {
+    if (
+      !prevState.visible &&
+      this.state.visible &&
+      this.sliderRef.current &&
+      this.sliderRef.current.sliderContainerRef.current
+    ) {
       this.sliderRef.current.sliderContainerRef.current.focus();
     }
   }
--- a/src/Application/Gallery/Gallery.tsx
+++ b/src/Application/Gallery/Gallery.tsx
@@ -69,10 +69,6 @@ class Gallery extends React.Component<Props, State> {
     }
   }
 
-  componentWillUnmount() {
-    this.sliderRef.current.sliderContainerRef.current.blur();
-  }
-
   render() {
     const { children, imagesDisplayed = defaultImagesDisplayed } = this.props;
     const { visible, currentIndex, imageLeft } = this.state;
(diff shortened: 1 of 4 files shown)


src/Application/Gallery/Gallery.tsx
```diff
@@ -63,16 +63,12 @@ class Gallery extends React.Component<Props, State> {
     }
   }
 
-  componentDidUpdate(prevProps: Props, nextState: State) {
-    if (!nextState.visible) {
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    if (!prevState.visible && this.state.visible) {
```
ritikamotwani: Is this to check if the the state has moved from not visible to visible  for the focused container on update?

src/Display/Modal/Modal.test.tsx
```diff
@@ -96,6 +96,16 @@ describe('when modal is closed', () => {
   });
 });
 
+it('children should not be mounted', () => {
```
ritikamotwani: should we change this to `children should not be mounted when modal isn't visible`? :D

ritikamotwani:
LGTM!
Checked the gallery component too, working fine haha :D 
Thanks for testing the QuestionCard locally!

src/Application/Gallery/Gallery.tsx
rosslo: yes, once the Slider is mounted, it should be focused to make sure the arrow keys can control it.

src/Display/Modal/Modal.test.tsx
rosslo: Oh I intended to put inside `describe('when modal is closed', () => {})` haha. 😅 
Just changed it, thanks!

src/Application/Gallery/Gallery.tsx
rosslo: Btw, I added more conditions to ensure the focus event won't cause a reference error.

ritikamotwani:
@rosslo , I found a bug in the gallery component.
When you open an image. The arrow keys work.
Then 
1. Select an image from the smaller thumbnails in the below row.
2. The arrow keys no more work
Screenshot 2020-05-26 at 12 20 43 PM

It's from before unrelated to this PR. Will open an issue for this

ritikamotwani (approved):

PR #347, opened 2020-05-26
Convert Dropdown to Functional Component

### LINKED ISSUE #346: Dropdown: Convert to functional component

westwood846:
Closes #346

### DIFF
--- a/src/Navigation/Dropdown/Dropdown.test.tsx
+++ b/src/Navigation/Dropdown/Dropdown.test.tsx
@@ -311,18 +311,18 @@ describe('<Dropdown/> keydown event', () => {
 
   test('handleEscKeydown listener should be removed before component unmount', async () => {
     const removeEventListenerSpy = jest.spyOn(document, 'removeEventListener');
-    const ref = React.createRef<Dropdown>();
     const { unmount } = render(
-      <Dropdown label="Career" ref={ref}>
+      <Dropdown label="Career">
         <DropdownItem value="Product Manager">Product Manager</DropdownItem>
         <DropdownItem value="Software Engineer">Software Engineer</DropdownItem>
       </Dropdown>
     );
 
-    const handleEscKeydown = ref.current.handleEscKeydown;
     unmount();
     expect(removeEventListenerSpy).toHaveBeenCalledTimes(1);
-    expect(removeEventListenerSpy.mock.calls[0][1]).toEqual(handleEscKeydown);
+    expect(
+      (removeEventListenerSpy.mock.calls[0][1] as () => void).name
+    ).toEqual('listener');
     removeEventListenerSpy.mockRestore();
   });
 });
--- a/src/Navigation/Dropdown/Dropdown.tsx
+++ b/src/Navigation/Dropdown/Dropdown.tsx
@@ -17,89 +17,69 @@ import {
 
 import DropdownItem, { Props as DropdownItemProps } from './DropdownItem';
 
-class Dropdown extends React.Component<Props, State> {
-  static Item = DropdownItem;
+const Dropdown = (props: Props) => {
+  const {
+    label,
+    showHoverLine = false,
+    dropDownPlacement = 'left',
+    noLineBreak = false,
+    iconDefaultColor = 'black',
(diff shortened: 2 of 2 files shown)


src/Navigation/Dropdown/Dropdown.test.tsx
```diff
@@ -311,18 +311,15 @@ describe('<Dropdown/> keydown event', () => {
...
+      <Dropdown label="Career">
         <DropdownItem value="Product Manager">Product Manager</DropdownItem>
         <DropdownItem value="Software Engineer">Software Engineer</DropdownItem>
       </Dropdown>
     );
 
-    const handleEscKeydown = ref.current.handleEscKeydown;
     unmount();
     expect(removeEventListenerSpy).toHaveBeenCalledTimes(1);
-    expect(removeEventListenerSpy.mock.calls[0][1]).toEqual(handleEscKeydown);
```
westwood846, line 325: I wasn't sure how to export the esc event from the functional component to keep this expectation working. Do you think testing that the `removeEventListenerSpy` is sufficient?

src/Navigation/Dropdown/Dropdown.tsx
```diff
@@ -17,231 +17,180 @@ import {
...
-    const { onChange } = this.props;
     const itemElement = e.target as HTMLLIElement;
 
     if (itemElement.dataset.value) {
-      this.setState({
-        dropdownLabel: itemElement.dataset.value,
-        isOpen: false,
-      });
+      // TODO: This should probably set to itemElement.innerHTML
+      setDropdownLabel(itemElement.dataset.value);
```
westwood846, line 75: As you can see, when an item is clicked, the label gets set to the item's value. Of course that makes little sense and indeed further down, where enter key presses are handled, the handler there sets the label to the item's `innerHTML`. So I'm pretty sure this is unintended.

But since this is just a conversion PR I won't change the behaviour and keep setting the label to the item's value here for now.

src/Navigation/Dropdown/Dropdown.test.tsx
tingtingdisco: I think we can do this at least, `listener` is the name of the function returned by `escEvent`:
```
expect((removeEventListenerSpy.mock.calls[0][1] as () => void).name).toEqual(
  'listener'
);
```

src/Navigation/Dropdown/Dropdown.tsx
```diff
@@ -17,231 +17,180 @@ import {
...
-                );
-              }
-            )}
-          </DropdownBody>
-        </DropdownWrapper>
-      </DropdownContainer>
-    );
-  }
-}
+  React.useLayoutEffect(() => {
```
tingtingdisco: why need to use `useLayoutEffect`?

src/Navigation/Dropdown/Dropdown.tsx
tingtingdisco: do you mind open an issue for this? I think it's ok to not change it in this PR, but this is so weird haha

westwood846: Opened it here: https://github.com/glints-dev/glints-aries/issues/348

src/Navigation/Dropdown/Dropdown.tsx
westwood846: When I'm replacing `onComponentDidMount` I'm defaulting to `useLayoutEffect` because I had some trouble with tests in the past: https://glints.slack.com/archives/C956ZDAP7/p1584325219041900

src/Navigation/Dropdown/Dropdown.test.tsx
westwood846: Done!

src/Navigation/Dropdown/Dropdown.tsx
tingtingdisco: hmm, because this effect is not related to update dom elements, so useEffect is ok.

tingtingdisco (approved):
LGTM, can change useLayoutEffect to useEffect.

PR #358, opened 2020-06-02
Slider: Add tests

### LINKED ISSUE #290: Tests: Slider

tingtingdisco:
closes #290

### DIFF
--- a/src/Display/Slider/Slider.test.tsx
+++ b/src/Display/Slider/Slider.test.tsx
@@ -1,0 +1,311 @@
+import 'jest-styled-components';
+import '@testing-library/jest-dom/extend-expect';
+import { render, wait, fireEvent } from '@testing-library/react';
+import * as React from 'react';
+
+import Slider, { Props } from './Slider';
+
+type SliderComponentProp = Omit<Props, 'children'>;
+const SliderComponent = (props: SliderComponentProp) => (
+  <Slider {...props}>
+    <Slider.Item>a</Slider.Item>
+    <Slider.Item>b</Slider.Item>
+    <Slider.Item>c</Slider.Item>
+  </Slider>
+);
+
+describe('<Slider/> prop className', () => {
+  const matchSnapshot = (className: string) => {
+    it(`should match snapshot when class name ${className} is passed`, () => {
+      const { asFragment } = render(<SliderComponent className={className} />);
+      expect(asFragment()).toMatchSnapshot();
+    });
+  };
+
+  ['test', undefined].forEach(className => matchSnapshot(className));
+});
+
+describe('<Slider/> prop initialItem', () => {
+  const matchSnapshot = (initialItem: number) => {
+    it(`should match snapshot when initialItem ${initialItem} is passed`, async () => {
+      const sliderRef: any = {
+        current: null,
+      };
+      const { rerender, asFragment } = render(
+        <Slider ref={sliderRef} initialItem={initialItem}>
+          <Slider.Item>a</Slider.Item>
+          <Slider.Item>b</Slider.Item>
+          <Slider.Item>c</Slider.Item>
+        </Slider>
+      );
+
+      const index = initialItem > 0 ? initialItem - 1 : 0;
+      await wait(() => {
+        expect(
+          document.querySelectorAll('li')[index].classList.contains('active')
+        ).toBeTruthy();
+      });
+
+      // mock getSliderContainerDOMNode to calculate translateX
+      sliderRef.current.getSliderContainerDOMNode = () => ({
+        getBoundingClientRect: () => ({
+          width: 500,
+        }),
+      });
+
+      rerender(
+        <Slider ref={sliderRef} initialItem={initialItem}>
+          <Slider.Item>a</Slider.Item>
+          <Slider.Item>b</Slider.Item>
+          <Slider.Item>c</Slider.Item>
+        </Slider>
+      );
+      expect(asFragment()).toMatchSnapshot();
+    });
+  };
+
+  [0, 1, 2, 3].forEach(initialItem => matchSnapshot(initialItem));
+});
+
+describe('<Slider/> prop fullContent', () => {
+  const matchSnapshot = (fullContent: boolean) => {
+    it(`should match snapshot when fullContent ${fullContent} is passed`, async () => {
+      const { asFragment } = render(
+        <SliderComponent fullContent={fullContent} />
+      );
+      expect(asFragment()).toMatchSnapshot();
+    });
+  };
+
+  [true, false].forEach(fullContent => matchSnapshot(fullContent));
+});
+
+describe('<Slider/> prop arrowWhite', () => {
+  const matchSnapshot = (arrowWhite: boolean) => {
+    it(`should match snapshot when arrowWhite ${arrowWhite} is passed`, async () => {
+      const { asFragment } = render(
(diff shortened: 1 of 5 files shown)


src/Display/Slider/Slider.test.tsx
```diff
@@ -0,0 +1,311 @@
...
+      expect(asFragment()).toMatchSnapshot();
+    });
+  };
+
+  ['test', undefined].forEach(className => matchSnapshot(className));
+});
+
+describe('<Slider/> prop initialItem', () => {
+  const matchSnapshot = (initialItem: number) => {
+    it(`should match snapshot when initialItem ${initialItem} is passed`, async () => {
```
westwood846, line 31: Could you describe in more detail what's going on in this test? Why do you use `wait`? Why exactly do you need to mock `getSliderContainerDOMNode`? And why `rerender`? Some more or more elaborate comments would be helpful.

src/Display/Slider/Slider.test.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+    // validate initial active slide
+    expect(
+      document.querySelectorAll('li')[0].classList.contains('active')
+    ).toBeTruthy();
+    expect(asFragment()).toMatchSnapshot();
+
+    // validate autoplay
+    jest.advanceTimersByTime(6000);
+    expect(
```
westwood846: Should we test here that the `li` with index 0 is no longer `active`?

src/Display/Slider/Slider.test.tsx
```diff
@@ -0,0 +1,311 @@
...
+      document.querySelectorAll('li')[0].classList.contains('active')
```
westwood846: Also could test here that next `li` is _not_ `active`.

src/Display/Slider/Slider.test.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+  it(`should go to the first slide after 6000ms when autoplay is truthy and the current slide is the last slide`, () => {
+    jest.useFakeTimers();
+    const { asFragment } = render(
+      <SliderComponent autoplay={true} initialItem={3} />
+    );
+
+    // validate initial active slide
+    expect(
+      document.querySelectorAll('li')[2].classList.contains('active')
```
westwood846: Would it be possible to have a `const listItems = document.querySelectorAll('li')` at the top of the test case and then do `listItems[0]`? And `listItems[listItems.length - 1]` when we're targeting the last item, to make it clearer that it's the last?

src/Display/Slider/Slider.test.tsx
```diff
@@ -0,0 +1,311 @@
...
+describe('<Slider/> prop afterChange', () => {
+  test(`when active slide is the first slide`, () => {
+    const afterChange = jest.fn();
+    const { queryByTestId } = render(
+      <SliderComponent afterChange={afterChange} />
+    );
+
+    const leftArrow = queryByTestId('slider_left-arrow');
+    const rightArrow = queryByTestId('slider_right-arrow');
+    const firstDot = document.querySelectorAll('li')[0];
```
westwood846, line 174: How about the other dots?

westwood846 (changes requested):
Interesting technique to put the snapshot tests into loops like that. Now I'm wondering if we could use something similar to test all possible combinations of attributes, not just every single attribute in isolation from the others...

src/Display/Slider/Slider.test.tsx
tingtingdisco: because the slider content has a style `transformX` which is calculated based on `getSliderContainerDOMNode. getBoundingClientRect.width`, and `getBoundingClientRect` is not implemented in JSDOM, so in order to validate `transformX` is being calculated correctly, we need to mock `getSliderContainerDOMNode` to get the element width.

Since the component recalculate `transformX` in `componentWillReceiveProps`, I use rerender to trigger it.

using `wait` is because the component do a setState in componentDidMount to set the active dot, setState is asynchronous so I need to `wait` after `render` to validate active dot position.

src/Display/Slider/Slider.test.tsx
tingtingdisco: ok, I will validate there is only one active li.

tingtingdisco: oh, I validate it already using snapshot, in the snapshot only the second li has class name active.

src/Display/Slider/Slider.test.tsx
tingtingdisco: ok I will name the li array.

src/Display/Slider/Slider.test.tsx
tingtingdisco: validate through snapshot

src/Display/Slider/Slider.test.tsx
tingtingdisco: my idea is that this case is testing `afterChange`, so I only test the things can trigger it based on the implementation of the component. For an example based testing, I think this is enough.

src/Display/Slider/Slider.test.tsx
tingtingdisco: this is related to useLayoutEffect and useEffect, I am not super sure about this, but I think I read somewhere that if setState is executed in useEffect or componentDidMount, it's not guaranteed that setState will be executed synchronously, only useLayoutEffect guarantees synchronous setState.

src/Display/Slider/Slider.test.tsx
westwood846: True

tingtingdisco:
> Interesting technique to put the snapshot tests into loops like that. Now I'm wondering if we could use something similar to test all possible combinations of attributes, not just every single attribute in isolation from the others...

That will be an extra generative test, and I think isolate test for each property is still needed.

westwood846:
> > Interesting technique to put the snapshot tests into loops like that. Now I'm wondering if we could use something similar to test all possible combinations of attributes, not just every single attribute in isolation from the others...
> 
> That will be an extra generative test, and I think isolate test for each property is still needed.

Yep that would be in addition to testing each property individually. Although the combinations could also include combinations where attributes are not set, so it would be a superset of what you're doing here.

tingtingdisco:
> > 
> > That will be an extra generative test, and I think isolate test for each property is still needed.
> 
> Yep that would be in addition to testing each property individually. Although the combinations could also include combinations where attributes are not set, so it would be a superset of what you're doing here.

I think we do the current example based test is enough, I don't want to read all the snapshots generated by the combination test.

tingtingdisco:
@westwood846 MR updated, named the dot elements and added comments to explain the wait and rerender.

src/Display/Slider/Slider.test.tsx
westwood846: >this is related to useLayoutEffect and useEffect, I am not super sure about this, but I think I read somewhere that if setState is executed in useEffect or componentDidMount, it's not guaranteed that setState will be executed synchronously, only useLayoutEffect guarantees synchronous setState.
Also not sure but that makes sense based on my observations.

westwood846: Can you please add your explanations from above as comments in the code? This is a quite advanced test case so it would be cool if we could point to it in future and say "just look how it's done there" when we have something similar to test.

westwood846 (changes requested):

westwood846 (approved):
LGTM

PR #361, opened 2020-06-04
Fix dropdown onclick

### LINKED ISSUE #348: Clicking a Dropdown Item Sets its Label to the Item's Value
See https://github.com/glints-dev/glints-aries/pull/347

The label should be set to the Item's `innerHTML`, not the `value`.

westwood846:
Fixes #348 

This change is safe to make because the `handleKeyDown` function already sets the label to the option's label.

### DIFF
--- a/src/Navigation/Dropdown/Dropdown.tsx
+++ b/src/Navigation/Dropdown/Dropdown.tsx
@@ -71,8 +71,7 @@ const Dropdown = (props: Props) => {
     const itemElement = e.target as HTMLLIElement;
 
     if (itemElement.dataset.value) {
-      // TODO: This should probably set to itemElement.innerHTML
-      setDropdownLabel(itemElement.dataset.value);
+      setDropdownLabel(itemElement.innerHTML);
       setIsOpen(false);
 
       if (onChange !== undefined) {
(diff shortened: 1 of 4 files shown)


src/Navigation/Dropdown/Dropdown.tsx
```diff
@@ -71,8 +71,7 @@ const Dropdown = (props: Props) => {
     const itemElement = e.target as HTMLLIElement;
 
     if (itemElement.dataset.value) {
-      // TODO: This should probably set to itemElement.innerHTML
-      setDropdownLabel(itemElement.dataset.value);
+      setDropdownLabel(itemElement.innerHTML);
```
tingtingdisco, line 74: shall we move `setDropdownLabel(itemElement.innerHTML)` out of this `if` block? I think the dropdown label should always be updated whether `itemElement.dataset.value` has value or not. What do you think?

tingtingdisco (approved):
LGTM

src/Navigation/Dropdown/Dropdown.tsx
westwood846: Oh actually I think this would break some usages of this component because people have been using it as a dropdown menu where clicking an option will trigger an action without changing the label. Those don't set the value to achieve the effect.

PR #366, opened 2020-06-08
Adds a new variant to the underlined tabs, with tests

### LINKED ISSUE #208: Tabs: Fix UI Mismatches
GDS: https://app.abstract.com/share/e51821dd-be59-4c6f-99ad-6aeba45fa6e7?collectionLayerId=697d2f21-b538-4525-82bc-e9278ca156e0

ritikamotwani:
Fixes: #208 

https://docs.google.com/document/d/1pHImZ2YXcG0WxMU5ll1y9zBygyErR2EjR8xu3Y2lgDw/edit

I have added a new colour variant (black and grey) only for the `underlined` tabs, because the other tabs are being used in the Q&A, and maybe some other parts. So for now it's ok to keep them. https://glints.slack.com/archives/CS09D0LJX/p1590990467027600
Will remove it after they are refactorred.

Also made the tab to take input as string and node.

### DIFF
--- a/src/Display/Tabs/TabsStyle.ts
+++ b/src/Display/Tabs/TabsStyle.ts
@@ -1,7 +1,16 @@
 import styled from 'styled-components';
 
-import { SecondaryColor } from '../../Utils/Colors';
-import { ScreenSize } from '../../Utils/StyleConfig';
+import {
+  ScreenSize,
+  ETabColourVariant,
+  ETabAlignment,
+} from '../../Utils/StyleConfig';
+import { SecondaryColor, Greyscale } from '../../Utils/Colors';
+
+interface TabHeader {
+  colour: ETabColourVariant;
+  alignment: string;
+}
 
 export const TabsContainer = styled.div`
   position: relative;
@@ -14,7 +23,7 @@ export const TabsContainer = styled.div`
   }
 `;
 
-export const TabsHeader = styled.div`
+export const TabsHeader = styled.div<TabHeader>`
   position: relative;
 
   &.vertical-tabs-header {
@@ -63,7 +72,8 @@ export const TabsHeader = styled.div`
     }
 
     .horizontal-tab.underlined {
-      margin: 0 10px;
+      margin: ${({ colour }) =>
+        colour === ETabColourVariant.GREY ? '0 15px' : '0 10px'};
     }
 
     .vertical-tab {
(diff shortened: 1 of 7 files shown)


stories/Display/TabsStory.tsx
```diff
@@ -31,6 +33,14 @@ const props = {
       description:
         'Sets alignment of Tab. The vertical tabs are changed to horizontal ones for screen size below 768',
     },
+    {
+      name: 'colour',
+      type: 'string',
+      defaultValue: 'black',
+      possibleValue: 'blue, grey',
```
rosslo: There is no `blue` in `ETabColourVariant`

stories/Display/TabsStory.tsx
```diff
@@ -31,6 +33,14 @@ const props = {
       description:
         'Sets alignment of Tab. The vertical tabs are changed to horizontal ones for screen size below 768',
     },
+    {
+      name: 'colour',
```
rosslo: I think we can rename it to `theme` to have the consistent name of the color prop because `Button` and `Divider` both are using `theme` prop.

stories/Display/TabsStory.tsx
rosslo: But I think we can use `blue` to instead of `grey` because using active color should be more intuitive. Wdyt?

stories/Display/TabsStory.tsx
```diff
@@ -134,6 +144,196 @@ const TabsStory = () => (
...
+    <Divider theme="grey" />
+
+    <StorybookComponent
+      title="Tabs"
+      code="import { Tabs } from 'glints-aries'"
+      usage={`<Tabs variant="underlined" colour="grey">
+        <Tabs.Pane tab={<div><FileAlternateIcon />Jobs</div>}>
+        Software Engineer <Badge label="1" />
+        </Tabs.Pane>
+        <Tabs.Pane tab={<div><FileAlternateIcon />Company</div>}>Glints</Tabs.Pane>
```
rosslo, line 187: I think we can have an `icon` prop instead of makeing `tab` props accept `React.ReactNode` like material-ui https://material-ui.com/zh/components/tabs/#icon-tabs

src/Display/Tabs/TabsStyle.ts
```diff
@@ -63,7 +72,8 @@ export const TabsHeader = styled.div`
     }
 
     .horizontal-tab.underlined {
-      margin: 0 10px;
+      margin: ${({ colour }) =>
+        colour === ETabColourVariant.GREY ? '0 15px' : '0 10px'};
```
rosslo: There are a lot of `colour === ETabColourVariant.GREY` conditions, I was thinking if we can have an intuitive variable instead. e.g. `const isNewTab = colour !== ETabColourVariant.BLACK`. Because it is a bit hard to infer why we change the style based on color.

rosslo:
For vertical tabs, I think we should set `padding-left` for each tabs.
image
image

stories/Display/TabsStory.tsx
ritikamotwani: https://github.com/glints-dev/glints-aries/issues/208#issuecomment-647971997 it was requested here to make it accept elements too. Also this was a limitation, that formatted message or <Style>tab heading</Styled> couldn't be added here, resulting in css change through classNames.
Therefore I didn't see a point of adding an additional prop?

rosslo (approved):

PR #370, opened 2020-06-09
Select: defaultOpen & close behavior

### LINKED ISSUE #357: Select: add a prop to control the initial open state of the Select
Currently, we can use `autoFocus` to trigger the `onFocus` to open the Select after mounting.
But I think it might not be safe for the long term. For example, if the `onFoucs` is refactored to `onClick` somehow, this behavior won't work.

### LINKED ISSUE #369: Select: improve the close behavior
The user should be able to close the select box by:

- Clicking on the select input again (for non-typing select)
- Clicking on the arrow (for typing select)
- Clicking on one of the option on the dropdown
- Clicking anything outside the select box

image

rosslo:
## What Changed

1. commit https://github.com/glints-dev/glints-aries/commit/146bbd2089bf6d1531b2f4189eb3c110efe6996f resolves issue https://github.com/glints-dev/glints-aries/issues/369

   The user should be able to close the select box by:
   - Clicking on the select input again (for non-typing select)
   - Clicking on the arrow (for typing select)
   - Clicking on one of the option on the dropdown
   - Clicking anything outside the select box

2. commit https://github.com/glints-dev/glints-aries/commit/3e698a63a3a473c55c706f34987785e4efce25d1 resolves issue https://github.com/glints-dev/glints-aries/issues/357

   Add `defaultOpen` prop to allow we set the initial open state.
   The name refers to https://ant.design/components/select/

### DIFF
--- a/src/Input/Select/Select.test.tsx
+++ b/src/Input/Select/Select.test.tsx
@@ -639,6 +639,75 @@ describe('<Select/> disableTyping', () => {
   });
 });
 
+describe('<Select/> defaultOpen', () => {
+  it('should be open', async () => {
+    const Component = () => (
+      <React.Fragment>
+        <Select label="defaultOpen" defaultOpen={true}>
+          <Select.Option value="defaultOpen">defaultOpen</Select.Option>
+        </Select>
+      </React.Fragment>
+    );
+
+    const { queryByTestId } = render(<Component />);
+    const selectList = queryByTestId('listbox');
+
+    expect(selectList.hasAttribute('open')).toEqual(true);
+  });
+
+  it('should be focused', async () => {
+    const Component = () => (
+      <React.Fragment>
+        <Select label="defaultOpen" defaultOpen={true}>
+          <Select.Option value="defaultOpen">defaultOpen</Select.Option>
+        </Select>
+      </React.Fragment>
+    );
+
+    const { queryByRole } = render(<Component />);
+    const selectInput = queryByRole('combobox');
+
+    expect(selectInput).toHaveFocus();
+  });
+
+  it('should be closed by clicking outside', async () => {
+    const Component = () => (
+      <React.Fragment>
+        <Select label="defaultOpen" defaultOpen={true}>
+          <Select.Option value="defaultOpen">defaultOpen</Select.Option>
+        </Select>
+        <div>outside</div>
+      </React.Fragment>
+    );
+
+    const { queryByTestId, queryByText } = render(<Component />);
+    const selectList = queryByTestId('listbox');
+    const outside = queryByText('outside');
+
+    expect(selectList.hasAttribute('open')).toEqual(true);
+    fireEvent.click(outside);
+    expect(selectList.hasAttribute('open')).toEqual(false);
+  });
+
+  it('should be closed by clicking option', async () => {
+    const Component = () => (
+      <React.Fragment>
+        <Select label="defaultOpen" defaultOpen={true}>
+          <Select.Option value="defaultOpen">defaultOpen</Select.Option>
+        </Select>
+      </React.Fragment>
+    );
+
+    const { queryByTestId } = render(<Component />);
+    const selectList = queryByTestId('listbox');
+    const option = queryByTestId('option');
+
+    expect(selectList.hasAttribute('open')).toEqual(true);
+    userEvent.click(option);
+    expect(selectList.hasAttribute('open')).toEqual(false);
+  });
+});
+
 describe('<Select/> esc key', () => {
   it('esc keydown event should close the option list', async () => {
     const Component = () => (
--- a/src/Input/Select/Select.test.tsx
+++ b/src/Input/Select/Select.test.tsx
@@ -512,7 +512,7 @@ describe('<Select />', () => {
     );
 
     const { queryByTestId } = render(<Component />);
-    const dropIcon = queryByTestId('select-drop-icon');
+    const dropIcon = document.querySelector('.select-icon');
     const selectList = queryByTestId('listbox');
 
     userEvent.click(dropIcon);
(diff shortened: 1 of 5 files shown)


tingtingdisco:
Any projects need defaultOpen?
if onFocus is refactored to onClick, we should still keep the onFocus logic.

src/Input/Select/Select.test.tsx
```diff
@@ -480,6 +480,78 @@ describe('<Select/> prop onInputChange', () => {
   });
 });
 
+describe('<Select />', () => {
```
tingtingdisco: maybe you can put all the new tests in this describe block and describe it as `<Select /> click to close option dropdown`?

src/Input/Select/Select.tsx
```diff
@@ -332,7 +347,12 @@ const Select: ISelect = (props: Props) => {
           </SelectLabel>
         )}
         {!removeDropIcon && (
-          <div className="select-icon" aria-label="show options">
+          <div
+            className="select-icon"
+            aria-label="show options"
+            data-testid="select-drop-icon"
```
tingtingdisco: data-testid is not needed, can do  `document.querySelector('.select-icon')`

src/Input/Select/Select.tsx
```diff
@@ -157,6 +158,18 @@ const Select: ISelect = (props: Props) => {
...
+  const handleClick = () => {
+    setIsFocus(!isFocus);
+  };
+
+  const handleDropIconClick = () => {
+    setIsFocus(!isFocus);
+
+    if (!disableTyping && !isFocus) {
+      selectInputRef.current.focus();
+    }
```
tingtingdisco, line 169: I am not sure whether `setIsFocus` can be run synchronously, so this is safer:
```
if (!disableTyping && !isFocus) {
      selectInputRef.current.focus();
}
setIsFocus(!isFocus);
```

tingtingdisco: and why don't wrap the function in useCallback?

src/Input/Select/Select.tsx
```diff
@@ -302,17 +315,19 @@ const Select: ISelect = (props: Props) => {
...
           placeholder={removeFloatingLabel && label}
           role="combobox"
           aria-expanded={isFocus}
           aria-autocomplete="list"
           status={deprecatedStatus}
           disabled={disabled}
-          onFocus={handleFocus}
-          onBlur={handleFocusOut}
+          onFocus={!disableTyping ? handleFocus : null}
+          onBlur={!disableTyping ? handleFocusOut : null}
```
tingtingdisco: `disableTyping ? null : handleFocusOut` is easier to read

src/Input/Select/Select.tsx
tingtingdisco: and make this icon `cursor: pointer`

src/Input/Select/Select.test.tsx
```diff
@@ -639,6 +639,75 @@ describe('<Select/> disableTyping', () => {
   });
 });
 
+describe('<Select/> defaultOpen', () => {
+  it('should be open', async () => {
+    const Component = () => (
+      <React.Fragment>
```
tingtingdisco: haha, you copied my redundant fragment

src/Input/Select/Select.tsx
```diff
@@ -333,6 +334,7 @@ const Select: ISelect = (props: Props) => {
           small={small}
           disableTyping={disableTyping}
           readOnly={disableTyping}
+          autoFocus={defaultOpen}
```
tingtingdisco, line 336: what will happen if we have 2 Select components on the same page, and they are both `defaultOpen = true`?

src/Input/Select/Select.tsx
```diff
@@ -122,6 +122,18 @@ const Select: ISelect = (props: Props) => {
...
   }, []);
 
+  React.useEffect(
+    function indicateBodyIsClickable() {
+      if (isFocus) {
+        // This indicates the document is clickable, so that onClickOutside works on ios
+        document.body.style.cursor = 'pointer';
+      }
+
+      return () => document.body.removeAttribute('style');
```
tingtingdisco: on desktop this will make the whole page has pointer cursor, and we can not make sure we can do this `document.body.removeAttribute('style')`

please try other solution.

tingtingdisco (changes requested):
indicateBodyIsClickable is not a good solution

src/Input/Select/Select.tsx
rosslo: Because it is passed to an `input` rather than `PureComponent`. 🤔

src/Input/Select/Select.tsx
rosslo: Yes, so usually we won't just pass `ture` to `defaultOpen` when there are multip Selects.
e.g. `<Select defaultOpen={isLastSelect} />`

rosslo:
> Any projects need defaultOpen?
> if onFocus is refactored to onClick, we should still keep the onFocus logic.

- In the new onboarding flow, we will auto open the last Select added by the user. 
- I think for disableTyping Select, the onFocus doesn't make sense, because there won't be a cursor in input when the dropdown is open. Btw, I also checked there is no use of `onFocus` in both projects.

src/Input/Select/Select.test.tsx
rosslo: I'd like to just rename this test block as you suggested but without moving the other new tests to it. Because the others are in `<Select disableTypeing />` and `<Select defaultOpen />` which I think is categorized enough.

src/Input/Select/Select.tsx
tingtingdisco: yes, you are right.

tingtingdisco (approved):
LGTM

PR #372, opened 2020-06-10
Gallery: Add a prop to decide if the image modal should be opened by default

ritikamotwani:
Add a prop, `isModalVisible` to decide if the image modal should be open by default.

### DIFF
--- a/src/Application/Gallery/Gallery.tsx
+++ b/src/Application/Gallery/Gallery.tsx
@@ -18,14 +18,13 @@ const defaultImagesDisplayed = 8;
 class Gallery extends React.Component<Props, State> {
   sliderRef: React.RefObject<Slider>;
 
-  state = {
-    visible: false,
-    currentIndex: 1,
-    imageLeft: 0,
-  };
-
   constructor(props: Props) {
     super(props);
+    this.state = {
+      visible: props.initialVisibility,
+      currentIndex: 0,
+      imageLeft: 0,
+    };
     this.sliderRef = React.createRef();
   }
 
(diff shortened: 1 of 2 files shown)


src/Application/Gallery/Gallery.tsx
```diff
@@ -55,12 +55,19 @@ class Gallery extends React.Component<Props, State> {
   };
 
   componentDidMount() {
-    const { children, imagesDisplayed = defaultImagesDisplayed } = this.props;
+    const {
+      children,
+      imagesDisplayed = defaultImagesDisplayed,
+      isModalVisible = false,
```
tingtingdisco: @ritikamotwani do you think this works?
``` 
constructor(props: Props) {
    super(props);
    this.sliderRef = React.createRef();
    this.state = {
      visible: props.initialVisibility,
      currentIndex: 1,
      imageLeft: 0,
    };
}
```

ritikamotwani: No, it works fine in case when the modal is not visible initially. But for the other case, the 2nd image opens, and the arrow keys do not work when the modal opens on the initial render. When I do `currentIndex: 0`, then the first image opens correctly on the initial render, but the arrow keys do not work.
They start working when I close the modal, and reopen it by clicking on an image.

ritikamotwani: The arrow keys are not working because the focus isn't set to the slider somehow, and the key down event isn't getting detected. When I click somewhere on the image, the slider window is focused and the onKeyDown event is handled again

ritikamotwani: Had to remove the comparison of prevState visibility of false from `componentDidUpdate`. Updated it now.

src/Application/Gallery/Gallery.tsx
```diff
@@ -63,9 +62,8 @@ class Gallery extends React.Component<Props, State> {
     }
   }
 
-  componentDidUpdate(prevProps: Props, prevState: State) {
+  componentDidUpdate() {
     if (
-      !prevState.visible &&
```
tingtingdisco: why need to remove this line?

tingtingdisco: I think maybe we should add this to `componentDidMount`
```
if (this.state.visible) {
      this.sliderRef.current.sliderContainerRef.current.focus();
}
```

ritikamotwani: Added this condition to componentDidMount.
But I think if this is added, we don't need the previous condition. As whenever the modal is made visible, the focus is added to the slider with this condition.
But maybe can remove the previous condition after writing tests for this component?

tingtingdisco: `componentDidMount` only being executed once, so I think we still need the condition in `componentDidUpdate` right?

Yes adding test is needed, please add thank you!

ritikamotwani: Yes yes, sorry wrote the reply wrong, I have added it to componentDidUpdate only!

ritikamotwani: Added to `componentDidMount` and opened an issue https://github.com/glints-dev/glints-aries/issues/374

tingtingdisco (approved):
LGTM

Issue #350, opened 2020-05-27
Select: Pressing Enter on No Results Found Option Throws Error


westwood846:
```
vendors~main.2b252dcf4a7d20866a9f.bundle.js:1 Uncaught TypeError: Cannot read property 'textContent' of null
```

when pressing enter key on the _No results found_ pseudo-option. Can easily be reproduced by typing into the Select until the _No results found_ text appears and then pressing enter on it. There should be no error message.

Bugsnag report: https://glints.slack.com/archives/CQMM9C2N7/p1590542138009100

![Screenshot 2020-05-27 at 10 53 00](https://user-images.githubusercontent.com/3341028/82972887-a95f5080-a008-11ea-95d5-a3b6399be3c4.png)

PR #375, opened 2020-06-17
Select: Abort option select handler if there is no active option

westwood846:
This was causing a TypeError when attempting to read the `textContent` from
`undefined` in the next line. When enter is pressed on the 'no results' option, `handleClickOnOption` is called without `e`, so `getActiveElement` will search for an element with the `.active` class. But the 'no results' option is never active, so `getActiveElement` returns `undefined` in that case.

Please read the bug report at #350 for context.


tingtingdisco (approved):
LGTM

PR #378, opened 2020-06-17
Slider: Convert to functional component

### LINKED ISSUE #379: Gallery Depends on Deprecated Slider Behaviour
When converting the Slider to a functional component, I was assuming that `initialItem` would only be relevant on the first render and implemented it as a default value for the `index` state. This makes sense name-wise and is how it's described in the Storybook. Unfortunately `Gallery` is using that field to control the Slider's state at all times, which does not work anymore after my conversion:

```
<Slider
  initialItem={currentIndex + 1}
  arrowWhite
  removeDots
  afterChange={this.getCurrentIndex}
  containerRef={this.sliderRef}
>
```

We might have to add a prop to Slider to make it controllable.

westwood846:
Due to the poor state handling and redundancies in this component, the conversion ended up being almost a complete rewrite.

Most notably I got rid of the `translateValue` state, which really is just a value that can be easily be calculated from `index`. That enabled me to remove the `screenSize` state too, which is in essence just a mostly static value that gets used in when calculating `translateValue`. Also I made the autoplay interval use the same logic as `nextSlide` which just needed an additional param to specify whether the requested page change should loop back to the first page or not when on the last page. Also makes the page indexing start from zero.

During the conversion, I noticed that the conversion broke the Gallery, which uses the Slider internally. Turns our Gallery was using `initialItem` to take full control of the Slider (not just to set the initial item). During the conversion I assumed that `initialItem` was really just to set the initial item, so the Gallery couldn't control the state anymore. That's why I've added the `index` prop to Slider, which can be used to take full control, while the `initialItem` can continue to be used for setting the initial item for the internal uncontrolled state.

DST will require one tiny migration for this: https://gitlab.glints.com/glints/glints-dst/blob/feca2b1384e806399c0cecb0045cdd7ee20f915c/app/components/Testimonial/TestimonialSlider/TestimonialSlider.tsx#L71 uses the `initialItem`. After upgrading Aries, the prop can just be removed to make the component fully uncontrolled.

This PR adapts the Gallery, so it fixes #379.

### DIFF
--- a/src/Display/Slider/Slider.tsx
+++ b/src/Display/Slider/Slider.tsx
@@ -45,32 +45,37 @@ const Slider = ({
 
   const [index, setIndex] = React.useState<number>(initialItem || 0);
 
+  const setSlide = React.useCallback(
+    (index: number, shouldLoop = false) => {
+      setIndex(currentIndex => {
+        if (index < 0) {
+          index = 0;
+        }
+        if (index >= childrenCount) {
+          index = shouldLoop ? 0 : childrenCount - 1;
+        }
+        if (afterChange !== undefined && currentIndex !== index) {
+          afterChange(index);
+        }
+        return index;
+      });
+    },
+    [childrenCount, afterChange]
+  );
+
   const previousSlide = () => {
-    if (index > 0) {
-      setIndex(index - 1);
-    }
+    setSlide(index - 1);
   };
 
   const nextSlide = React.useCallback(
     (loop = false) => {
-      setIndex(index =>
-        index >= childrenCount - 1 ? (loop ? 0 : index) : index + 1
-      );
+      setSlide(index + 1, loop);
     },
-    [childrenCount]
+    [index, setSlide]
   );
 
-  React.useEffect(
-    function callAfterChange() {
-      if (afterChange !== undefined) {
-        afterChange(index);
-      }
-    },
-    [afterChange, index]
-  );
-
-  const handleDotClick = (idx: number) => {
-    setIndex(idx);
+  const handleDotClick = (index: number) => {
+    setSlide(index);
   };
 
   const handleKeyDown = (e: React.KeyboardEvent) => {
--- a/src/Display/Slider/Slider.tsx
+++ b/src/Display/Slider/Slider.tsx
@@ -28,9 +28,11 @@ const Slider = ({
   fullContent,
   arrowWhite,
   removeDots,
+  containerRef,
 }: Props) => {
   const interval = React.useRef<ReturnType<typeof setTimeout>>();
-  const sliderContainerRef = React.useRef<HTMLDivElement>();
+  const privateContainerRef = React.useRef<HTMLDivElement>();
+  const sliderContainerRef = containerRef || privateContainerRef;
 
   const childrenCount = React.Children.toArray(children).filter(
     child => !isNil(child)
@@ -178,6 +180,7 @@ export interface Props {
   removeDots?: boolean;
   afterChange?: Function;
   autoplay?: boolean;
+  containerRef?: React.RefObject<HTMLDivElement>;
 }
 
 export default Slider;
--- a/src/Display/Slider/Slider.tsx
+++ b/src/Display/Slider/Slider.tsx
@@ -24,6 +24,7 @@ const Slider = ({
   children,
   autoplay,
   initialItem,
+  index: controlledIndex,
   className,
   fullContent,
   arrowWhite,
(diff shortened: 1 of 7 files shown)


tingtingdisco:
when we set the original `translateValue` state while window resizing, the component re-renders with the new `translateValue`.

Now the state is removed and the new `forceUpdate` is basically only to re-render `translateValue`, but we lose the connection between `translateValue` and `resize` event.

I suggest to keep a `translateValue` state.

Another concern is `getBoundingClientRect` is slow, we should only call it when screen resizes, currently implementation call it at every re-render, since this is a shared component, we can also debounce the window resize event handler which calls `getBoundingClientRect`.

tingtingdisco:
I think we only need to add a `key={currentIndex}` prop to the Slider component in the Gallery component, so when Gallery's `currentIndex` changes, the Slider will be unmount and remount and be initialized with `initialItem`.

So maybe we don't need to add the `index` prop to Slider, I cannot see the need of controlled index, WDYT?

src/Display/Slider/Slider.tsx
```diff
@@ -10,234 +10,193 @@ import LeftArrow from './LeftArrow';
...
+  removeDots,
+  containerRef,
+}: Props) => {
+  const interval = React.useRef<ReturnType<typeof setTimeout>>();
+  const privateContainerRef = React.useRef<HTMLDivElement>();
+  const sliderContainerRef = containerRef || privateContainerRef;
+
+  const childrenCount = React.Children.toArray(children).filter(
+    child => !isNil(child)
+  ).length;
```
tingtingdisco, line 40: can use `useMemo` for this calculation

src/Display/Slider/Slider.tsx
```diff
@@ -10,234 +10,193 @@ import LeftArrow from './LeftArrow';
...
+  const [internalIndex, setInternalIndex] = React.useState<number>(
+    initialItem || 0
+  );
 
-      if (afterChange !== undefined) {
-        afterChange(index - 1);
-      }
-    }
-  };
+  const index = isNil(controlledIndex) ? internalIndex : controlledIndex;
```
tingtingdisco, line 63: I think we can have an `useEffect` for the controlled index, and I guess it will be easier to understand logics between controlled and uncontrolled indexes.
```
useEffect(() => {
  setInternalIndex(controlledIndex);
}, [controlledIndex] )
```

tingtingdisco: but I think the real problem is the original implementation initialized many internal states only at `componentDidMount`, so when `children` or `initialItem` changes, the component might not work anymore.

src/Display/Slider/Slider.tsx
```diff
@@ -10,234 +10,193 @@ import LeftArrow from './LeftArrow';
...
-    const childrenCount = React.Children.toArray(children).filter(
-      child => !isNil(child)
-    ).length;
-
-    if (index !== childrenCount) {
-      this.setState({
-        index: index + 1,
-        translateValue: translateValue - screenSize,
+  const setSlide = React.useCallback(
+    (index: number, shouldLoop = false) => {
```
tingtingdisco, line 66: how to determine the value of `shouldLoop`?

src/Display/Slider/Slider.tsx
```diff
@@ -10,234 +10,193 @@ import LeftArrow from './LeftArrow';
...
+      setInternalIndex(currentIndex => {
+        if (index < 0) {
+          index = 0;
```
tingtingdisco, line 69: mutate parameter is not good, and why index can be < 0?

src/Display/Slider/Slider.tsx
```diff
@@ -10,234 +10,193 @@ import LeftArrow from './LeftArrow';
...
-      }
-    }
+  const previousSlide = () => {
+    setSlide(index - 1);
   };
 
-  setSize = () => {
-    const { index } = this.state;
+  const nextSlide = React.useCallback(
+    (loop = false) => {
```
tingtingdisco, line 88: how to determine the value of `loop`?

src/Display/Slider/Slider.tsx
```diff
@@ -10,234 +10,193 @@ import LeftArrow from './LeftArrow';
...
+      return () => {
+        window.removeEventListener('resize', forceUpdate);
+      };
+    },
+    [forceUpdate]
+  );
+
+  const sliderContainerElement = ReactDOM.findDOMNode(
+    sliderContainerRef.current
+  ) as Element;
```
tingtingdisco, line 132: do we need `ReactDOM.findDOMNode` here?

tingtingdisco (changes requested):
Suggest to only do functional component conversion, do conversion and refactor together is hard to review, and the new implementation can be improved.

westwood846:
Closing because of #461

PR #383, opened 2020-06-19
Select: Handle non-string value

tingtingdisco:
mainly to make `number` type value not to break the component, but other non-string types are also handled.

### DIFF
--- a/src/Input/Select/Select.noTypeChecking.test.tsx
+++ b/src/Input/Select/Select.noTypeChecking.test.tsx
@@ -94,8 +94,29 @@ describe('<Select /> can still work when receives non-string option children or 
       [1, '1'],
       [{ key: 'value' }, '[object Object]'],
       [[1, 2, 3], '1,2,3'],
-    ].forEach(([defaultValue, expectedValue]) => {
-      testNonStringValue([defaultValue, expectedValue]);
+    ].forEach(([value, expectedValue]) => {
+      testNonStringValue([value, expectedValue]);
     });
+  });
+
+  test('rerender with non-string value', async () => {
+    const Component = ({ value }: { value: string }) => (
+      <Select label="rerender" value={value}>
+        <Select.Option value="rerender">Rerender</Select.Option>
+      </Select>
+    );
+
+    const { queryByRole, rerender } = render(<Component value="Rerender" />);
+    const selectInput = queryByRole('combobox');
+    expect(selectInput.getAttribute('value')).toEqual('Rerender');
+
+    rerender(<Component value={1} />);
+    expect(selectInput.getAttribute('value')).toEqual('1');
+
+    rerender(<Component value={[1, 2, 3]} />);
+    expect(selectInput.getAttribute('value')).toEqual('1,2,3');
+
+    rerender(<Component value={{ key: 'value' }} />);
+    expect(selectInput.getAttribute('value')).toEqual('[object Object]');
   });
 });
--- a/src/Input/Select/Select.noTypeChecking.test.tsx
+++ b/src/Input/Select/Select.noTypeChecking.test.tsx
@@ -1,0 +1,101 @@
+// @ts-nocheck
+
+import * as React from 'react';
+import 'jest-styled-components';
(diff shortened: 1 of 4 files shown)


storybook-config/webpack.config.js
```diff
@@ -15,7 +15,11 @@ module.exports = async ({ config }) => {
   config.resolve.extensions.push('.ts');
   config.resolve.extensions.push('.tsx');
 
-  config.plugins.push(new ForkTsCheckerWebpackPlugin());
+  config.plugins.push(
+    new ForkTsCheckerWebpackPlugin({
+      reportFiles: ['**', '!**/?(*.)test.*'],
```
tingtingdisco, line 20: disabled storybook type checking on test files during storybook build process

src/Input/Select/Select.noTypeChecking.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    [
+      [1, '1'],
+      [{ key: 'value' }, '[object Object]'],
+      [[1, 2, 3], '1,2,3'],
+    ].forEach(([defaultValue, expectedValue]) => {
+      testNonStringValue([defaultValue, expectedValue]);
+    });
+  });
+});
```
rosslo, line 122: Can consider adding a test case for rerender new value prop like what you did for the test case
`<Select/> handleValueChange > input value should be the same as passed prop value`

Because you also used `String()` in `handleValueChange` `useEffect`.

src/Input/Select/Select.noTypeChecking.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+      expect(selectInput.value).toEqual(expectedValue);
+      unmount();
+    };
+
+    [
+      [1, '1'],
+      [{ key: 'value' }, '[object Object]'],
+      [[1, 2, 3], '1,2,3'],
+    ].forEach(([defaultValue, expectedValue]) => {
+      testNonStringValue([defaultValue, expectedValue]);
```
rosslo: I assume you would like to name the first argument to `value` according to the use of the test case above.

src/Input/Select/Select.tsx
```diff
@@ -54,12 +56,30 @@ const Select: ISelect = (props: Props) => {
     [status]
   );
 
+  // warn non-string type value, defaultValue
+  React.useEffect(
+    function warnNonStringValueAndDefaultValue() {
+      const isValueNotString = value && typeof value !== 'string';
+      const isDefaultValueNotString =
+        defaultValue && typeof defaultValue !== 'string';
```
rosslo: I think we can just leave `typeof value !== 'string';` without `value &&`. It will be more intuitive to me.

Because I was thinking about whether it works when the value is an empty string or `true`.

src/Utils/canWarn.ts
```diff
@@ -0,0 +1,2 @@
+export const canWarn =
+  typeof console !== 'undefined' && process.env.NODE_ENV === 'development';
```
rosslo, line 2: This util will be useful, cool!

src/Input/Select/Select.noTypeChecking.test.tsx
tingtingdisco: ok

src/Input/Select/Select.noTypeChecking.test.tsx
tingtingdisco: right...

src/Input/Select/Select.tsx
tingtingdisco: value and defaultValue are optional, so I think we should not warn when they are undefined. So maybe I make it 
```
const shouldWarnValue = typeof value !== 'undefined' && typeof value !== 'string';
```

WDYT?

rosslo: Yeah, that'd be more clear. 👍

tingtingdisco:
@rosslo updated please review thank you!

rosslo (approved):

PR #386, opened 2020-06-23
Co locate styles

westwood846:


rosslo:
css lint task failed https://github.com/glints-dev/glints-aries/pull/386/checks?check_run_id=797782120 I think we need to update the filename pattern for stylelint.

rosslo (approved):

PR #398, opened 2020-07-03
Typography: Fix UI mismatches in the current typography component

### LINKED ISSUE #202: Typography: Fix UI Mismatches
GDS: https://share.goabstract.com/49b91e60-f84d-47e1-bf65-d81d6b8bb9a1

ritikamotwani:
Fixes: #202 

For **title** the line height is different for upper case and lower case titles, so added a upperCase prop to check.

For **paragraph** the line height values are set only for one of the two projects according to design, so added a lineHeight prop, to check if lineHeight should be set for the paragraph.

Doc: https://docs.google.com/document/d/1pHImZ2YXcG0WxMU5ll1y9zBygyErR2EjR8xu3Y2lgDw/edit

Abstract: https://app.abstract.com/projects/da6f9bdf-2057-45bf-9dda-8dbf0d261eac/branches/master/commits/latest/files/DDF7C217-39F9-45D7-8C53-8D8C88D5FC37/layers/02442D8E-2A5E-4B3E-8A67-BE74ACFAC7DD?collectionId=dbdd8e8c-4db0-40f5-b871-69245a509a63&collectionLayerId=7911adde-85ae-4992-8c0a-7969529f2b73

### DIFF
--- a/src/General/Typography/Paragraph.tsx
+++ b/src/General/Typography/Paragraph.tsx
@@ -15,6 +15,7 @@ const Paragraph: React.FunctionComponent<Props> = props => {
     bold,
     color = SecondaryColor.black,
     ellipsis,
+    lineHeight = false,
     ...restProps
   } = props;
 
@@ -25,6 +26,7 @@ const Paragraph: React.FunctionComponent<Props> = props => {
       bold={bold}
       color={color}
       ellipsis={ellipsis}
+      lineHeight={lineHeight}
       {...restProps}
     >
       {children}
@@ -35,6 +37,7 @@ const Paragraph: React.FunctionComponent<Props> = props => {
 interface Props extends ParagraphProps {
   className?: string;
   children: React.ReactNode;
+  lineHeight?: boolean;
 }
 
 export default Paragraph;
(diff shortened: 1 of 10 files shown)


src/General/Typography/Title.tsx
```diff
@@ -35,6 +37,7 @@ interface Props {
   color?: string;
   ellipsis?: boolean;
   tag?: tagType;
+  upperCase?: boolean;
```
tingtingdisco: uppercase is a single word not upper and case right?

src/General/Typography/Paragraph.tsx
```diff
@@ -15,6 +15,7 @@ const Paragraph: React.FunctionComponent<Props> = props => {
     bold,
     color = SecondaryColor.black,
     ellipsis,
+    lineHeight = false,
```
tingtingdisco: 1. it's hard to understand what this prop does, maybe name it `shouldSetLineHeight`?
2. why do we need this prop? when we use this component, why we want the line-height to be `normal`?

ritikamotwani: The design has two types of paragraph, different for employers and candidates sites. One has line-height, for the other it is normal

src/General/Typography/Title.tsx
ritikamotwani: Yes, I don't know why did I break it into two!

src/General/Typography/Paragraph.tsx
ritikamotwani: Updated

tingtingdisco (approved):
LGTM

PR #403, opened 2020-07-09
Tab: Fix the enum issue and make children optional

### LINKED ISSUE #401: Tab: make children prop optional & fix enum issues
image

ritikamotwani:
Fixes: #401

### DIFF
--- a/src/Display/Tabs/Tabs.tsx
+++ b/src/Display/Tabs/Tabs.tsx
@@ -109,12 +109,12 @@ Tabs.Pane = TabPane;
 
 interface Props {
   children: React.ReactNode;
-  variant?: string;
-  alignment?: string;
+  variant?: 'underlined' | 'colored';
+  alignment?: 'horizontal' | 'vertical';
+  theme?: 'blue' | 'black';
   activeTab?: string | number;
   onTabClick?(tab: React.ReactText | React.ReactNode): void;
   className?: string;
-  theme?: string;
 }
 
 export default Tabs;
--- a/src/Display/Tabs/Tabs.tsx
+++ b/src/Display/Tabs/Tabs.tsx
@@ -7,6 +7,10 @@ import TabPane, { Props as TabPaneProps } from './TabPane';
 
 import { TabsContainer, TabsHeader, TabsBody } from './TabsStyle';
 
+export type Variant = 'underlined' | 'colored';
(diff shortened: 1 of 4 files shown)


src/Display/Tabs/Tabs.tsx
```diff
@@ -112,12 +109,12 @@ Tabs.Pane = TabPane;
 
 interface Props {
   children: React.ReactNode;
-  variant?: EHorizontalTabVariant;
+  variant?: string;
```
rosslo: @ritikamotwani could you help experiment the following?
```
type Variants = keyof typeof EHorizontalTabVariant;
variant?: Variants;
```
I was thinking if we can have `variant? 'underlined' | 'colored'`. It might help us throw the error when we pass a wrong `variant` value in other projects.

ritikamotwani: Oh I think `variant? 'underlined' | 'colored'` would be better

ritikamotwani: Updated with `variant?: 'underlined' | 'colored'`
It would throw an error, if the value is incorrect for variant, theme, alignment

rosslo: I thought `type Variants = keyof typeof EHorizontalTabVariant;` equals `'underlined' | 'colored'`. So we can just need to update one place (enum) when we have a new value.

Btw, I think we can still have `type Variant = ''underlined' | 'colored'` and `variants? Variant` instead of setting `varaints? 'underlined' | 'colored'` everywhere.

ritikamotwani: The reason I didn't use this is, when I write this it becomes `"UNDERLINED" | "COLORED"`, and then if I try using `typeof ETabAlignment[keyof typeof ETabAlignment]` it's the same as ETabVariant type (https://stackoverflow.com/questions/53662208/types-from-both-keys-and-values-of-object-in-typescript)

rosslo: Ohhh right, it is `keyof` 😅 , then what do you think of using `type Variant = ''underlined' | 'colored'`?

ritikamotwani: Yes, that's much better! Thanks

rosslo (approved):

PR #408, opened 2020-07-16
Tooltip: fix null pointer of contentRef.current

### LINKED ISSUE #368: Tooltip: null pointer of contentRef.current
happened on https://glints.com/id/opportunities/jobs/recommended

image

rosslo:
I suspected the cause is because the `hideTooltipIfTouchOutside` was triggered twice if the user clicked the trigger button when Tooltip was showing. However, I cannot reproduce it so I added the check for `contentRef.current` to be safer.

### DIFF
--- a/src/Display/Tooltip/Tooltip.tsx
+++ b/src/Display/Tooltip/Tooltip.tsx
@@ -22,9 +22,9 @@ const Tooltip: React.FunctionComponent<Props> = ({
 
   const hideTooltipIfTouchOutside = React.useCallback(
     (event: TouchEvent | React.TouchEvent) => {
-      const hasTouchedOutsideOfTooltipContent = !contentRef.current.contains(
-        event.target as HTMLElement
-      );
+      const hasTouchedOutsideOfTooltipContent =
+        contentRef.current &&
+        !contentRef.current.contains(event.target as HTMLElement);
 
       if (hasTouchedOutsideOfTooltipContent) {
         hideTooltip();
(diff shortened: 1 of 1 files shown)


src/Display/Tooltip/Tooltip.tsx
```diff
@@ -22,9 +22,9 @@ const Tooltip: React.FunctionComponent<Props> = ({
 
   const hideTooltipIfTouchOutside = React.useCallback(
     (event: TouchEvent | React.TouchEvent) => {
-      const hasTouchedOutsideOfTooltipContent = !contentRef.current.contains(
-        event.target as HTMLElement
-      );
+      const hasTouchedOutsideOfTooltipContent =
+        contentRef.current &&
```
ritikamotwani, line 26: Should we also check isShow here? Because this condition should only be checked if the tooltip is shown (incase the event happens when isShow is true) ?

ritikamotwani:
@rosslo , just a comment added. Cannot reproduce the issue opened haha :P 
But yes, the check for the ref looks good.

src/Display/Tooltip/Tooltip.tsx
rosslo: Since `hideTooltipIfTouchOutside` is wrapped in `useCallback` already, I would not like to add `isShow` in dependency that will recreate  `hideTooltipIfTouchOutside` function when `isShow` is updated.

ritikamotwani: Oh I didn't see the useCallback, I thought it's a function. Yes, makes sense else it would be called every time when isShow updates.

ritikamotwani (approved):

PR #416, opened 2020-07-20
Storybook: read version from package.json

tingtingdisco:
To display the correct version.


stories/IntroStory.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
+const { version } = require('../package.json');
```
tingtingdisco, line 2: not sure why `process.env.npm_package_version` is undefined...

franciscrispin: Oh cool I didn't know you can do this haha

franciscrispin (approved):
@tingtingdisco LGTM!

PR #443, opened 2020-07-27
Remove Unused Components

westwood846:
See https://github.com/glints-dev/glints-aries/projects/2


ritikamotwani (approved):
Checked the storybook for all components. Everything works fine. 
Will you also move the components imported > 1 times to the projects?
 Or you can also not remove those components in this PR and we can decide about them in the sprint meeting?

westwood846:
> Will you also move the components imported > 1 times to the projects?
> Or you can also not remove those components in this PR and we can decide about them in the sprint meeting?

I think let's leave this PR unmerged until the SPM and until somebody has time to add the removed components to DST. Just so we don't block releases.

ritikamotwani:
> I think let's leave this PR unmerged until the SPM and until somebody has time to add the removed components to DST. Just so we don't block releases.

Sure!

PR #445, opened 2020-07-29
Add type as a prop to buttons, with default values as button

### LINKED ISSUE #405: Button: should be able to receipt prop type
Typescript complaints prop type
Screen Shot 2020-07-10 at 5 51 15 PM

ritikamotwani:
Fixes: #405

### DIFF
--- a/src/General/Button/Button.tsx
+++ b/src/General/Button/Button.tsx
@@ -16,6 +16,7 @@ const SOLID_BUTTON_THEME_MAP = {
   [ButtonVariant.SOLID_WHITE]: ButtonTheme.WHITE,
   [ButtonVariant.SOLID_BLUE]: ButtonTheme.BLUE,
 };
+type buttonType = 'button' | 'submit' | 'reset';
 
 export const transformVariant = (variant: string, theme?: string) => {
   if (theme) {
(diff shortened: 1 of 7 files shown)


src/General/Button/Button.tsx
```diff
@@ -16,6 +16,7 @@ const SOLID_BUTTON_THEME_MAP = {
   [ButtonVariant.SOLID_WHITE]: ButtonTheme.WHITE,
   [ButtonVariant.SOLID_BLUE]: ButtonTheme.BLUE,
 };
+type buttonType = 'button' | 'submit' | 'reset';
```
ritikamotwani: Have added a new proptype, with default value as 'button'. Didn't want to directly add type='button' to as the types can be these too?

tingtingdisco: @ritikamotwani can we make our button prop to extend `React.ButtonHTMLAttributes<HTMLButtonElement>`? I think we can make aries button to be able to receive all html button attributes.

ritikamotwani: Yes, then the `type` by default has the types button, submit and reset.
But I still have set the default type as button else it would be submit from what I read.

ritikamotwani:
@tingtingdisco updated it

tingtingdisco (approved):
LGTM

PR #447, opened 2020-08-03
Fix/select disable typing onblur

### LINKED ISSUE #404: Select: still need onBlur when disableTyping is true
We still need onBlur to be called when disableTyping is true, otherwise we cannot do field validation.

rosslo:


src/Input/Select/Select.disableTyping.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+import '@testing-library/jest-dom/extend-expect';
+import { fireEvent, render, act } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import Select from './Select';
+
+describe('<Select disableTyping> onBlur', () => {
+  it('should call onBlur when close menu', () => {
+    const onBlurSpy = jest.fn();
+    const Component = () => (
```
ritikamotwani: Should this component be made common as these tests are only for checking onBlur is called when disabled typing is true?

ritikamotwani (approved):

src/Input/Select/Select.disableTyping.test.tsx
rosslo: Updated!

PR #451, opened 2020-08-14
Select: Fix null active element caused by dynamic select options

### LINKED ISSUE #442: Select's scrollToActiveElement results in null error
image

`getActiveElement()` becomes null when user inputs up arrow key

Related dst bug: https://glints.atlassian.net/browse/GM-494

franciscrispin:
This bug only occurs for a dynamic list i.e. the number of items in the select list changes

- When the nth item in the list is selected, the active index is n
- When the SelectList is opened again, there may be < n items so there is no active item
- For this case, we need to reset the active index to 0 to ensure that there is an 'active' item in the SelectList.

In this example, when I open the select list the second time, the first item in the list is not `active`
https://www.loom.com/share/33571f7ae01941f39e7c8544bf13cc24

Resolves https://github.com/glints-dev/glints-aries/issues/442

### DIFF
--- a/src/Input/Select/Select.tsx
+++ b/src/Input/Select/Select.tsx
@@ -127,10 +127,19 @@ const Select: ISelect = (props: Props) => {
   );
 
   React.useEffect(
-    function updateOptions() {
+    function updateOptionsAndActiveIndex() {
       setOptions(availableOptions);
+
+      if (!isFocus) {
+        const selectedOptionIndex = availableOptions.findIndex(option =>
+          toLower(option.props.children).includes(toLower(inputValue))
+        );
+        if (activeOptionIndex !== selectedOptionIndex) {
+          setActiveOptionIndex(selectedOptionIndex);
+        }
+      }
     },
-    [availableOptions]
+    [activeOptionIndex, availableOptions, inputValue, isFocus]
   );
 
   React.useEffect(function registerClickOutsideListener() {
--- a/src/Input/Select/select.dynamicOptions.test.tsx
+++ b/src/Input/Select/select.dynamicOptions.test.tsx
@@ -1,0 +1,52 @@
+import * as React from 'react';
+import 'jest-styled-components';
+import '@testing-library/jest-dom/extend-expect';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import Select from './Select';
+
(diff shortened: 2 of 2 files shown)


src/Input/Select/Select.test.tsx
```diff
@@ -736,3 +736,48 @@ describe('<Select.Option/>', () => {
     expect(asFragment()).toMatchSnapshot();
   });
 });
+
+const DynamicSelect = () => {
```
rosslo: Can create a `select.dynamicOptions.test.tsx` for this test case to avoid the `Select.test.tsx` file becoming too big and separate different tests clearly.

src/Input/Select/Select.tsx
```diff
@@ -162,6 +162,19 @@ const Select: ISelect = (props: Props) => {
...
+      // When the nth item in the list is selected, the active index is n
+      // When the SelectList is opened again, there may be < n items if the
+      // list is filtered. For this case, we need to reset the active index
+      // to 0 to ensure that there is an 'active' item in the SelectList.
+      if (activeOptionIndex > options.length - 1) {
+        setActiveOptionIndex(0);
+      }
+    },
+    [activeOptionIndex, options.length]
+  );
```
rosslo: I think it will be safer to `findActiveOptionIndex` before update `activeOptionIndex`  instead of `setActiveOptionIndex(0)`.

And can consider handling this logic within current `updateOptions` `useEffect`.
```
  React.useEffect(
    function updateOptionsAndActiveIndex() {
      setOptions(availableOptions);

      const findActiveOptionIndex = availableOptions.findIndex(option =>
        toLower(option.props.children).includes(toLower(inputValue))
      );
      if (activeOptionIndex !== findActiveOptionIndex) {
        setActiveOptionIndex(findActiveOptionIndex);
      }
    },
    [availableOptions]
  );
```

franciscrispin: Seems like this will break some tests if I pass in the full dependency array `[activeOptionIndex, availableOptions, inputValue]`. Oh and it doesn't work with the current hover behaviour. Will see if there's a safer way.

franciscrispin: Okay it works! Just needed to add a condition. Thanks!

rosslo (approved):

PR #455, opened 2020-08-25
feat: Add RSS Icon

Flashrob:
Added RSS Icon to be used for RSS Modal on Employer ATS page


src/index.ts
```diff
@@ -184,3 +184,4 @@ export { default as WarningOutlineIcon } from './General/Icon/components/Warning
 export { default as WarningSolidIcon } from './General/Icon/components/WarningSolidIcon';
 export { default as WhatsappIcon } from './General/Icon/components/WhatsappIcon';
 export { default as YoutubeIcon } from './General/Icon/components/YoutubeIcon';
+export { default as RssIcon } from './General/Icon/components/RssIcon';
```
westwood846: This list of exports should be sorted alphabetically ;)

westwood846 (changes requested):

src/index.ts
Flashrob: Fixed!

westwood846 (approved):

PR #456, opened 2020-08-25
Button: Remove the default button type

ritikamotwani:
Removed the explicitly defined button type.
It would keep the default value as submit.
Also now that the button extends `React.ButtonHTMLAttributes<HTMLButtonElement>` , typescript won't complain on having type as a prop.


rosslo (approved):

rosslo:
@ritikamotwani LGTM!

PR #459, opened 2020-08-28
ToggleSwitch: add test cases

rosslo:


westwood846 (changes requested):
Overall good PR but I think we should test the cases where one or both `iconOptions` are undefined as well.

westwood846 (approved):
Approving already but you'll want to remove the unused `has` import in `src/Input/ToggleSwitch/ToggleSwitch.tsx`.

PR #461, opened 2020-08-31
Functional slider 2

westwood846:
Another attempt at converting the Slider to a functional component, this time without changing any of its logic.

Problem: The Gallery tests test the Slider's width. I only know one way to mock the Slider's container's width, which is by using a spy on `React.useRef`. That doesn't work for that test because Gallery also uses a Modal which contains another `useRef` call (that call obviously breaks when it tries to use the `mockContainer` mock return value. This is why the tests fail.

### DIFF
--- a/src/Display/Slider/Slider.tsx
+++ b/src/Display/Slider/Slider.tsx
@@ -10,31 +10,29 @@ import LeftArrow from './LeftArrow';
 import RightArrow from './RightArrow';
 import SliderItem from './SliderItem';
 
-class Slider extends React.Component<Props, State> {
-  static Item = SliderItem;
+const Slider = ({
+  afterChange,
+  children,
+  autoplay,
+  initialItem,
+  className,
+  fullContent,
+  arrowWhite,
+  removeDots,
+  containerRef,
+}: Props) => {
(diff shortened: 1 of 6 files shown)


tingtingdisco:
maybe we can mock `ReactDOM.findDOMNode` so `getSliderContainerDOMNode` can return a mocked dom node with a mocked `getBoundingClientRect` method?

src/Display/Slider/Slider.tsx
```diff
@@ -10,218 +10,193 @@ import LeftArrow from './LeftArrow';
...
+            setTranslateValue(-(screenSize * index));
+          } else {
+            setIndex(1);
+            setTranslateValue(0);
+          }
+        }, 6000);
+      }
+      window.addEventListener('resize', setSize);
+
+      return function componentWillUnmount() {
```
tingtingdisco: this event listener is being registered and unregistered when `index` or `screenSize` changes, so maybe call this function `unregisterAutoplay`?

tingtingdisco (approved):
LGTM, can try to mock `ReactDOM.findDOMNode` to get the Gallery test working.

PR #462, opened 2020-08-31
Pointing modal tests

westwood846:


src/Display/PointingModal/PointingModalFooter.tsx
```diff
@@ -17,7 +17,7 @@ const PointingModalFooter: React.FunctionComponent<Props> = props => {
 
 interface Props {
   children: React.ReactNode;
-  className: string;
+  // className: string;
```
rosslo: Oh, why do we need to comment out this line?

I guess maybe we just need to make it optional with a question mark.

rosslo (approved):

PR #472, opened 2020-09-08
Drawer: Add position prop

### LINKED ISSUE #458: Drawer: Should be able to choose side
Currently the drawer is always on the right side of the screen but I need it on the left side.

westwood846:
Closes #458

### DIFF
--- a/src/Navigation/Drawer/Drawer.test.tsx
+++ b/src/Navigation/Drawer/Drawer.test.tsx
@@ -4,30 +4,34 @@ import * as renderer from 'react-test-renderer';
 import '@testing-library/jest-dom/extend-expect';
 import { fireEvent, render, queryByTestId } from '@testing-library/react';
 
-import Drawer from './Drawer';
+import Drawer, { DrawerPosition } from './Drawer';
 
 describe('<Drawer /> render', () => {
-  test('should match snapshot when open', () => {
-    const snapshot = renderer
-      .create(
-        <Drawer isOpen={true} onClose={jest.fn()}>
-          test-content
-        </Drawer>
-      )
-      .toJSON();
-    expect(snapshot).toMatchSnapshot();
-  });
+  const matchSnapshot = (position: DrawerPosition) => {
+    test(`should match snapshot when position is ${position} and drawer is open`, () => {
+      const snapshot = renderer
+        .create(
+          <Drawer isOpen={true} position={position} onClose={jest.fn()}>
+            test-content
+          </Drawer>
+        )
+        .toJSON();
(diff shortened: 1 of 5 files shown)


src/Navigation/Drawer/Drawer.test.tsx
```diff
@@ -28,6 +28,50 @@ describe('<Drawer /> render', () => {
...
+  test('should match snapshot when position is left and drawer is closed', () => {
+    const snapshot = renderer
+      .create(
+        <Drawer isOpen={false} position="left" onClose={jest.fn()}>
+          test-content
+        </Drawer>
+      )
+      .toJSON();
+    expect(snapshot).toMatchSnapshot();
+  });
```
ritikamotwani: @westwood846 , for these tests can we have just two tests for when the drawer is open and when it is closed, with dynamic positions.
Examplae:
```
describe('<Drawer /> prop position when drawer is open', () => {
  const matchSnapshot = (position: string) => {
    it(`should match snapshot when position = ${position} is passed`, () => {
      const { asFragment } = render(render the component);
    });
  };

  ['left', 'right', undefined].forEach(matchSnapshot);
});
```

ritikamotwani (commented):
@westwood846 , it would be better to update the storybook with the new prop too.

src/Navigation/Drawer/DrawerStyle.ts
```diff
@@ -64,15 +84,34 @@ export const DrawerWrapper = styled.div<DrawerWrapperProps>`
...
+      left: open ? 'translateX(0%)' : 'translateX(-100%)',
+    }[position])};
   outline: none;
   box-shadow: -2px 0 8px rgba(0, 0, 0, 0.15);
   overflow-y: auto;
 `;
 
 interface DrawerWrapperProps {
   open: boolean;
+  position: 'left' | 'right';
```
ritikamotwani: Would be better to create a type for this in the component, and use it here. 
In case of changes it would be easier to update just one type, instead of updating in the style and the component files.

src/Navigation/Drawer/Drawer.test.tsx
westwood846: Done!

src/Navigation/Drawer/DrawerStyle.ts
westwood846: Done!

ritikamotwani:
@westwood846 , could you add the position prop in the storybook, it would be easier for other devs to know.  Adding an example along is optional. 😛 

Rest LGTM

PR #478, opened 2020-09-22
Spacing & Box

### LINKED ISSUE #199: Spacing: New System
**BLOCKED: GDS is not ready for this component. Before working on this, please wait until you have the designs.**

Track the status here https://docs.google.com/spreadsheets/d/1O9WY_x9MmzRIbskPE-2_TsKgOz3N7C61iEqcx6zWln8/edit#gid=1560414742

rosslo:
For the case using `margin: 0 auto` to horizontal center UI.

I prefer
```
<Box my={0} mx="auto" />
```
to
```
<Box m="0 auto" />
```

Because I think to accept `0 auto` value will make the type check complicated. 

And actually I think we will just need `<Box mx="auto" />` for this case.

### DIFF
--- a/src/Layout/Box/index.tsx
+++ b/src/Layout/Box/index.tsx
@@ -2,17 +2,44 @@ import * as React from 'react';
 import styled from 'styled-components';
 import { space } from 'styled-system';
 
+import { ScreenSize } from '../../Utils/StyleConfig';
 import { SpacingScaleValues } from '../Spacing';
 
+// Responsive
+type BreakpointAliases =
+  | 'default'
+  | 'ms'
+  | 'mm'
+  | 'ml'
+  | 't'
+  | 'ds'
+  | 'dm'
+  | 'dl';
+export const breakpointAliasMap = {
+  ms: `${ScreenSize.mobileS}px`,
+  mm: `${ScreenSize.mobileM}px`,
+  ml: `${ScreenSize.mobileL}px`,
+  t: `${ScreenSize.tablet}px`,
+  ds: `${ScreenSize.desktopS}px`,
+  dm: `${ScreenSize.desktopM}px`,
+  dl: `${ScreenSize.desktopL}px`,
+};
+const theme = {
+  breakpoints: breakpointAliasMap,
+};
+
 // Space
+type ResponsiveSpacing = {
+  [key in BreakpointAliases]?: SpacingScaleValues;
+};
 type MarginTypes = 'm' | 'mt' | 'mb' | 'ml' | 'mr' | 'my' | 'mx';
 type MarginProps = {
-  [key in MarginTypes]?: SpacingScaleValues;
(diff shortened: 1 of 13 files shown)


src/Layout/Box/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  | 'default'
+  | 'ms'
+  | 'mm'
+  | 'ml'
+  | 't'
+  | 'ds'
+  | 'dm'
+  | 'dl';
+export const breakpointAliasMap = {
+  ms: `${ScreenSize.mobileS}px`,
```
ritikamotwani: @rosslo , should we name the keys as mobileS, mobileM instead of just characters, I find them a bit harder to read haha. But I'm okay if that's the norm we have already decided to follow fo the naming.

src/Layout/Spacing/index.tsx
```diff
@@ -0,0 +1,30 @@
+export type SpacingScaleTypes =
```
ritikamotwani, line 1: Oh I didn't understand the purpose for this?

src/Layout/Box/index.tsx
ritikamotwani: Oh is it because of the library we're using supports the naming like this https://styled-system.com/guides/build-a-box/

src/Layout/Box/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  ds: `${ScreenSize.desktopS}px`,
+  dm: `${ScreenSize.desktopM}px`,
+  dl: `${ScreenSize.desktopL}px`,
+};
+const theme = {
+  breakpoints: breakpointAliasMap,
+};
+
+// Space
+type ResponsiveSpacing = {
```
ritikamotwani: @rosslo , should we define all types in a separate file? Would it be more readable?

src/Layout/Box/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  [key in BreakpointAliases]?: SpacingScaleValues | 'auto';
```
ritikamotwani: Can we use keyof typeof breakpointAliasMap here instead of a new type?

src/Layout/Box/index.tsx
rosslo: Oh because we use acronym in spacing properties, I think it is also fine to name breakpoint in the same way for consistency. And it just reminds me that `default` might look not consistent. Will rename it to `_` as the library example below. 

https://styled-system.com/responsive-styles
image

src/Layout/Box/index.tsx
rosslo: Nope, got TS complaint 😅

src/Layout/Box/index.tsx
rosslo: Oh okay, will use the current `ScreenSize`'s keys instead.

ritikamotwani (approved):
@rosslo , LGTM

Issue #418, opened 2020-07-22
Select: cannot display translated value in input


rosslo:

westwood846:
Too complicated, won't fix. Wait for new Select component.

PR #482, opened 2020-09-28
Select: Make handleValueChange set selectedValue to option's label

westwood846:
#418

Storybook test here: https://github.com/glints-dev/glints-aries/pull/481


westwood846:
Too complicated, won't fix. Wait for new Select component.

PR #487, opened 2020-09-30
TextField: add clear button & correct disabled style

### LINKED ISSUE #484: TextField: add clean button & correct the disabled style
image

rosslo:


src/Input/TextField/TextField.test.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 
 import * as renderer from 'react-test-renderer';
 import '@testing-library/jest-dom/extend-expect';
-import { render } from '@testing-library/react';
+import { render, wait } from '@testing-library/react';
```
westwood846: Can remove `wait` here.

src/Input/TextField/TextField.tsx
```diff
@@ -39,6 +46,12 @@ const TextField: React.FunctionComponent<Props> = props => {
     setInputType(inputType === 'password' ? 'text' : 'password');
   }, [setInputType, inputType]);
 
+  const handleClearInput = (e: React.MouseEvent<SVGSVGElement>) => {
+    const event = Object.create(e);
```
westwood846, line 50: Is this for cloning the event object? If so, can we use lodash's `cloneDeep`? That's more explicit about what's happening. Also I'm a bit scared that the docs on `Object.create` say that the input will be used as the "prototype" of the new object. I generally avoid the prototype system, so I don't know exactly if this is a problem or not.

stories/Input/TextFieldStory.tsx
```diff
@@ -75,6 +67,20 @@ export const props = {
...
+      type: 'boolean',
+      defaultValue: <code>false</code>,
+      possibleValue: <code>true | false</code>,
+      require: 'no',
+      description:
+        'If allow to remove input content with clear icon. Works for type "text" only.',
+    },
+  ],
+};
+
```
westwood846: Use this below, or remove.

src/Input/TextField/TextField.test.tsx
```diff
@@ -239,3 +239,52 @@ describe('when a non-empty value is passed to', () => {
...
+        allowClear={true}
+        onChange={onChange}
+      />
+    );
+
+    const clearButton = queryByTestId('clear-button');
+    userEvent.click(clearButton);
+    expect(onChange).toHaveBeenCalledTimes(1);
+  });
+});
```
westwood846, line 312: Do you think you can add a test case for when `type="password"` and `allowClear={true}`. Current behavior in that case is that only the password icon is shown, which is good imo. But we should make that behavior explicit.

westwood846 (changes requested):

src/Input/TextField/TextField.tsx
rosslo: I think `Object.create` helps us to return the event object which includes `__proto__` same as native behavior. 
And actually I found this approach from `ant-design` sourcecode.

westwood846 (approved):

PR #488, opened 2020-09-30
Remove the "r" dependency.

yjwong:
This dependency doesn't seem to be used anywhere. Additionally, this package also does not seem to have a proper license nor code repository. The package was also last published 8 years ago.

See: https://www.npmjs.com/package/r


westwood846 (approved):
LGTM, tests, build and storybook look fine.

PR #490, opened 2020-09-30
Modal: responsive footer

### LINKED ISSUE #485: Modal: responsive footer buttons
https://app.zeplin.io/project/5efe8c616e1cfb13f088fc06/screen/5f1543a7c5b99d9c703e1693

image

rosslo:


src/Display/Modal/Modal.tsx
```diff
@@ -61,6 +68,25 @@ const Modal = (props: Props) => {
       document.removeEventListener('keydown', escapeKeyEventListener, false);
   }, [isVisible, onClose]);
 
+  React.useLayoutEffect(() => {
```
ritikamotwani: Should we name this function, as there's already one useLayoutEffect hook called, would be better to know the responsibility for this.

src/Display/Modal/Modal.tsx
```diff
@@ -61,6 +68,25 @@ const Modal = (props: Props) => {
...
+    if (footerElementCount > 1) {
+      const checkIsFooterChildrenInMultiLines = () => {
```
ritikamotwani, line 72: @rosslo ,  should this function be outside the hook. It's just weird to me that inside the hook there's a function in the hook, then it's called and then an event listener is set.

ritikamotwani:
Another question here @rosslo . Is there a need to check if the top button is a primary button? 
For example if I keep button 1 as primary and button 2 as ghost, then button 2 would be on the top.
Or maybe we can remind the designers to add this in the guideline that the display of buttons would be like, if there's a primary button then it would be at the second position.

src/Display/Modal/Modal.tsx
rosslo: ok!

src/Display/Modal/Modal.tsx
rosslo: If we put this function be outside the hook, we will need to wrap useCallback for it, I think we can just declare it inside useLayoutEffect to save a bit performance.

rosslo:
> Or maybe we can remind the designers to add this in the guideline that the display of buttons would be like, if there's a primary button then it would be at the second position.

Yeah, good suggestion, will remind designers later. 🙏

ritikamotwani:
> Yeah, good suggestion, will remind designers later. 🙏

Cool :P
I think we should open an issue to add the guidelines to storybook after the designers send a proper write up.
1. Only 1 or 2 buttons preferred to be there in the modal.
2. The primary button should be the second one, so that the devs know about it.

ritikamotwani (approved):

ritikamotwani:
LGTM @rosslo !

PR #494, opened 2020-10-07
Add size prop to icons

### LINKED ISSUE #194: Iconography: Fix UI Mismatches
Track the status here https://docs.google.com/spreadsheets/d/1O9WY_x9MmzRIbskPE-2_TsKgOz3N7C61iEqcx6zWln8/edit#gid=1560414742

https://app.zeplin.io/project/5efe8c616e1cfb13f088fc06/screen/5f1547d8aaabad9c0e60ba95

ritikamotwani:
Fixes: #194

### DIFF
--- a/src/General/Icon/Icon.tsx
+++ b/src/General/Icon/Icon.tsx
@@ -9,6 +9,7 @@ const Icon: React.FunctionComponent<Props> = props => {
     width = '1em',
     height = '1em',
     onClick,
+    size = 's',
     ...restProps
   } = props;
 
@@ -20,6 +21,7 @@ const Icon: React.FunctionComponent<Props> = props => {
       height={height}
       onClick={onClick}
       fill={color}
+      fontSize={size}
       viewBox="0 0 100 100"
       {...restProps}
     >
@@ -28,12 +30,23 @@ const Icon: React.FunctionComponent<Props> = props => {
   );
 };
 
+export type iconSizeType =
+  | 'xxs'
+  | 'xs'
+  | 's'
+  | 'm'
+  | 'l'
+  | 'xl'
+  | 'xxl'
+  | 'xxxl';
+
 export interface Props {
   className?: string;
   children: React.ReactNode;
   color?: string;
   width?: string | number;
   height?: string | number;
+  size?: iconSizeType;
   onClick?(e: React.MouseEvent<SVGSVGElement, MouseEvent>): void;
 }
 
--- a/src/General/Icon/Icon.tsx
+++ b/src/General/Icon/Icon.tsx
@@ -9,7 +9,7 @@ const Icon: React.FunctionComponent<Props> = props => {
     width = '1em',
     height = '1em',
     onClick,
-    size = 's',
+    size,
     ...restProps
   } = props;
 
@@ -17,11 +17,10 @@ const Icon: React.FunctionComponent<Props> = props => {
     <VerticalCenteredSvg
       data-testid="icon-svg"
       className={className}
-      width={width}
-      height={height}
+      width={iconSizeMap[size] || width}
+      height={iconSizeMap[size] || height}
       onClick={onClick}
       fill={color}
-      fontSize={size}
       viewBox="0 0 100 100"
       {...restProps}
     >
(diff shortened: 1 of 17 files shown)


src/General/Icon/Icon.tsx
```diff
@@ -9,6 +9,7 @@ const Icon: React.FunctionComponent<Props> = props => {
     width = '1em',
     height = '1em',
     onClick,
+    size = 's',
```
rosslo: Is `size = 's'` backward compatible? Asking because IIRC, our global font-size on DST is 14px and default `width` and `height` is `1em`, probably we need `14px` as default?

src/General/Icon/Icon.tsx
```diff
@@ -20,6 +21,7 @@ const Icon: React.FunctionComponent<Props> = props => {
       height={height}
       onClick={onClick}
       fill={color}
+      fontSize={size}
```
rosslo: Since we have `width` and `height` props for size style why we need to add `fontSize`?

src/General/Icon/IconStyle.ts
```diff
@@ -2,4 +2,44 @@ import styled from 'styled-components';
 
 export const VerticalCenteredSvg = styled.svg`
   vertical-align: middle;
+  ${({ fontSize }) => {
+    switch (fontSize) {
```
rosslo: I think we can just have something like
```
const iconSizeMap = {
  xxs:  '12px',
  xs:  '16px',
 // ...
}
```

```
<VerticalCenteredSvg
   width={iconSizeMap[size] || width}
   height={iconSizeMap[size] || height}
/>
```

stories/General/IconStory.tsx
```diff
@@ -110,6 +155,69 @@ const LikeButton = styled(ThumbsUpOutlineIcon)\`
...
+        <div style={{ flex: '1 1 20%', margin: '1em' }}>
+          <StarIcon size="xxl" />
+          <p style={{ marginTop: '1em', fontSize: '12px' }}>
+            <code style={{ fontSize: '14px' }}>XX-Large (96px)</code>
+          </p>
+        </div>
+        <div style={{ flex: '1 1 20%', margin: '1em' }}>
+          <StarIcon size="xxxl" />
+          <p style={{ marginTop: '1em', fontSize: '12px' }}>
+            <code style={{ fontSize: '14px' }}>XX-Small (12px)</code>
```
rosslo: is it supposed to be `XXX-Large (128px)`?

stories/General/IconStory.tsx
```diff
@@ -30,6 +31,50 @@ const props = {
...
+          style={{
+            marginBottom: '0',
+            marginTop: '1em',
+            fontWeight: 'bold',
+          }}
+        >
+          Size chart:
+        </p>,
+        <p key="xxs" style={{ margin: '0' }}>
+          xxs: 12px
```
rosslo: once we have something like `sizeMap`, we can use it with the rendering loop to simplify the code here

stories/General/IconStory.tsx
```diff
@@ -110,6 +155,69 @@ const LikeButton = styled(ThumbsUpOutlineIcon)\`
...
+      <StarIcon size='xl' />
+      <StarIcon size='xxl' />
+      <StarIcon size='xxxl' />
+      `}
+    >
+      <div style={{ display: 'flex', flexWrap: 'wrap' }}>
+        <div style={{ flex: '1 1 20%', margin: '1em' }}>
+          <StarIcon size="xxs" />
+          <p style={{ marginTop: '1em', fontSize: '12px' }}>
+            <code style={{ fontSize: '14px' }}>XX-Small (12px)</code>
```
rosslo: same as above

src/General/Icon/Icon.tsx
ritikamotwani: Oh had to ask this to the designers about which one to keep as default. Thanks for catching this!!

src/General/Icon/Icon.tsx
ritikamotwani: Oh I didn't notice the width and height. 
I just checked material UI and it was using font sizes for differentiating between sizes so just used that. 
Makes sense if we have the width and height as props already.

ritikamotwani:
Thanks @rosslo ! Updated.

stories/General/IconStory.tsx
```diff
@@ -110,6 +137,32 @@ const LikeButton = styled(ThumbsUpOutlineIcon)\`
...
+      <StarIcon size='xs' />
+      <StarIcon size='s' />
+      <StarIcon size='m' />
+      <StarIcon size='l' />
+      <StarIcon size='xl' />
+      <StarIcon size='xxl' />
+      <StarIcon size='xxxl' />
+      `}
+    >
+      <div style={{ display: 'flex', flexWrap: 'wrap' }}>
```
rosslo: would like to suggest adding `align-items: flex-end;` style here to make Icon examples looks tidy. 

like
image

rosslo (approved):

PR #496, opened 2020-10-09
Drawer: Remove isDisplay, use better css instead

### LINKED ISSUE #479: Drawer: Erratic Behavior When Containing Input Components
I added some advanced input component like a date picker, selects and a range slider to the drawer. Using them causes the open/close animation to play. Not quite sure what's the cause/connection.

Video in this thread: https://glints.slack.com/archives/CS09D0LJX/p1600662340004800

westwood846:
Fixes #479

### DIFF
--- a/src/Navigation/Drawer/Drawer.tsx
+++ b/src/Navigation/Drawer/Drawer.tsx
@@ -9,15 +9,6 @@ const Drawer = ({
   position = 'right',
   ...defaultProps
 }: Props) => {
-  const [firstRenderDone, setFirstRenderDone] = React.useState(false);
-  React.useEffect(
-    function markFirstRenderAsDone() {
-      if (!firstRenderDone) {
-        setFirstRenderDone(true);
-      }
-    },
-    [firstRenderDone]
-  );
   return (
     <DrawerContainer
       className="aries-drawer"
@@ -32,7 +23,6 @@ const Drawer = ({
         role="dialog"
         data-testid="drawer-wrapper"
         open={isOpen}
-        firstRenderDone={firstRenderDone}
         position={position}
         tabIndex={0}
         onClick={e => e.stopPropagation()}
--- a/src/Navigation/Drawer/Drawer.tsx
+++ b/src/Navigation/Drawer/Drawer.tsx
@@ -9,35 +9,13 @@ const Drawer = ({
   position = 'right',
   ...defaultProps
 }: Props) => {
-  const [isDisplay, setIsDisplay] = React.useState<boolean>(false);
-
-  const handleAnimationStart = () => {
-    setIsDisplay(true);
-  };
-
-  const handleAnimationEnd = () => {
-    setIsDisplay(false);
-  };
-
-  React.useEffect(
-    function startAnimationOnOpen() {
-      if (isOpen && !isDisplay) {
-        handleAnimationStart();
-      }
-    },
-    [isDisplay, isOpen]
-  );
-
   return (
     <DrawerContainer
       className="aries-drawer"
       data-testid="drawer-container"
       aria-modal="true"
-      aria-hidden={isDisplay ? 'false' : 'true'}
-      isDisplay={isDisplay}
+      aria-hidden={isOpen ? 'false' : 'true'}
       open={isOpen}
-      onAnimationStart={handleAnimationStart}
-      onAnimationEnd={handleAnimationEnd}
       onClick={() => onClose()}
     >
       <DrawerWrapper
--- a/src/Navigation/Drawer/Drawer.tsx
+++ b/src/Navigation/Drawer/Drawer.tsx
@@ -9,6 +9,15 @@ const Drawer = ({
   position = 'right',
   ...defaultProps
 }: Props) => {
+  const [firstRenderDone, setFirstRenderDone] = React.useState(false);
+  React.useEffect(
+    function markFirstRenderAsDone() {
+      if (!firstRenderDone) {
+        setFirstRenderDone(true);
+      }
+    },
+    [firstRenderDone]
+  );
   return (
     <DrawerContainer
       className="aries-drawer"
@@ -23,6 +32,7 @@ const Drawer = ({
         role="dialog"
         data-testid="drawer-wrapper"
         open={isOpen}
+        firstRenderDone={firstRenderDone}
         position={position}
         tabIndex={0}
         onClick={e => e.stopPropagation()}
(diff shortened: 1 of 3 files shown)


westwood846:
Tested this on both Employers (new feature in E-ATS) and DST (mobile menu).

src/Navigation/Drawer/DrawerStyle.ts
```diff
@@ -44,38 +44,47 @@ const closeToLeft = keyframes`
...
   top: 0px;
   left: 0px;
-  background-color: rgba(0, 0, 0, 0.5);
+  opacity: ${({ open }) => (open ? '1' : '0')};
+  visibility: ${({ open }) => (open ? 'visible' : 'hidden')};
+  background-color: ${({ open }) =>
+    open ? 'rgba(0, 0, 0, 0.5)' : 'transparent'};
   animation: ${({ open }) => (open ? fadeIn : fadeOut)} 0.3s ease-in-out;
-  z-index: ${({ isDisplay }) => (isDisplay ? '9999' : '-1')};
+  z-index: 9999;
```
ritikamotwani, line 66: z-index should be this only when it's opened, right? Should we have the condition?

src/Navigation/Drawer/DrawerStyle.ts
```diff
@@ -44,38 +44,47 @@ const closeToLeft = keyframes`
...
+    visibility: visible;
+    background-color: rgba(0, 0, 0, 0.5);
   }
-`;
+  `;
 
 const fadeOut = keyframes`
   from {
     opacity: 1;
+    visibility: visible;
```
ritikamotwani: Not sure about this, as the visibility is set here, does it mean for the `hidden` case we want the drawer to exist but just hidden? When it's faded out it shouldn't be in the display?

ritikamotwani (approved):
Some minor questions added. Should this be considered a breaking change and maybe we can manually add the release note for it in the frontend channel?

src/Navigation/Drawer/DrawerStyle.ts
westwood846: Somehow it seemed to make a difference when I set it to `-1`, it was glitching out. I think having it at this index at all times is fine.

src/Navigation/Drawer/DrawerStyle.ts
westwood846: Yep, it is just hidden when it is "closed". That's essentially how I "fixed" this, because otherwise you have to do more complicated state management to ensure that display is only set to `none` _after_ the close animation is done or _before_ the open animation runs.

ritikamotwani: Oh weird, shouldn't be just hidden when closed 🤔 .

src/Navigation/Drawer/DrawerStyle.ts
ritikamotwani: Oh even if we set index when it's open and do not set any index value when it's not open?

westwood846: Let me try that...

westwood846:
> Some minor questions added. Should this be considered a breaking change and maybe we can manually add the release note for it in the frontend channel?

I think no need, nothing is breaking ^^

src/Navigation/Drawer/DrawerStyle.ts
westwood846: Yeah not setting it doesn't work either. The problem is, it needs to be at z-index 9999 when the animation is running. Otherwise the overlay could get rendered behind other elements.

ritikamotwani:
@westwood846 , I get  errors while running this, after the build is complete. Could you check? :)

ritikamotwani:
@westwood846 , I think there's some problem with the animation when the Close button is clicked for the drawer. It just disappears. Whereas for [the current version](http://aries.glints.design/?path=/story/navigation--drawer), it gradually disappears from the screen, the fade out is proper.


I don't think the fadeOut is working. I changed the time to 1s, and on clicking close it abruptly closes

src/Navigation/Drawer/DrawerStyle.ts
```diff
@@ -44,39 +44,51 @@ const closeToLeft = keyframes`
...
   min-height: 100vh;
   max-height: 100vh;
   width: 100vw;
   top: 0px;
   left: 0px;
-  background-color: rgba(0, 0, 0, 0.5);
-  animation: ${({ open }) => (open ? fadeIn : fadeOut)} 0.3s ease-in-out;
-  z-index: ${({ isDisplay }) => (isDisplay ? '9999' : '-1')};
+  opacity: ${({ open }) => (open ? '1' : '0')};
+  visibility: ${({ open }) => (open ? 'visible' : 'hidden')};
```
ritikamotwani: Why not use the display property to control this? Opacity won't be required then.

src/Navigation/Drawer/Drawer.tsx
```diff
@@ -9,42 +9,30 @@ const Drawer = ({
...
-  const handleAnimationEnd = () => {
-    setIsDisplay(false);
-  };
-
+  const [firstRenderDone, setFirstRenderDone] = React.useState(false);
   React.useEffect(
-    function startAnimationOnOpen() {
-      if (isOpen && !isDisplay) {
-        handleAnimationStart();
+    function markFirstRenderAsDone() {
```
ritikamotwani: So this flag is only required for the animation? Hmm I guess it's a hacky way and someone might not understand the purpose while going through the component easily?
Maybe we should add a comment with a TO-DO to fix the way?

src/Navigation/Drawer/DrawerStyle.ts
westwood846: `display` can not be animated.

westwood846:
@ritikamotwani so I gave this another try. I think I've got a good solution now. The trick is not to use `keyframes` and to use the `transition` css prop to make sure that the `visibility` prop gets transitioned before/after the other props. The `isDisplay` prop was essentially doing the same thing with `display`.

src/Navigation/Drawer/DrawerStyle.ts
ritikamotwani: Oh it was being used earlier, I thought it should be working following the same styles.

westwood846: The previous implementation did it roughly like this: Changing the `open` prop triggered the animation. And the start/end of that animation then triggered the changing of the `isDisplay` state, which triggered the `display` css prop to change. It was important to get the timing right for the `display` prop because you can't set `display` to `none` while the animation is still running. But with `visibility` combined with `transition` this is possible.

ritikamotwani (approved):
Looks much simpler and readable now.

PR #497, opened 2020-10-12
Core/remove glints poppins from root

franciscrispin:
Remove glints/poppins from src/index
- UI libraries should not load their own fonts for performance reasons
- Developers using the libraries should be responsible for loading their
own fonts
- Reference: https://material-ui.com/components/typography/#general

Moved @glints/poppins to `optionalDependencies`

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,9 +119,7 @@
     "@glints/poppins": "^1.0.0",
     "classnames": "^2.2.6",
     "create-react-context": "^0.2.3",
-    "moment": "^2.24.0",
-    "react-datetime": "^2.16.3",
-    "react-responsive": "^6.1.2"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -116,12 +116,9 @@
     "webpack-glob-entry": "2.1.1"
   },
   "dependencies": {
-    "@glints/poppins": "^1.0.0",
     "classnames": "^2.2.6",
     "create-react-context": "^0.2.3",
-    "moment": "^2.24.0",
-    "react-datetime": "^2.16.3",
```
franciscrispin, line 123: react-datetime and react-responsive seem to be unused

rosslo (approved):

PR #498, opened 2020-10-14
Checkbox: Make checkbox controllable

westwood846:


src/Input/Checkbox/Checkbox.test.tsx
```diff
@@ -49,6 +49,19 @@ it('when toggling checkbox, it should fire onClick once and become checked, then
...
+  const { checkboxInput, checkboxLabel } = setupCheckbox({
+    ...props,
+    checked: false,
+  });
+
+  fireEvent.click(checkboxLabel);
+  expect(props.onClick).toHaveBeenCalledTimes(1);
+  expect(checkboxInput.checked).toEqual(false);
+
+  props.onClick.mockClear();
```
rosslo: I think we can remove this since we do it with `afterEach` in line 26 already.

src/Input/Checkbox/Checkbox.test.tsx
```diff
@@ -69,6 +82,16 @@ describe('when it is rendered', () => {
...
+  it('should display a checked checkbox if checked=true', () => {
+    const { checkboxInput } = setupCheckbox({ ...props, checked: true });
+    expect(checkboxInput.checked).toEqual(true);
+  });
+
+  it('should display an unchecked checkbox if checked=false', () => {
+    const { checkboxInput } = setupCheckbox({ ...props, checked: false });
+    expect(checkboxInput.checked).toEqual(false);
+  });
+
```
rosslo, line 109: I think we can consider checking controlled value with rerender. 

like this
```
describe('<Select/> handleValueChange', () => {
  it('input value should be the same as passed prop value', async () => {
    const Component = ({ value }: { value: string }) => (
      <Select label="handleValueChange" value={value}>
        <Select.Option value="handleValueChange">
          handleValueChange
        </Select.Option>
      </Select>
    );

    const mockValue = 'mock';
    const { queryByRole, rerender } = render(<Component value="" />);
    const selectInput = queryByRole('combobox');
    expect(selectInput.getAttribute('value')).toEqual('');
    rerender(<Component value={mockValue} />);
    expect(selectInput.getAttribute('value')).toEqual(mockValue);
    rerender(<Component value="" />);
    expect(selectInput.getAttribute('value')).toEqual('');
  });
});
```

westwood846: Hmm but value isn't the property that's being controlled..? `checked` is the one that we need to be concerned about. Not quite sure what's the use-case for changing/controlling `value`.

rosslo: Oh sorry for being unclear, yeah I meant `checked` not `value` haha. 
So it might look like this. 
```
    const Component = props => <Checkbox {...props} />;

    const { rerender, getByLabelText } = render(
      <Component {...props} checked={false} />
    );

    const checkboxInput = getByLabelText(
      props.value as string
    ) as HTMLInputElement;
    expect(checkboxInput.checked).toEqual(false);

    rerender(<Component checked={true} />);

    expect(checkboxInput.checked).toEqual(true);

```

westwood846: Oh yeah that makes sense. I've added the test case above.

rosslo (approved):

PR #502, opened 2020-10-21
Add Disabled State for Checkbox

### LINKED ISSUE #495: Checkbox: Add disabled state

westwood846:
Closes #495 

There is a breaking change here: #498 made the checkbox controllable but that causes a warning about controlling in `input` with `checked` without a `onChange` prop. We were using `onClick` instead. In React, for a controlled component, it should be `onChange`. Changing this will also require the consumers of this component to switch to the `onChange` prop.

### DIFF
--- a/src/Input/Checkbox/Checkbox.tsx
+++ b/src/Input/Checkbox/Checkbox.tsx
@@ -8,7 +8,7 @@ const Checkbox: React.FunctionComponent<CheckboxProps> = ({
   id,
   label,
   value,
-  onClick,
+  onChange,
   size = 'small',
   border = false,
   className,
@@ -18,10 +18,10 @@ const Checkbox: React.FunctionComponent<CheckboxProps> = ({
 }: CheckboxProps) => {
   const [internalChecked, setInternalChecked] = React.useState(false);
 
-  const handleClick = (e: React.MouseEvent<HTMLInputElement, MouseEvent>) => {
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     setInternalChecked(internalChecked => !internalChecked);
-    if (onClick !== undefined) {
-      return onClick(e);
+    if (onChange !== undefined) {
+      return onChange(e);
     }
   };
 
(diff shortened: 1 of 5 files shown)


src/Input/Checkbox/Checkbox.tsx
```diff
@@ -36,13 +37,16 @@ const Checkbox: React.FunctionComponent<CheckboxProps> = ({
...
       border={border}
       checked={combinedChecked}
+      disabled={disabled}
     >
       <input
         type="checkbox"
         id={id}
         value={value}
-        onClick={handleClick}
+        // onClick={handleClick}
```
rosslo, line 46: Do you think we can make Checkbox backward compatible for `onClick` event?  `onClick={onClick}`

And show the warning for it like https://github.com/glints-dev/glints-aries/blob/v4/src/General/Button/Button.tsx#L23 to let the developers gradually replace the onClick event.

rosslo:
I think the designers have defined the disabled style haha.

image

https://app.abstract.com/projects/da6f9bdf-2057-45bf-9dda-8dbf0d261eac/branches/e6bfab17-35a1-4e66-ae41-f35d0b58e540/commits/db35c1fadc1892f4fc857e4583e3658f1d2281f9/files/DDF7C217-39F9-45D7-8C53-8D8C88D5FC37/layers/C5041BEC-528E-42F6-84BA-EB540AA09B90?collectionId=96995f6c-2417-4fbd-8448-3fe70962abbe&collectionLayerId=d2b49f68-4908-4770-afcf-89bf7cb226ae

rosslo:
Minor: could we add a gap between the checkboxes 🙏 
image

src/Input/Checkbox/Checkbox.tsx
westwood846: I tried this but the problem is that the `Change` and `Click` events are not compatible. So we'd have to convert them somehow. I think it may be easier to migrate the few places where this is being used instead.

westwood846:
Putting this PR on pause for now, there is a chance that we won't need the disabled state as per the design system. This should be decided in two weeks.

westwood846:
Not needed after all.

PR #503, opened 2020-10-21
Select: Add filterFunction prop

### LINKED ISSUE #444: Select.Options unable to display accented characters
**Description:**

When the search term is unaccented and the fetch results have labels with accented characters, these options are filtered out as the initial search term is unaccented.

**Suggested Solution:**

Allow Select to accept a custom filter function, eg.

`filterOptions(options, searchTerm) {
    ...
}`

**Ref:** 
https://gitlab.glints.com/glints/glints-dst/-/merge_requests/1779

**Side investigation:** 
Select.Option takes children as any[] instead of string 
https://gitlab.glints.com/glints/glints-dst/-/merge_requests/1779#note_52508

westwood846:
Closes #444. Solution is to provide a prop for specifying a custom filter function. For the accented characters problem, such a function can transform the option string and the search string to remove the accent characters as required.

### DIFF
--- a/src/Input/Select/Select.test.tsx
+++ b/src/Input/Select/Select.test.tsx
@@ -30,11 +30,13 @@ const SelectChildren = props.options.map(option => (
   </Select.Option>
 ));
 
-const SelectComponent = (
-  <Select label={props.label} onChange={props.onChange}>
+const SelectComponentWithExtraProps = (extraProps = {}) => (
+  <Select label={props.label} onChange={props.onChange} {...extraProps}>
     {SelectChildren}
   </Select>
 );
+
+const SelectComponent = SelectComponentWithExtraProps();
 
 function setupOpenSelectMenu() {
   const { getByRole, queryAllByTestId, getByTestId } = render(SelectComponent);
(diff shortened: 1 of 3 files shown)


stories/Input/SelectStory.tsx
```diff
@@ -139,6 +139,17 @@ const props = {
       require: 'no',
       description: 'Sets initial open state of Select.',
     },
+    {
```
ritikamotwani, line 142: @westwood846 , could you add a usage for this prop in the examples. 
Also the prop type isn't boolean, right?

src/Input/Select/Select.test.tsx
```diff
@@ -53,6 +55,27 @@ function setupOpenSelectMenu() {
...
 
+function setupWithCustomFilterFunction(
+  filterFunction: (option: string, search: string) => boolean
+) {
+  const { getByRole, queryAllByTestId, getByTestId } = render(
+    SelectComponentWithExtraProps({ filterFunction })
+  );
+  const selectInput = getByRole('combobox') as HTMLInputElement;
+  const selectList = getByTestId('listbox');
+  // const selectLabel = getByTestId('select-label');
```
ritikamotwani: Why is this comment here?

src/Input/Select/Select.tsx
```diff
@@ -116,11 +117,13 @@ const Select: ISelect = (props: Props) => {
     }
 
     // for search results, so we do toLower to filter matched options
```
ritikamotwani: This comment should be changed now?

stories/Input/SelectStory.tsx
westwood846: Added an example and updated the prop here.

src/Input/Select/Select.test.tsx
westwood846: Removed

src/Input/Select/Select.tsx
westwood846: Removed the whole comment, not really helpful anyway

ritikamotwani (approved):
LGTM

PR #506, opened 2020-11-04
Update breakpoints/GlintsContainer

westwood846:
I believe releasing this is safe. I tested the new breakpoints/GlintsContainer on Employers and things look fine. As for DST, they are not using GlintsContainer right now.

https://app.abstract.com/projects/da6f9bdf-2057-45bf-9dda-8dbf0d261eac/branches/61a11574-e24b-4b2d-87d3-61c323ac23a5/collections/0f689c79-8999-432e-a43b-31de99d2fdb9


src/Layout/GlintsContainer/GlintsContainer.ts
```diff
@@ -1,43 +1,52 @@
...
-    padding: 0 20px;
+  @media ${Device.desktopS} {
+    width: 960px;
+    margin: 0 auto;
   }
-  @media ${`(min-width: ${ScreenSize.mobileS}px)`} and (max-width: 479px) {
-    padding: 0 10px;
+  @media ${Device.desktopM} {
+    width: 1140px;
+    margin: 0 auto;
```
ritikamotwani: Do we have to set margin: 0 auto again for desktopM and desktopL? It's the same for desktop sizes can just be added to desktopS?

src/Utils/StyleConfig.ts
```diff
@@ -59,11 +54,11 @@ export const ScreenSize = {
...
 
 export const Device = {
-  mobileS: `(max-width: ${ScreenSize.mobileS}px)`,
-  mobileM: `(max-width: ${ScreenSize.mobileM}px)`,
-  mobileL: `(max-width: ${ScreenSize.mobileL}px)`,
-  tablet: `(max-width: ${ScreenSize.tablet}px)`,
-  desktopS: `(max-width: ${ScreenSize.desktopS}px)`,
-  desktopM: `(max-width: ${ScreenSize.desktopM}px)`,
-  desktopL: `(max-width: ${ScreenSize.desktopL}px)`,
+  mobileS: `(max-width: ${ScreenSize.mobileM - 1}px)`,
```
ritikamotwani, line 57: Device is being used at some places in dst? These changes can affect those too.

src/Layout/GlintsContainer/GlintsContainer.ts
westwood846: Removed

src/Utils/StyleConfig.ts
westwood846: Right. I checked with `yalc` for DST too. Looks good to me, couldn't find any problems. There might be some slight differences but nothing is breaking, so let's merge this PR and deal with the problems as we go.

ritikamotwani: Okay!

ritikamotwani (approved):

src/Utils/StyleConfig.ts
franciscrispin: Haha I don't think we can conclude that nothing breaks by just visually looking around the codebase. Since we swapped `max` with `min`, it should follow that all instances of `Device` used on our codebases would have to be refactored. https://glints.atlassian.net/browse/GM-780

franciscrispin: Oh also could someone help check the rest of the cases where `Device` is used? I suspect there may be other affected components.

westwood846: I'll talk to the designers about the change in the definition.

tingtingdisco: another bug here, https://glints.atlassian.net/browse/GM-789
I am going to update all the Device references on DST

ritikamotwani: @westwood846 , I think we should let the QA team check once after changes for components that have been used a lot in the projects. 🤔

PR #512, opened 2020-11-09
Test: Box & Flex

rosslo:


src/Layout/Flex/Flex.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+  });
+});
+
+describe('<Flex /> flex properties', () => {
+  it('should match snapshot when flexDirection is set', () => {
+    const { asFragment } = render(<Flex flexDirection="column" />);
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('should match snapshot when justifyContent is set', () => {
```
ritikamotwani: @rosslo , here we can test for all options? 
Like have all options in an array and match snapshot for each? Same for flexWrap, alignItems

ritikamotwani (approved):

PR #526, opened 2020-11-24
Integrate Chromatic

rosslo:
Currently, we plan to use chromatic's free plan first which has 5,000 free snapshots/month only, so we need to prevent unnecessary deployment from happening.
Therefore, we only allow the following events to trigger the Chromatic deployment job.
- Push event on `feature/*` branch
- Push event on `v4`, this should be triggered when merging a PR because `v4` is our production base branch.
- Push event on `storybook`, this should also be triggered when merging a PR because it is our feature base branch for storybook v6 migration.

Chromatic Dashboard: https://www.chromatic.com/builds?appId=5fbcb813bd13c900214bb971
Up-to-date storybook for v4 branch: https://v4--5fbcb813bd13c900214bb971.chromatic.com


.github/workflows/cicd.yml
```diff
@@ -0,0 +1,63 @@
+name: CI/CD
+on: push
+jobs:
+  validate:
```
rosslo, line 4: The validate job doesn't have any change, it was copied from `validate.yml`.

.github/workflows/cicd.yml
```diff
@@ -0,0 +1,63 @@
...
+        with:
+          node-version: '10.x'
+      - name: Install dependencies
+        run: yarn install
+      - name: Deploy to Chromatic
+        uses: chromaui/action@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          buildScriptName: storybook:build
```
westwood846, line 63: So the build actually runs on GitHub now? Not on Chroma?

westwood846 (approved):
Looks good to me, can't see any issues with this config ^^ If it works then we should be good.

.github/workflows/cicd.yml
rosslo: Yes, actually it is just running something like `yarn chromatic -t < projectToken> -b storybook:build` on Github cloud.

PR #527, opened 2020-11-25
Modal: Restore previous media queries

westwood846:
This fixes a regression where the modal didn't have the correct width
because the definitions of our device classes changed.


westwood846:
Continued in https://github.com/glints-dev/glints-aries/pull/529

PR #529, opened 2020-11-25
Modal: Restore previous media queries (Recreated PR)

westwood846:
This fixes a regression where the modal didn't have the correct width
because the definitions of our device classes changed.

### DIFF
--- a/src/Display/Gallery/__snapshots__/Gallery.test.tsx.snap
+++ b/src/Display/Gallery/__snapshots__/Gallery.test.tsx.snap
@@ -299,13 +299,13 @@ exports[`<Gallery /> active item click on gallery item 1`] = `
   border: 1px solid #FFFFFF;
 }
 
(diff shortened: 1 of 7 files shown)


tingtingdisco:
@westwood846 I checked that `Device` has been used by ModalStyle, AlertStyle, SearchFilterStyle and GlintsContaner, do we also need to update them?

westwood846:
> @westwood846 I checked that `Device` has been used by ModalStyle, AlertStyle, SearchFilterStyle and GlintsContaner, do we also need to update them?

Hmm you're right, there is one reference in `AlertStyle` and three in `SearchFilterStyle` which need to be updated. Modal is updated now and GlintsContainer has been updated in the original PR.

tingtingdisco:
LGTM thank u @westwood846

tingtingdisco (approved):
LGTM

PR #530, opened 2020-11-30
Set up publish automation with protected default branch

rosslo:


.github/workflows/cicd.yml
```diff
@@ -1,5 +1,10 @@
 name: CI/CD
-on: push
+on:
+  push:
+    tags-ignore:
+      - '*.*'
+    paths-ignore:
+      - package.json
```
rosslo, line 7: To prevent unnecessary jobs from being triggered by pushing bump version commit with admin github token.

Below is the experiment result in my forked repo:
**Before:**
image
**After:**
image

.github/workflows/publish.yml
```diff
@@ -47,7 +48,7 @@ jobs:
       - name: Push changes
         uses: ad-m/github-push-action@master
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+          github_token: ${{ secrets.ADMIN_GITHUB_TOKEN }}
```
rosslo, line 51: Use `glints-aries-publisher@glints.com`'s admin GitHub token to bypass the protected branch setting, so we cannot enable `Include administrators` option.

westwood846 (approved):

PR #531, opened 2020-12-02
Feature/storybook6 alert

Kevintirta:
Purpose: Make Alert Component for storybook 6

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -18,6 +18,7 @@
     "@typescript-eslint/no-var-requires": "off",
     "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/ban-ts-ignore": "off",
     "react/no-find-dom-node": "off",
     "react/prop-types": "off",
     "react-hooks/rules-of-hooks": "error",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -18,7 +18,6 @@
     "@typescript-eslint/no-var-requires": "off",
     "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/no-explicit-any": "off",
-    "@typescript-eslint/ban-ts-ignore": "off",
     "react/no-find-dom-node": "off",
     "react/prop-types": "off",
     "react-hooks/rules-of-hooks": "error",
--- a/src/General/Alert/Alert.stories.tsx
+++ b/src/General/Alert/Alert.stories.tsx
@@ -1,0 +1,39 @@
+import React, { useState } from 'react';
+import { Story, Meta } from '@storybook/react';
+
+import { Alert, Props } from './Alert';
+import { Button } from '../Button';
+import { BaseContainer } from '../../Layout/GlintsContainer/GlintsContainer';
+
+export default {
+  title: 'General/Alert',
+  component: Alert,
+  decorators: [Story => <BaseContainer>{Story()}</BaseContainer>],
+  argTypes: {
+    type: {
+      control: {
+        type: 'select',
+      },
+    },
+  },
+} as Meta;
+
+const Template: Story<Props> = args => {
+  const [isOpen, setOpen] = useState(args.isOpen);
+
+  return (
+    <>
+      <Button onClick={() => setOpen(true)} variant="solid-blue">
+        Show Pop up
+      </Button>
+      <Alert {...args} isOpen={isOpen} onClose={() => setOpen(false)} />
+    </>
+  );
+};
+
+export const Interactive = Template.bind({});
+Interactive.args = {
+  type: 'info',
+  message: 'You have successfully applied to Glints as Software Engineer.',
+  autoClose: 3000,
+};
--- a/src/General/Alert/Alert.stories.tsx
+++ b/src/General/Alert/Alert.stories.tsx
@@ -28,5 +28,4 @@ export const Interactive = Template.bind({});
 Interactive.args = {
   type: 'info',
   message: 'You have successfully applied to Glints as Software Engineer.',
-  autoClose: 3000,
 };
--- a/src/General/Alert/Alert.stories.tsx
+++ b/src/General/Alert/Alert.stories.tsx
@@ -9,14 +9,6 @@ export default {
   title: 'General/Alert',
   component: Alert,
   decorators: [Story => <BaseContainer>{Story()}</BaseContainer>],
-  argTypes: {
-    type: {
-      control: {
-        type: 'select',
-        options: ['success', 'warning', 'danger', 'info'],
-      },
-    },
-  },
 } as Meta;
 
 const Template: Story<Props> = args => {
--- a/src/General/Alert/Alert.stories.tsx
+++ b/src/General/Alert/Alert.stories.tsx
@@ -9,6 +9,13 @@ export default {
   title: 'General/Alert',
   component: Alert,
   decorators: [Story => <BaseContainer>{Story()}</BaseContainer>],
+  argTypes: {
+    className: {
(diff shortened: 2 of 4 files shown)


src/Display/Tooltip/Tooltip.tsx
```diff
@@ -86,15 +86,13 @@ export interface Classes {
...
-export type Position = 'top' | 'right' | 'bottom' | 'left';
-
 export interface Props extends HTMLAttributes<HTMLHeadingElement> {
   /** This is an object with three keys: <code>container</code>,
    * <code>content</code> and <code>message</code>. Use them to attach
    * additional classes to the respective elements. */
   classes?: Classes;
   text: string;
-  position?: Position;
+  position?: string;
```
rosslo: I have used another approach to resolve this type error, can refer to this commit https://github.com/glints-dev/glints-aries/pull/532/commits/2ea2b5fa47fa479420e24a275d059c08a220a9c0. 🙏

src/General/Alert/Alert.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export default {
+  title: 'General/Alert',
+  component: Alert,
+  decorators: [Story => <BaseContainer>{Story()}</BaseContainer>],
+  argTypes: {
+    type: {
+      control: {
+        type: 'select',
+        options: ['success', 'warning', 'danger', 'info'],
```
rosslo: I think we can use intuitive control type (radio button) like 
image

src/General/Alert/Alert.tsx
```diff
@@ -18,11 +18,11 @@ import {
 
 import { PrimaryColor, SecondaryColor } from '../../Utils/Colors';
 
-const Alert = ({
+export const Alert = ({
   isOpen,
-  autoClose,
+  autoClose = 3000,
```
rosslo: Oh why we need to set the default for `autoClose`?

src/General/Alert/Alert.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+      },
+    },
+  },
+} as Meta;
```
rosslo, line 19: Can disable the control func for className
image

src/General/Alert/Alert.tsx
```diff
@@ -130,12 +130,18 @@ const Alert = ({
...
 };
 
-interface Props {
+export interface Props {
+  /** Set Alert's style. */
   type: string;
-  message: React.ReactNode;
+  /** Set message for Alert. */
+  message: string;
+  /** A function to close Alert. */
```
rosslo: I think we should also elaborate on when `onClose` will be triggered. 
Like press which key, or click the close icon, and so on.

src/General/Alert/Alert.tsx
Kevintirta: Hi @rosslo, oh since `autoClose` it's not a mandatory field, if default `autoClose` is not specified then the alert will be open indefinitely (unless user wants to close it), thus is that the wanted behaviour?

src/General/Alert/Alert.test.tsx
```diff
@@ -95,12 +90,6 @@ describe('<Alert/> should render correct icon color', () => {
     const icon = queryAllByTestId('icon-svg')[0];
     expect(icon).toHaveAttribute('fill', SecondaryColor.actionblue);
   });
-
-  test('type is not a predefined type', () => {
```
rosslo, line 99: Oh why deleted these two tests?

src/General/Alert/Alert.tsx
rosslo: Hmm, I think there might be a case is that we want to ensure the users see the message. So we don't auto close alert until the users see the message and close it by themseleves.

src/General/Alert/Alert.tsx
```diff
@@ -137,7 +137,7 @@ export interface Props {
   type: AlertType;
   /** Set message for Alert. */
   message: string;
-  /** A function to close Alert. */
+  /** A function to close Alert. onClose will be triggered if the close icon is clicked or the Alert has ran its autoClose duration */
```
rosslo: Also pressing the Enter or ESC keys, can refer to
https://github.com/glints-dev/glints-aries/blob/v4/src/General/Alert/Alert.tsx#L65

src/General/Alert/Alert.test.tsx
Kevintirta: Hi @rosslo , oh because the Alert type only contains 4 types,
`export type AlertType = 'success' | 'warning' | 'danger' | 'info';`
thus, should we still have the test? since if we still have it it will throw an error

src/General/Alert/Alert.tsx
```diff
@@ -130,12 +130,20 @@ const Alert = ({
...
-  type: string;
-  message: React.ReactNode;
+export type AlertType = 'success' | 'warning' | 'danger' | 'info';
+
+export interface Props {
+  /** Set Alert's style. */
+  type: AlertType;
+  /** Set message for Alert. */
+  message: string;
+  /** A function to close Alert. onClose will be triggered if the close icon is clicked or the Alert has ran its autoClose duration or ESC/Enter is clicked */
```
rosslo: Should we use `press` instead of `click`? `ESC/Enter is pressed`.

src/General/Alert/Alert.test.tsx
rosslo: Hmm, I think those test cases are to test the fallback for the wrong type value in the javascript file which doesn't enable the TypeScript. Can use `//@ts-ignore` to bypass the error.

src/General/Alert/Alert.tsx
```diff
@@ -130,12 +130,20 @@ const Alert = ({
...
 };
 
-interface Props {
-  type: string;
-  message: React.ReactNode;
+export type AlertType = 'success' | 'warning' | 'danger' | 'info';
+
+export interface Props {
+  /** Set Alert's style. */
+  type: AlertType;
```
rosslo: Since you set the default value `info` in line 25, I think this property should be optional now.

src/General/Alert/Alert.tsx
Kevintirta: Make sense got it 🙏

src/General/Alert/Alert.test.tsx
Kevintirta: Hi @rosslo , Okay have updated the eslintrc.json file too to allow //@ts-ignore, ready for rereview

rosslo: Oops, sorry, I just realized, mb `<AlertComponent type={'xxx' as AlertType} />` can resolve this error too.

.eslintrc.json
```diff
@@ -21,7 +21,8 @@
     "react/no-find-dom-node": "off",
     "react/prop-types": "off",
     "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "warn"
+    "react-hooks/exhaustive-deps": "warn",
+    "@typescript-eslint/ban-ts-ignore": "off"
```
rosslo: I think can add it to line 21 to group  @typescript-eslint together.

.eslintrc.json
```diff
@@ -18,6 +18,7 @@
     "@typescript-eslint/no-var-requires": "off",
     "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/ban-ts-ignore": "off",
```
rosslo: Oh if we no need @ts-ignore, then we had better not enable this.

rosslo (approved):

PR #532, opened 2020-12-03
Storybook6: Button & Select

rosslo:


src/General/Button/Button.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { FC, ReactNode } from 'react';
```
westwood846, line 1: Is there a problem with this?

src/Input/Select/Select.stories.tsx
```diff
@@ -0,0 +1,210 @@
...
+  title: 'Input/Select',
+  component: Select,
+  argTypes: {
+    status: {
+      control: null,
+    },
+    error: {
+      control: { type: 'boolean' },
+    },
+  },
```
westwood846, line 17: You can wrap this into the `BaseContainer` to get the Poppins font. Looks better.
```suggestion
  decorators: [Story => <BaseContainer>{Story()}</BaseContainer>],
  },
```

src/Input/Select/Select.stories.tsx
```diff
@@ -0,0 +1,210 @@
...
+  </div>
+);
+
+export const DefaultSelect = Template.bind({});
+DefaultSelect.args = {
+  label: 'Jobs',
+};
+
+export const BasicUsage: Story<SelectProps> = () => (
+  <div style={{ height: '200px' }}>
```
westwood846: We should break the repetition here. You can use `BasicUsage.args` to give the template sets of props. See this as an example: https://github.com/glints-dev/glints-aries/pull/528/files#diff-f78cf8d58a1d0fa429b35b62323a518a8c3f579d316777392fb11bdf0dfff5fc

westwood846 (changes requested):
Thanks for this! You're a hero for dealing with the `Select`. Let's clean up a little and then this is good to go.

src/General/Button/Button.tsx
rosslo: Nope, it is just my preference that using `React.FC` or `React.useEffect` instead of importing them one by one. 😅

src/Input/Select/Select.stories.tsx
```diff
@@ -0,0 +1,140 @@
...
+const Template: Story<SelectProps> = args => (
+  <div style={{ height: '200px' }}>
+    <Select
+      label="Jobs"
+      {...args}
+      // set undefined to prevent these properties are shown in code sample
+      onBlur={undefined}
+      onFocus={undefined}
+      onChange={undefined}
+      onInputChange={undefined}
```
rosslo, line 42: I'm afraid that those properties might distract the user's attention so use this approach to hide them.
Screen Shot 2020-12-03 at 10 29 41 PM

src/Input/Select/Select.stories.tsx
```diff
@@ -0,0 +1,140 @@
...
+  </div>
+);
+
+export const CustomizeFilterFunction: Story<SelectProps> = () => (
+  <div style={{ height: '200px' }}>
+    <Select
+      label="Jobs"
+      filterFunction={(option, search) => {
+        return !option.toLowerCase().includes(search.toLowerCase());
+      }}
```
rosslo, line 135: Didn't use `Template` here because it won't show the custom `filterFunction` content.
Screen Shot 2020-12-03 at 10 39 59 PM

src/Input/Select/Select.stories.tsx
westwood846: I think we can accept that for the time being. Hiding the problem won't make it go away. We can investigate this more in the future or hope that StoryBook pushes an update that deals with this.

westwood846 (approved):
LGTM. I'll leave the decision about the `undefined` callbacks to you.

src/Input/Select/Select.stories.tsx
rosslo: Hmm, I think I will keep undefined there because those functions are not listed in the control table. But yeah, we should investigate this more in the future as well. 🙏

PR #533, opened 2020-12-03
Breakpoints: Go back to max-width based device definitions

westwood846:
The recent changes to `Device` broke too many designs for no reason. See discussion here: https://glints.slack.com/archives/CS09D0LJX/p1606182342022200

This PR effectively reverts those changes and adds a new device class `desktopXL` for viewports wider than 1440px.

Also reverts the changes made on `Modal` in https://github.com/glints-dev/glints-aries/pull/529 as those are not necessary anymore.


ritikamotwani (approved):
@westwood846 , approving the revert. Could you just once check the modals on dst, if nothing breaks.

westwood846:
> @westwood846 , approving the revert. Could you just once check the modals on dst, if nothing breaks.

Checked, looks fine.

PR #535, opened 2020-12-04
Storybook6: Toggle switch and radio button.

ritikamotwani:


src/Input/ToggleSwitch/ToggleSwitch.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+} as Meta;
+
+const Template: Story<ToggleSwitchProps> = args => <ToggleSwitch {...args} />;
+
+export const Interactive = Template.bind({});
+Interactive.args = {
+  defaultActive: false,
+};
+
+export const ToggleSwitchWithIcons = Template.bind({});
```
westwood846: I think better to write a full story here. Like export 

```
const ToggleSwitchWithIcons: Story<ToggleSwitchProps> = () => <ToggleSwitch
  active={...}
  inactive={...}
/>
```

Otherwise the code for the example looks useless.

src/Input/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -5,17 +5,21 @@ import { SecondaryColor } from '../../Utils/Colors';
...
 
 export interface ToggleSwitchProps {
+  /** This will determine whether or not the component renders with an active or inactive state at the start. */
   defaultActive?: boolean;
+  /** The icon components that will be used in both active and inactive states. Both icons have to be defined when opting to use this. */
   iconOptions?: {
     active: React.ReactType;
     inactive: React.ReactType;
   };
+  /** Called when the state of the toggle has been changed. Returns true or false depending on whether the toggle is active or inactive. */
```
westwood846:
```suggestion
  /** Called when the state of the toggle has been changed. Called with true or false depending on whether the toggle became active or inactive. */
```
Actually it's called with the boolean, it doesn't return it.

westwood846 (changes requested):
Please check the errors on GitHub, something's wrong with the types for RadioButton.

src/Input/ToggleSwitch/ToggleSwitch.stories.tsx
ritikamotwani: Oh didn't see the example code. Thanks for taking a look!!

ritikamotwani:
@westwood846 , updated this.

westwood846 (approved):
LGTM!

PR #537, opened 2020-12-04
Feature/storybook6 tag

Kevintirta:
Purpose: make tag component for glints aries

### DIFF
--- a/src/General/Tag/Tag.tsx
+++ b/src/General/Tag/Tag.tsx
@@ -12,8 +12,8 @@ const Tag: React.FunctionComponent<Props> = ({
   action,
   children,
   icon,
-  block,
-  outline,
+  block = false,
+  outline = false,
   onClick,
   ...resetProps
 }) => {
@@ -55,7 +55,9 @@ const Tag: React.FunctionComponent<Props> = ({
 };
 
 interface BasicTagProps {
+  /** Set Tag into block style. */
   block?: boolean;
+  /** Set Tag into outline style. */
   outline?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


src/General/Tag/Tag.tsx
```diff
@@ -55,22 +55,29 @@ const Tag: React.FunctionComponent<Props> = ({
...
 }
 
 interface ButtonTagProps {
   icon?: React.ReactNode;
 }
 
+export type TagActionType = 'add' | 'reset';
 interface ActionTagProps {
-  action?: string;
+  /** Set the Action Tag's type. (action must be chosen as variant props value) */
```
westwood846:
```suggestion
  /** (Only for tags with <code>variant</code>=<code>action</code>). Chooses which kind of action the tag will indicate. */
```

src/General/Tag/Tag.tsx
```diff
@@ -55,22 +55,29 @@ const Tag: React.FunctionComponent<Props> = ({
...
+  /** A function triggered when icon or action tag is clicked */
   onClick?(e: React.MouseEvent<HTMLDivElement, MouseEvent>): void;
 }
 
-interface Props extends BasicTagProps, ButtonTagProps, ActionTagProps {
+export type TagVariantType = 'button' | 'action';
+export interface Props extends BasicTagProps, ButtonTagProps, ActionTagProps {
   children: React.ReactNode;
-  variant?: string;
+  /** Set the Tag's type. */
```
westwood846: I think the prop name is clear enough, better to use this space to explain what the values do:
```suggestion
  /** <code>button</code> tags usually show a single piece of information, while <code>action</code> tags are elements in a list (e.g. a list of selected options) */
```

westwood846 (changes requested):
Hi, only two comments about the prop descriptions.

But more importantly, can you please clean up the commits? First rebase onto the latest `storybook6` which should help you get rid of those weird changes in Tooltip.test.tsx. Then squash some or all of your commits and give them more descriptive messages (without typos). Specifically these commit messages aren't good:

* "Tag: make tag component" -- Can't understand what this means
* "Tag: make ttag stories" -- Has a typo and later there is a commit with almost the same message
* "Tag: change to camelcase" -- Hard to understand what this does, just going by the commit message: What changed to camelcase?

westwood846 (approved):
Much better, thanks!

PR #539, opened 2020-12-11
Tag: make BasicTag clickable

rosslo:
## Why and What Changed
To make `BasicTag` clickable to support the new sort tag in the job filter feature.

I haven't written a test for this change because I plan to let Kevin improve the entire Tag testing later.


ritikamotwani (approved):
Oh I thought for div components the restProps mean properties like onClick?

rosslo:
> Oh I thought for div components the restProps mean properties like onClick?

Because `onClick` was extracted for other Tags. 😅

PR #547, opened 2020-12-15
Slider: Migrate to SB6

westwood846:


src/Display/Slider/Slider.tsx
```diff
@@ -10,15 +10,15 @@ import LeftArrow from './LeftArrow';
 import RightArrow from './RightArrow';
 import SliderItem from './SliderItem';
 
-const Slider = ({
+export const Slider = ({
   afterChange,
   children,
-  autoplay,
+  autoplay = false,
   initialItem,
```
ritikamotwani, line 17: Should there be a default value for initialItem?

src/Display/Slider/Slider.tsx
```diff
@@ -201,11 +201,17 @@ Slider.Item = SliderItem;
 export interface Props {
   children: React.ReactNode;
   className?: string;
+  /** The index of the page that will be shown initally (before the first manual- or auto-scroll). Note that the index starts at 1. */
```
ritikamotwani, line 204: The initialItem can have the default value as 1, then we don't need the note, right?

src/Display/Slider/Slider.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { BaseContainer } from '../../Layout/GlintsContainer/GlintsContainer';
+
+export default {
+  title: 'Display/Slider',
+  component: Slider,
+  decorators: [Story => <BaseContainer>{Story()}</BaseContainer>],
+  argTypes: {
+    className: {
+      table: {
+        disable: true,
```
ritikamotwani, line 14: Oh this is nice. I didn't know this part about controlling the display of the props!

src/Display/Slider/Slider.tsx
westwood846: I've thought about this too, but `initialItem` is so weird and messy, I decided not to give one until that's cleaned up.

src/Display/Slider/Slider.tsx
westwood846: I think still important to note, because every other index that I've ever seen is 0-based, not 1-based.

src/Display/Slider/Slider.stories.tsx
westwood846: I think I got it from Ross ^^

src/Display/Slider/Slider.tsx
ritikamotwani: Oh but we're already adding a note about the default value, so it would be good to have it in the table too? Not a major concern though, the purpose is to tell the devs about the default and the note is solving the purpose.

ritikamotwani (approved):
Just a minor comment. LGTM.

src/Display/Slider/Slider.tsx
westwood846: I think you misunderstand. There is no default value. The default value is `undefined`, which, when I do `initialItem=undefined`, unfortunately still doesn't show up in the docs table. The description for the param says that the index starts at 1, meaning that if the users choose to control this component, they have to use number from 1 to n-1 to choose which page should be shown.

PR #548, opened 2020-12-15
Button: Add isRoundedCorner prop

tingtingdisco:
The new rounded corner ghost button has appeared in 3 places on DST, it's time to add it to Aries.

Since we haven't seen the rounded corner style to be applied to other button variants, not applying it to the variants.

Screenshots:
Screen Shot 2020-12-15 at 11 07 29 AM
Screen Shot 2020-12-15 at 11 07 38 AM

### DIFF
--- a/src/General/Button/Button.tsx
+++ b/src/General/Button/Button.tsx
@@ -149,6 +149,7 @@ export interface Props extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   endIcon?: React.ReactNode;
   theme?: string;
   forwardedRef?: React.RefObject<HTMLButtonElement>;
+  isRoundedCorner?: boolean;
 }
 
 const forwardRef = (props: Props, ref: React.RefObject<HTMLButtonElement>) => (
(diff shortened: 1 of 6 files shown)


ritikamotwani:
@tingtingdisco , could you add this to the storybook too? 

Also it was pointed out, that on just adding the border radius, in the while clicking state the button loses it's rounded corners. Do you want to handle that case as well :P

tingtingdisco:
> @tingtingdisco , could you add this to the storybook too?
> 
> Also it was pointed out, that on just adding the border radius, in the while clicking state the button loses it's rounded corners. Do you want to handle that case as well :P

tingtingdisco:
@ritikamotwani I think we can update storybook later, and I will update the active state.

tingtingdisco:
@ritikamotwani the active state also has the rounded corner.

tingtingdisco:
@ritikamotwani all the buttons expect link button and yellow button are now with rounded corner, and their active state are all good.

ritikamotwani (approved):

PR #552, opened 2020-12-17
Tabs: Migrate to storybook 6

westwood846:


src/Display/Tabs/Tabs.tsx
```diff
@@ -111,12 +111,14 @@ type Tabs = React.FunctionComponent<Props> & {
 
 Tabs.Pane = TabPane;
 
-interface Props {
+export interface Props {
   children: React.ReactNode;
   variant?: Variant;
   alignment?: Alignment;
   theme?: Theme;
```
rosslo, line 113: Could we just copy & paste the previous descriptions for `theme` `alignment` and `variant` to here? 🙏

src/Display/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+    <Tabs.Pane tab="Company">Tab Company</Tabs.Pane>
+  </Tabs>
+);
+
+export const Interactive = Template.bind({});
+
+export const Variant = Template.bind({});
+Variant.args = {
+  variant: 'colored',
+};
```
rosslo, line 44: The current variant example seems to have the wrong style effect. When the tab is hovered, the whole tab will bound.  

https://www.loom.com/share/5e50bec973ff4eb2b9d63c946699441e

src/Display/Tabs/Tabs.tsx
```diff
@@ -111,12 +111,14 @@ type Tabs = React.FunctionComponent<Props> & {
 
 Tabs.Pane = TabPane;
 
-interface Props {
+export interface Props {
   children: React.ReactNode;
   variant?: Variant;
```
rosslo, line 109: I think `variant` only works for `alignment` prop is `horizontal`. So probably we should have something like `{ alignment: 'vertical', variant: never } | {alignment: 'horizontal',  variant: Variant}` to type check it. Btw, haven't checked if it works. 😅

src/Display/Tabs/Tabs.tsx
westwood846: Fixed

src/Display/Tabs/Tabs.stories.tsx
westwood846: Ooh, somehow the default for `theme` wasn't working.

src/Display/Tabs/Tabs.tsx
westwood846: How would I add this type check? Oo

rosslo: I think below would work. 🙏 
```
type Variant = 'underlined' | 'colored';
type Alignment = 'horizontal' | 'vertical';
type Theme = 'blue' | 'black';
type CommonProps = {
  children: React.ReactNode;
  theme?: Theme;
  activeTab?: string | number;
  onTabClick?(tab: React.ReactText | React.ReactNode): void;
  className?: string;
}
type Props =
  CommonProps & ({
      alignment: 'vertical';
    }
  | {
      alignment: 'horizontal';
      variant: Variant;
    });
```

westwood846: This seems to work, but I'm getting an error `Property 'variant' does not exist on type 'PropsWithChildren<Props>'.` because of the default variant.

Let's just keep it the way it is right now, we've got a redesign of the tabs in the pipeline anyway that will remove the variants.

src/Display/Tabs/Tabs.stories.tsx
westwood846: This should be solved.

src/Display/Tabs/Tabs.tsx
```diff
@@ -2,23 +2,18 @@ import * as React from 'react';
...
-const Tabs: Tabs = ({
+export const Tabs: Tabs = ({
   activeTab,
   onTabClick,
   children,
   variant = 'underlined',
   className,
   alignment = 'horizontal',
-  theme = ETabThemeVariant.BLACK,
+  theme = 'black',
```
rosslo, line 16: Oh what is the issue of using enum ETabThemeVariant?

westwood846: It didn't play nice with the args table.

Screenshot from 2021-01-12 10-12-26

rosslo (approved):

src/Display/Tabs/Tabs.tsx
rosslo: Ah I see, I think can try to use `as const` instead of `enum`, but I'm ok with this. Because TS should help us check the default value already.

PR #555, opened 2020-12-21
Design system: Shadows

### LINKED ISSUE #543: Design-Sync Shadows
https://zpl.io/VKrm59E

To be discussed by the team: How to implement this?

rosslo:


src/Utils/Shadows/index.ts
```diff
@@ -0,0 +1,28 @@
...
+    down: ShadowOneDown,
+  },
+  2: {
+    down: ShadowTwoDown,
+  },
+  3: {
+    down: ShadowThreeDown,
+  },
+  4: {
+    down: ShadowFiveDown,
```
westwood846: Is it strange that the index is `4` but the var names used here say `Five`?

src/Layout/Box/Box.test.tsx
```diff
@@ -47,3 +48,15 @@ describe('<Box /> ensure default space theme is overwritten', () => {
...
 });
+
+describe('<Box /> shadow', () => {
+  it('should match snapshot when shadow down is set', () => {
+    const { asFragment } = render(<Box boxShadow={Shadows[1].down} />);
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('should match snapshot when shadow up is set', () => {
+    const { asFragment } = render(<Box boxShadow={Shadows[4].up} />);
```
westwood846: I'm not sure that I like this interface :thinking: It implies to me that there should be a `Shadow[1].up` too.

westwood846 (commented):
Overall this is good! But I don't agree with the interface, though I admit that this is quite hard. How about this?

`Shadow.sh1`, `Shadow.sh2`, `Shadow.sh3`, `Shadow.sh4`, `Shadow.neg`? To me this implies more that there are four levels of "regular" shadows and then one "irregular" shadow.

What do you think? We can discuss this together in the SPM too?

src/Utils/Shadows/index.ts
rosslo: Oops typo. 😅

src/Layout/Box/Box.test.tsx
rosslo: Ah, good point. 🤔

rosslo: Btw, the `up` and `down` took `ant-design` as reference.

image

https://ant.design/docs/spec/shadow-cn

src/Utils/Shadows/index.ts
rosslo: Update the interface to 
```
Shadow.up1
Shadow.down1
Shadow.up2
Shadow.down2
// ...
```
I think using direction is more scaleable if we have left or right in the future.
And even the designers have defined one up shadow only `up4`, I think we can still add `up1`, `up2` and `up3` based on the same rule by ourselves first in case it confuses the other devs as this situation https://github.com/glints-dev/glints-aries/pull/555#discussion_r546531565.

westwood846 (approved):
Ok I think this works for now, thanks!

PR #562, opened 2021-01-02
Breadcrumb and Swipeable documentations

ritikamotwani:


src/Display/Swipeable/Swipeable.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <Swipeable {...args}>
+        {Array.from(Array(10).keys()).map(number => (
+          <Swipeable.Item key={number}>
+            <SwipeableCardExample>
+              <h1>Card. {number}</h1>
+            </SwipeableCardExample>
+          </Swipeable.Item>
+        ))}
+      </Swipeable>
+    </>
```
westwood846: What's the Fragment for?

westwood846 (approved):
Tentatively approved, only one comment, you may have a good reason there or maybe it's an oversight.

src/Display/Swipeable/Swipeable.stories.tsx
ritikamotwani: It's 2 have 10 cards as Swipeable Item. 
Didn't understand the comment I think 🤔

westwood846: I mean the `<>...</>`. Usually that's only used because you need to return a single element in React. But `<Swipeable>...</Swipeable>` could be that element here.

ritikamotwani: Ooh I didn't notice them after the comment too, I was like there's no Fragment so why the comment :P . Yes, you're right. It's not required.

ritikamotwani:
@westwood846 , updated.

westwood846 (approved):

PR #566, opened 2021-01-07
TextArea: make test case for Textarea

Kevintirta:
Purpose: make unit test case for Textarea.tsx component of glints aries

### DIFF
--- a/src/Input/Textarea/Textarea.test.tsx
+++ b/src/Input/Textarea/Textarea.test.tsx
@@ -128,10 +128,9 @@ describe('<Textarea /> forwards ref to underlying textarea element', () => {
     expect(textarea).toEqual(ref.current);
   });
 
-  test('ref is being forwarded correctly with values, textareaLabel style will follow when floating is true', () => {
-    const ref = React.createRef<HTMLTextAreaElement>();
+  test('when value is passed, textareaLabel style will follow when floating is true', () => {
     const { getByTestId } = render(
-      <Textarea ref={ref} label={label} value={props.value} />
+      <Textarea label={label} value={props.value} />
     );
     const textareaLabel = getByTestId('textarea-label');
 
--- a/src/Input/Textarea/Textarea.test.tsx
+++ b/src/Input/Textarea/Textarea.test.tsx
@@ -10,11 +10,13 @@ import { SecondaryColor, PrimaryColor } from '../../Utils/Colors';
 
 interface Props {
   label: string;
+  value: string;
   onChange: jest.Mock;
 }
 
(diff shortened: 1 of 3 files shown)


src/Input/Textarea/Textarea.test.tsx
```diff
@@ -121,14 +123,19 @@ describe('<Textarea /> forwards ref to underlying textarea element', () => {
...
     const textarea = queryByLabelText(label);
     expect(textarea).toEqual(ref.current);
   });
 
   test('underlying textarea can be focused/blurred through ref', () => {
     const ref = React.createRef<HTMLTextAreaElement>();
-    const { queryByLabelText } = render(<Textarea ref={ref} label={label} />);
+    const fn = jest.fn();
+    const { queryByLabelText } = render(
+      <Textarea ref={ref} label={label} onBlur={fn} />
```
rosslo: For `onBlur` we should do something like `expect(onBlur).toHaveBeenCalledTimes(1);`

src/Input/Textarea/Textarea.test.tsx
```diff
@@ -121,14 +123,19 @@ describe('<Textarea /> forwards ref to underlying textarea element', () => {
 
   test('ref is being forwarded correctly', () => {
     const ref = React.createRef<HTMLTextAreaElement>();
-    const { queryByLabelText } = render(<Textarea ref={ref} label={label} />);
+    const { queryByLabelText } = render(
+      <Textarea ref={ref} label={label} value={props.value} />
```
rosslo: Oh why we need to add `value` prop here? I think we should only add related props for unit test.

So you can add another test case for `value` prop. Can refer to https://github.com/glints-dev/glints-aries/pull/568/files#diff-dfe5d21ab8f9599028ecdf7e998d103507dd877e2ecffb0a3e403c598ea68bddR52

Kevintirta: Ohh its to cover the lines below that I highlighted and it's only will be triggered if we forward a ref and value to the Textarea comp.

Code in Textarea.tsx

```
...
  const innerRef = React.useRef(null);
  const textareaInputRef = props.forwardedRef || innerRef;
...
  React.useLayoutEffect(() => {
    const textarea = textareaInputRef.current;

    if (!textarea) return;

    if (textarea.value.length > 0) { // this line of code
      setFloating(true);
    }

    setTextareaMaxHeight(~~(textarea.offsetHeight * 3.7 + 23 * 8 + 1));
  }, [textareaInputRef]);

```

rosslo (approved):

PR #568, opened 2021-01-08
Feature/controlled toggle

rosslo:
Allow `ToggleSwitch` to be a controlled component with `active` prop.


src/Input/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -15,32 +16,40 @@ export interface ToggleSwitchProps {
...
-  const [active, setActive] = React.useState(defaultActive);
-  const iconColor = active ? SecondaryColor.actionblue : SecondaryColor.grey;
+  const [innerActive, setInnerActive] = React.useState(defaultActive);
+  const iconColor = innerActive
+    ? SecondaryColor.actionblue
+    : SecondaryColor.grey;
   const ActiveIcon = iconOptions ? iconOptions.active : null;
   const InactiveIcon = iconOptions ? iconOptions.inactive : null;
-  const Icon = active ? ActiveIcon : InactiveIcon;
+  const Icon = innerActive ? ActiveIcon : InactiveIcon;
```
franciscrispin, line 33: Should this use `isActive` instead for consistency?
`const Icon = isActive ? ActiveIcon : InactiveIcon;`

src/Input/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -15,32 +16,40 @@ export interface ToggleSwitchProps {
...
+  active,
   defaultActive,
   iconOptions,
   onChange,
   className,
   ...defaultProps
 }: ToggleSwitchProps) => {
-  const [active, setActive] = React.useState(defaultActive);
-  const iconColor = active ? SecondaryColor.actionblue : SecondaryColor.grey;
+  const [innerActive, setInnerActive] = React.useState(defaultActive);
```
franciscrispin: Wonder if we can make innerActive the single source of truth for easier readability? 
e.g. `const [innerActive, setInnerActive] = React.useState(defaultActive || active);`

src/Input/ToggleSwitch/ToggleSwitch.test.tsx
```diff
@@ -88,6 +109,43 @@ describe('<ToggleSwitch /> click behavior', () => {
     expect(toggleSwitchContainer).toHaveAttribute('aria-checked', 'true');
   });
 
+  it('should change aria-checked correctly when it is controlled', () => {
+    let active = false;
+    const onChangeSpy = jest.fn(isActive => {
+      active = isActive;
+    });
+
+    console.log('active', active);
```
franciscrispin: Leftover console log haha

src/Input/ToggleSwitch/ToggleSwitch.test.tsx
```diff
@@ -49,6 +49,27 @@ describe('<ToggleSwitch /> prop iconOptions', () => {
...
+        <ToggleSwitch
+          active={active}
+          iconOptions={{
+            active: EyeIcon,
+            inactive: EyeSlashedIcon,
+          }}
+        />
+      );
+      // validate defaultActive from snapshot
+      expect(asFragment()).toMatchSnapshot();
```
franciscrispin, line 65: Haha I just wanted to bring up this morning that maybe we should stop relying too much on snapshot tests? They are quite brittle and I find myself often just running `yarn test -u` whenever they fail so i'm not sure how much protection they can provide as well. https://www.sitepen.com/blog/snapshot-testing-benefits-and-drawbacks

In this case the test case you added below should provide enough coverage

src/Input/ToggleSwitch/ToggleSwitch.test.tsx
```diff
@@ -88,6 +109,43 @@ describe('<ToggleSwitch /> click behavior', () => {
...
+    const onChangeSpy = jest.fn(isActive => {
+      active = isActive;
+    });
+
+    console.log('active', active);
+
+    const { getByRole, rerender } = render(
+      <ToggleSwitch
+        active={active}
+        iconOptions={{
```
franciscrispin, line 120: Maybe can omit iconOptions for this test to keep it simple?

src/Input/ToggleSwitch/ToggleSwitch.test.tsx
```diff
@@ -88,6 +109,43 @@ describe('<ToggleSwitch /> click behavior', () => {
...
+        onChange={onChangeSpy}
+      />
+    );
+    const toggleSwitchContainer = getByRole('switch');
+    expect(toggleSwitchContainer).toHaveAttribute('aria-checked', 'false');
+
+    userEvent.click(toggleSwitchContainer);
+    expect(onChangeSpy).toHaveBeenCalledTimes(1);
+
+    rerender(
```
franciscrispin: Oh why do we have to rerender it? I assume after onChangeSpy is called the aria-checked will be updated?

src/Input/ToggleSwitch/ToggleSwitch.test.tsx
```diff
@@ -88,6 +109,43 @@ describe('<ToggleSwitch /> click behavior', () => {
...
+        active={active}
+        iconOptions={{
+          active: EyeIcon,
+          inactive: EyeSlashedIcon,
+        }}
+        onChange={onChangeSpy}
+      />
+    );
+    const toggleSwitchContainer = getByRole('switch');
+    expect(toggleSwitchContainer).toHaveAttribute('aria-checked', 'false');
```
franciscrispin, line 128: Should we also check the `active` prop has the correct value?

franciscrispin (changes requested):

src/Input/ToggleSwitch/ToggleSwitch.test.tsx
rosslo: Because ToggleSwitch is controlled, it should only be updated when `active` prop is changed.

src/Input/ToggleSwitch/ToggleSwitch.test.tsx
rosslo: I agreed snapshot is unreadable, but it can help us catch the bug as you pointed above https://github.com/glints-dev/glints-aries/pull/568#discussion_r553725361 in the future. So I will leave it for now. 🙏

franciscrispin (approved):

PR #570, opened 2021-01-12
Gallery: Migrate stories

westwood846:


src/Display/Gallery/Gallery.tsx
```diff
@@ -13,12 +13,10 @@ import {
   GalleryThumbnailWrapper,
 } from './GalleryStyle';
 
-const defaultImagesDisplayed = 8;
-
 const Gallery = ({
   initialVisibility,
```
ritikamotwani: The default value should be false.

westwood846: Actually I think it's important that it is `undefined`. Anything else would mean that the component controlled.

ritikamotwani: I still think it should be false. Because for the default state the image doesn't open when the Gallery component is rendered. That means the default value of the prop is false. 
Don't understand what difference showing undefined there makes.

westwood846: Hmm yeah now that I think about it again, it doesn't make too much sense not to specify `false` here. Thanks!

ritikamotwani (approved):

PR #578, opened 2021-01-18
Feature/box story migration

rosslo:


src/Layout/Box/BoxStoryStyle.tsx
```diff
@@ -0,0 +1,16 @@
+import styled from 'styled-components';
+
+import { Box as BaseBox } from '../../../src/Layout/Box';
+
+export const GreyBackground = styled.div`
+  display: inline-block;
```
ritikamotwani, line 6: @rosslo , as the styles are all like one property added, how about we add the `style` prop inside the div tag and include it in the Stories JSX.
Else in the documentation's source code, it will be something like <S.Content> etc which might make less sense?

src/Layout/Flex/index.tsx
```diff
@@ -12,14 +12,24 @@ type Props = FlexProps &
...
+  flex-wrap: ${({ flexWrap }) => flexWrap};
 `;
 
-export const Flex: React.FC<Props> = props => <StyledFlex {...props} />;
+export const Flex: React.FC<Props> = ({
+  flexDirection = 'row',
+  justifyContent = 'flex-start',
+  alignItems = 'stretch',
+  flexWrap = 'nowrap',
+  ...restProps
```
ritikamotwani, line 27: Should there be a description of what restProps is at the top and to what component they are passed?

src/Layout/Box/BoxStoryStyle.tsx
rosslo: Oh I thought specific JSX component name will be easier to understand without reading the inline-style. Wdyt?
image

src/Layout/Flex/index.tsx
rosslo: The rest of the props have been listed in the control table of Flex story as well. But I also add the comment in the codebase. 🙏

ritikamotwani (approved):

PR #580, opened 2021-01-19
Select: drop button should not be rendered when select is disabled

tingtingdisco:
https://user-images.githubusercontent.com/11376798/105022504-d1faf580-5a84-11eb-9650-f1447a81ff8c.mov

### DIFF
--- a/src/Input/Select/Select.tsx
+++ b/src/Input/Select/Select.tsx
@@ -375,11 +375,11 @@ const Select: ISelect = (props: Props) => {
             {label}
           </SelectLabel>
         )}
-        {!removeDropIcon && !disabled && (
+        {!removeDropIcon && (
           <div
             className="select-icon"
             aria-label="show options"
-            onClick={handleDropIconClick}
+            onClick={disabled ? null : handleDropIconClick}
           >
             <ArrowDownIcon color="#777777" />
           </div>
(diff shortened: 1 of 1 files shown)


rosslo:
Just checked the other UI libraries, they don't hide the drop button when select is disabled.

- Material-UI: https://material-ui.com/components/selects/#select
image
- Ant Design: https://ant.design/components/select/
image


Do you think we can do `onClick={disabled ? undefined : handleDropIconClick}` instead?

rosslo:
Btw, do you think we should remove the hover state for disabled select together? 🙏

tingtingdisco:
> Btw, do you think we should remove the hover state for disabled select together? 🙏
hmm, we should ask designers.

tingtingdisco:
but I think we can follow other libraries first.

tingtingdisco:
@rosslo updated, thank u

rosslo (approved):

rosslo:
@tingtingdisco , LGTM, thanks for the fix.

tingtingdisco:
@rosslo what's the UI Tests task?

rosslo:
@tingtingdisco , the snapshot check for storybook, the reviewer should also review it and approve the change on chromatic. I thought I have approved it. 😅 

Btw, the change is the glints-aries version in the homepage.
image

PR #586, opened 2021-01-28
Button: fix forward ref

### LINKED ISSUE #575: forwarding ref of Button is now working
ref that is being passed to Button aries component is not passed properly

rosslo:


src/General/Button/Button.tsx
```diff
@@ -65,12 +66,15 @@ const Button: React.FunctionComponent<Props> = ({
     </>
   );
 
+  console.log('forwardedRef', forwardedRef);
```
westwood846: Leftover ;-)

westwood846 (approved):
Look good, just one leftover comment!

src/General/Button/Button.tsx
rosslo: Oops, thanks!

PR #587, opened 2021-01-28
Feature/typography sb6

westwood846:


rosslo:
I think we can add the font-size near the variant radio button. like `caption (14px)`. 
Because the developers might only know the font-size from zeplin or abstract without the variant name.
Therefore, I think we can also make `Title` `tag`'s options be radio button instead of select.


image

westwood846:
> I think we can add the font-size near the variant radio button. like `caption (14px)`.
> Because the developers might only know the font-size from zeplin or abstract without the variant name.
> Therefore, I think we can also make `Title` `tag`'s options be radio button instead of select.

Changed. I don't know of any good option for showing font size in the table. What do you think of this solution?

Screenshot from 2021-02-01 09-09-21

rosslo (approved):

PR #589, opened 2021-02-01
Migrate documentation for the newly added props.

ritikamotwani:


src/Display/Slider/Slider.stories.tsx
```diff
@@ -67,3 +80,59 @@ RemoveDots.args = { removeDots: true };
 
 export const InitialItem = Template.bind({});
 InitialItem.args = { initialItem: 2 };
+
+export const CustomizedRenderLeftIcon: Story<Props> = () => (
```
westwood846, line 84: Can't you do
```
export const InitialItem = Template.bind({});
CustomizedRenderLeftIcon.args = { 
    renderLeftIcon={(disabled: boolean) => {
      const color = disabled ? Greyscale.grey : SecondaryColor.actionblue;
      return <ArrowBackDoubleIcon color={color} />;
    }}
};
```
here?

src/Display/Slider/Slider.tsx
```diff
@@ -217,7 +217,9 @@ export interface Props {
   arrowWhite?: boolean;
   /** If <code>true</code>, dot controls will be removed */
   removeDots?: boolean;
+  /** Sets the left icon. Receives the "disabled" value as an argument. */
```
westwood846: >value
Might be slightly more helpful to call it "disabled flag"

westwood846 (commented):
Left a question and a comment.

src/Display/Slider/Slider.stories.tsx
ritikamotwani: I had done this initially but this function wasn't visible in the source code 
The source code becomes like
<Slider renderLeftIcon={() => {}}>
..........
</Slider>

westwood846: Ooh I see, then this is okay.

westwood846 (approved):

PR #592, opened 2021-02-02
Modal: fix background scrolling and onClose issues

### LINKED ISSUE #581: Modal should check that onClose prop is a function before calling it
Passing in any value that is not a function to `onClose` will cause an error `TypeError: r is not a function`.

Note: If onClose is not a function, I think the modal should still close as expected so that this fix remains backwards compatible.

Related thread: https://glints.slack.com/archives/CQMM9C2N7/p1611129205010700

### LINKED ISSUE #591: Modal should prevent background scrolling for iOS
https://drive.google.com/file/d/1K-wj6qiT4viVMx2J7BY_VlNORFb6TGZo/view

rosslo:
Issues:
- https://github.com/glints-dev/glints-aries/issues/581
- https://github.com/glints-dev/glints-aries/issues/591

### DIFF
--- a/src/Display/Modal/Modal.test.tsx
+++ b/src/Display/Modal/Modal.test.tsx
@@ -25,14 +25,8 @@ const OpenedModal = (
   </Modal>
 );
 
-const ModalComponent = ({ isVisible }: { isVisible: boolean }) => (
-  <Modal isVisible={isVisible} onClose={props.onClose}>
-    <p>{props.content}</p>
-  </Modal>
-);
-
 type TestProps = Omit<Props, 'onClose'> & {
-  onClose?: any;
+  onClose?: () => void;
 };
 
 function setupModal(isVisible: boolean, customProps: TestProps = {}) {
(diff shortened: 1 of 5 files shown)


src/Display/Modal/Modal.test.tsx
```diff
@@ -31,14 +31,19 @@ const ModalComponent = ({ isVisible }: { isVisible: boolean }) => (
   </Modal>
 );
 
-function setupModal(isVisible: boolean) {
+type TestProps = Omit<Props, 'onClose'> & {
+  onClose?: any;
```
ritikamotwani: The type has to be any?

ritikamotwani:
There's a warning in the modal tests

src/Display/Modal/Modal.test.tsx
rosslo: Oh just realized it could be `onClose?: () => void`.

Previously, it is declared as `onClose()?: void` which will make TS complain the line of `onClose: null` in the test file.

ritikamotwani (approved):

PR #598, opened 2021-02-05
Modal: add keepChildrenMountedOnClose prop

### LINKED ISSUE #597: Modal: add keepChildrenMountedOnClose prop
Currently, Modal always unmounts the children on close. 
However, for the new job filter feature, we need the children filter component to be mounted before the modal is opened to compute some initial states.  

`keepChildrenMountedOnClose ` takes `material-ui`'s `keepMounted` as reference https://material-ui.com/api/modal/

rosslo:
Please refer to https://github.com/glints-dev/glints-aries/issues/597 for more detail

### DIFF
--- a/src/Display/Modal/Modal.tsx
+++ b/src/Display/Modal/Modal.tsx
@@ -41,6 +41,7 @@ export const Modal: FC<Props> = ({
   footer,
   size,
   hideHeader,
+  destroyOnClose,
   ...restProps
 }) => {
   const modalContentAreaRef = useRef(null);
@@ -128,6 +129,8 @@ export const Modal: FC<Props> = ({
     [mouseDownTarget, handleClose]
   );
 
+  const shouldMountChildren = isVisible || (!isVisible && !destroyOnClose);
+
   return (
     <ModalContainer
       data-testid="modal-container"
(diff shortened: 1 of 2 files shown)


src/Display/Modal/Modal.tsx
```diff
@@ -192,16 +195,26 @@ export const Modal: FC<Props> = ({
...
+  /** Removes animation when opening and closing the modal. */
+  removeAnimation?: boolean;
+  /** Sets size of Modal. */
   size?: sizeType;
+  /** Hides header area. */
   hideHeader?: boolean;
+  /** Hides content area. */
+  hideContentArea?: boolean;
+  /** Whether to unmount child components on onClose */
+  destroyOnClose?: boolean;
```
westwood846: I'm wondering if we should rather name this `destoryChildrenOnClose` because the rest of the modal does not really get unmounted on close.

westwood846 (approved):
Good work! One comment, I'll leave it to you if you wanna change it or not.

src/Display/Modal/Modal.tsx
rosslo: Yep, specify `children` in the name is better. And I just felt the default boolean value should be `false`, so I rename it to `keepChildrenMountedOnClose` which taking `material-ui`'s `keepMounted` as reference https://material-ui.com/api/modal/

PR #603, opened 2021-02-15
Feature/checkbox fontsize

### LINKED ISSUE #590: Checkbox Changes Size With Font Size but Tick Doesn't Stay Centered
The size of the box changes with the inherited font size. The current implementation assumes a font-size of 14px. For any other, the tick icon won't be centered correctly.

* Clarify with the designers whether the checkbox should grow with the inherited font size. Force a certain size if necessary.
* Whatever the size of the box, make sure the checkmark icon has the right size and is centered.

Screenshot from 2021-02-01 15-41-44

westwood846:
Fixes #590.

Main change here is to use `em` instead of `px` in various places of the Checkbox styles to ensure that they scale with the `font-size`. Also, I made it so that the `size` prop now just determines the Checkbox's `font-size` and removed all other references to remove unnecessary complexity.

I wasn't sure how to write tests for this so I added an example to the storybook.

Plus some cleanup.

### DIFF
--- a/src/Input/Checkbox/CheckboxStyle.ts
+++ b/src/Input/Checkbox/CheckboxStyle.ts
@@ -2,6 +2,26 @@ import styled, { css } from 'styled-components';
 import { Greyscale, SecondaryColor } from '../../Utils/Colors';
 import { CheckboxProps } from './Checkbox';
 
+const borderedTickStyles = css`
+  margin-top: 0.61em;
+  margin-left: 0.89em;
+`;
+
+const borderedLabelStyles = css`
+  border: 1px solid #aaaaaa;
+  cursor: pointer;
+  border-radius: 8px;
+  padding: 0.57em 0.86em;
+  &:hover {
+    background: rgba(1, 126, 183, 0.1);
+    border-color: ${SecondaryColor.actionblue};
+  }
+`;
+
+const checkedLabelStyles = css`
+  border-color: ${SecondaryColor.actionblue};
+`;
+
 export const CheckboxContainer = styled.div<CheckboxProps>`
   position: relative;
   display: inline-flex;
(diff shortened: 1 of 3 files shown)


src/Input/Checkbox/CheckboxStyle.ts
```diff
@@ -24,30 +44,20 @@ export const CheckboxContainer = styled.div<CheckboxProps>`
       content: '';
       display: block;
       position: absolute;
-      top: ${({ size }) => (size === 'small' ? '1.8px' : '3.3px')};
-      left: ${({ size }) => (size === 'small' ? '6.3px' : '7.5px')};
-      width: 6px;
-      height: 12px;
+      top: 0.13em;
```
ritikamotwani, line 47: Hmm so this part is not affected by the size of the checkbox?

ritikamotwani:
@westwood846 , the code changes LGTM!
Is it possible to get it QC'd by the designers too?

src/Input/Checkbox/CheckboxStyle.ts
westwood846: What do you mean?

This part makes the checkbox's dimensions grow with the `font-size`:
```
width: 0.43em;
height: 0.86em;
```

This part positions the tick icon, relative to the top-left corner (which of course is not moving). It's a bit silly to do it this way imo, it would be better to use some alignment styles, but I can't change it to that easily:
```
top: 0.13em;
left: 0.45em;
```

ritikamotwani: Ah no, I mean the `size` prop that can be small.
`top: ${({ size }) => (size === 'small' ? '1.8px' : '3.3px')};`

This part is removed, so what will happen to the Checkboxes with the size prop set as small?

westwood846: There's a line ` font-size: ${({ size }) => (size === 'small' ? '14px' : '16px')};` further up. The `size` determines the `font-size` which in turn determines everything else (that uses `em`).

ritikamotwani: Oh cool had missed that line about the font-size.

ritikamotwani (approved):

PR #608, opened 2021-02-17
Modal: fix scroll issue

### LINKED ISSUE #602: Modals are not scrollable in the projects
For iOS devices the modals are not scrollable.
Bug report: https://glints.slack.com/archives/CBM88R7U1/p1613014751124600

rosslo:
The first commit is to fix this issue https://github.com/glints-dev/glints-aries/issues/602
And the second commit is to fix the footer UI below when the `centering` prop is `true`.

Screen Shot 2021-02-17 at 5 31 12 PM

### DIFF
--- a/src/Display/Modal/ModalStyle.ts
+++ b/src/Display/Modal/ModalStyle.ts
@@ -81,8 +81,9 @@ export const ModalContentArea = styled.div<InnerProps>`
   ${({ centering }) => {
     if (centering) {
       return `
+        display: flex;
+        flex-direction: column;
         max-height: 85vh;
-        overflow: auto;
       `;
     }
   }}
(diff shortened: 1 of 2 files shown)


src/Display/Modal/ModalStyle.ts
```diff
@@ -81,8 +81,9 @@ export const ModalContentArea = styled.div<InnerProps>`
   ${({ centering }) => {
     if (centering) {
       return `
+        display: flex;
+        flex-direction: column;
         max-height: 85vh;
-        overflow: auto;
```
rosslo, line 85: When the `centering` is `true`, only the modal body could be scrollable.

ritikamotwani (approved):
LGTM! Thanks!

PR #612, opened 2021-02-26
Tag: improve test coverage

rosslo:
Before:
image

After:
image


westwood846 (approved):
Nice work! I haven't ever seen tests structured like this, but it seems like a good idea to me, and I'm glad that we're finally paying attention to the semantics of `describe`/`it` :grin:

PR #613, opened 2021-02-26
Dropdown: improve test coverage

rosslo:
Before:
image

After: 
image

The lines 71 and 141 are related to the `itemElement` prop, I suspect it is not used anymore, so create an issue https://github.com/glints-dev/glints-aries/issues/614 to check it in the next sprint.

### DIFF
--- a/src/Navigation/Dropdown/Dropdown.test.tsx
+++ b/src/Navigation/Dropdown/Dropdown.test.tsx
@@ -314,6 +314,61 @@ describe('<Dropdown/> keydown event', () => {
     ).toBeVisible();
   });
 
+  test('when the first option is highlighted, pressing up arrow key should not affect anything', async () => {
+    render(
+      <Dropdown label="Career">
+        <DropdownItem value="pm">Product Manager</DropdownItem>
+        <DropdownItem value="se">Software Engineer</DropdownItem>
+      </Dropdown>
+    );
+
+    const dropdown = queryByRole(document.body, 'menuitem');
+    fireEvent.click(dropdown);
+
+    const optionPM = queryByText(document.body, 'Product Manager');
+    const optionSE = queryByText(document.body, 'Software Engineer');
+    expect(optionPM).toHaveStyle(`background: ${Greyscale.softgrey}`);
+    expect(optionSE).not.toHaveStyle(`background: ${Greyscale.softgrey}`);
+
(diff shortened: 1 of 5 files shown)


src/Navigation/Dropdown/Dropdown.test.tsx
```diff
@@ -292,6 +314,61 @@ describe('<Dropdown/> keydown event', () => {
...
+        <DropdownItem value="se">Software Engineer</DropdownItem>
+      </Dropdown>
+    );
+
+    const dropdown = queryByRole(document.body, 'menuitem');
+    fireEvent.click(dropdown);
+
+    const optionPM = queryByText(document.body, 'Product Manager');
+    const optionSE = queryByText(document.body, 'Software Engineer');
+    expect(optionPM).toHaveStyle(`background: ${Greyscale.softgrey}`);
```
westwood846: I guess there's no way around it without changing the implementation, but I don't think it's good that we're testing styles instead of testing logic here. If we change the styles of active elements, these tests will break.

src/Navigation/Dropdown/Dropdown.test.tsx
```diff
@@ -346,3 +423,86 @@ describe('<Dropdown/> logic', () => {
...
+  test('invalid onClick value should not break anything', async () => {
+    const onClick = 'invalid-on-click';
+    const menuText = 'Product Manager';
+    render(
+      <Dropdown label="Career">
+        <DropdownItem
+          value="pm"
+          // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+          // @ts-ignore
+          onClick={onClick}
```
westwood846, line 456: What was the problem here?

westwood846 (approved):

src/Navigation/Dropdown/Dropdown.test.tsx
rosslo: Update to checking class name instead. 🙏

src/Navigation/Dropdown/Dropdown.test.tsx
rosslo: This test case is to test the invalid onClick (non-function) won't break the website. 

So I made the `onClick` be a string type which gets TS to complain about it.

westwood846: Oooh lol. Okay makes sense. Though I kinda disagree with it not breaking when onClick is not a string. Well, out of scope I guess.

PR #616, opened 2021-03-01
Alert: Sync with designs

### LINKED ISSUE #517: Design-Sync Alert
https://app.abstract.com/share/dcc5438f-be8d-4348-84e9-cbf5d0f418f6?collectionLayerId=6afdbb97-6a1c-4f49-a031-ca3e9f55d80c&mode=design

* Make width responsive
* Update styles

westwood846:
Closes #517.

### DIFF
--- a/src/General/Alert/Alert.test.tsx
+++ b/src/General/Alert/Alert.test.tsx
@@ -276,6 +276,7 @@ describe('<Alert/> timeout', () => {
   test('autoCloseTimeout should be cleared before component unmount', async () => {
     const useRefSpy = jest.spyOn(React, 'useRef');
     const clearTimeoutSpy = jest.spyOn(window, 'clearTimeout');
+    const setTimeoutSpy = jest.spyOn(window, 'setTimeout');
     const { rerender, unmount } = render(
       <Alert
         type="success"
@@ -297,9 +298,10 @@ describe('<Alert/> timeout', () => {
     );
 
     unmount();
-    expect(clearTimeoutSpy).toHaveBeenCalledTimes(2);
-    expect(clearTimeoutSpy.mock.calls[1][0]).toEqual(
-      useRefSpy.mock.results[1].value.current
+    expect(setTimeoutSpy.mock.calls.length).toEqual(
+      // No idea what's calling the additional clearTimeout here. It's not us,
+      // I checked.
+      clearTimeoutSpy.mock.calls.length + 1
     );
 
     clearTimeoutSpy.mockRestore();
--- a/src/General/Alert/Alert.tsx
+++ b/src/General/Alert/Alert.tsx
@@ -9,58 +9,43 @@ import {
   CloseIcon,
 } from '../Icon/components';
 import {
+  AlertClose,
   AlertContainer,
-  AlertContent,
-  AlertMessage,
   AlertIcon,
-  AlertTypeIconContainer,
+  AlertMessage,
 } from './AlertStyle';
 
 import { Greyscale, PrimaryColor, SecondaryColor } from '../../Utils/Colors';
+import { isUndefined } from 'lodash';
+import { DelayedUnmount } from '../../Utils/DelayedUnmount';
 
 export const Alert = ({
-  isOpen,
+  isOpen: externalIsVisible = true,
   autoClose,
-  onClose,
+  onClose: externalOnClose,
   type = 'info',
   className,
   message,
 }: Props) => {
-  const alertContainerRef = React.useRef<HTMLDivElement>();
-  const autoCloseTimeout = React.useRef<ReturnType<typeof setTimeout>>();
-  const [isVisible, setIsVisible] = React.useState<boolean>(isOpen);
+  const [internalIsVisible, setInternalIsVisible] = React.useState<boolean>(
+    true
+  );
+
+  const isVisible = isUndefined(externalIsVisible)
+    ? internalIsVisible
+    : externalIsVisible;
+
+  const internalOnClose = () => setInternalIsVisible(false);
+  const onClose = isUndefined(externalOnClose)
+    ? internalOnClose
+    : externalOnClose;
 
   React.useEffect(() => {
-    if (isOpen && !isVisible) {
-      setIsVisible(true);
+    if (isVisible && autoClose) {
+      const timeout = setTimeout(onClose, autoClose);
+      return () => clearTimeout(timeout);
     }
-    if (!isOpen && isVisible) {
-      setTimeout(() => setIsVisible(false), 300);
-    }
-  }, [isOpen, isVisible, setIsVisible]);
-
-  // focus alert element for keydown event
-  React.useEffect(() => {
-    if (isVisible && alertContainerRef.current) {
-      alertContainerRef.current.focus();
-    }
-  }, [isVisible]);
-
-  React.useEffect(() => {
-    if (autoCloseTimeout.current) {
-      clearTimeout(autoCloseTimeout.current);
-    }
-
-    if (isOpen && autoClose) {
-      autoCloseTimeout.current = setTimeout(() => {
-        onClose();
-      }, autoClose);
-    }
-  }, [isOpen, isVisible, setIsVisible, autoClose, onClose]);
-
-  React.useEffect(() => {
-    return () => clearTimeout(autoCloseTimeout.current);
-  }, []);
+  }, [isVisible, autoClose, onClose]);
 
   const handleKeyDown = React.useCallback((onClose: () => void) => {
     const listener = (e: React.KeyboardEvent<HTMLDivElement>) => {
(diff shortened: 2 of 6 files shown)


src/General/Alert/AlertStyle.ts
```diff
@@ -31,16 +30,18 @@ export const AlertContainer = styled.div<AlertContainerProps>`
   transform: ${({ isVisible }) =>
     isVisible ? 'translate(-50%, 0%)' : 'translate(-50%, -100%)'};
   display: flex;
-  box-shadow: 0 10px 30px 0 rgba(0, 0, 0, 0.12);
-  padding: 10px;
+  box-shadow: ${Shadow.down4};
+  padding: 12px;
```
rosslo, line 44: Should we use ${Spacing.12} here? import { Spacing } from 'src/Layout/Spacing'

src/General/Alert/AlertStyle.ts
```diff
@@ -51,27 +52,22 @@ export const AlertContainer = styled.div<AlertContainerProps>`
       case 'success':
         return `
           background: #EEF6F4;
```
rosslo: Should change to `SecondaryColor.darkgreen`

And the following statuses should also update the `background-color`.

rosslo: We also need to style Box opacity 24%.
image

src/General/Alert/AlertStyle.ts
```diff
@@ -31,16 +30,18 @@ export const AlertContainer = styled.div<AlertContainerProps>`
   transform: ${({ isVisible }) =>
     isVisible ? 'translate(-50%, 0%)' : 'translate(-50%, -100%)'};
   display: flex;
-  box-shadow: 0 10px 30px 0 rgba(0, 0, 0, 0.12);
-  padding: 10px;
+  box-shadow: ${Shadow.down4};
+  padding: 12px;
   background-color: '#FFFAEA';
```
rosslo: Minor: looks like this background-color is unused, could you help to remove it? 🙏

rosslo:
Should `Alert` be full-screen width when the content is long?
image

src/General/Alert/AlertStyle.ts
westwood846: tbh I don't see the value in using those constants over a `px` value. It's not like we're ever going to change the values of those constants.

westwood846: Also I'd have to use `Spacing['12']` because `Spacing.12` is not valid ^^

src/General/Alert/AlertStyle.ts
westwood846: I asked for the right colors and didn't get an answer, but I think I've got it right now: https://glints.slack.com/archives/C018BD6GVS8/p1614576798001000

src/General/Alert/AlertStyle.ts
westwood846: Done.

src/General/Alert/Alert.test.tsx
```diff
@@ -297,9 +298,10 @@ describe('<Alert/> timeout', () => {
     );
 
     unmount();
-    expect(clearTimeoutSpy).toHaveBeenCalledTimes(2);
-    expect(clearTimeoutSpy.mock.calls[1][0]).toEqual(
-      useRefSpy.mock.results[1].value.current
+    expect(setTimeoutSpy.mock.calls.length).toEqual(
+      // No idea what's calling the additional clearTimeout here. It's not us,
+      // I checked.
+      clearTimeoutSpy.mock.calls.length + 1
```
westwood846, line 304: I was running into trouble with this test because I now have two timeouts (one in Alert and one in DelayedUnmount). But I figured this is really just testing that all timeouts are cleared in the end.

src/General/Alert/Alert.tsx
```diff
@@ -8,59 +8,40 @@ import {
...
+  const isVisible = isUndefined(externalIsVisible)
+    ? internalIsVisible
+    : externalIsVisible;
 
-  // focus alert element for keydown event
-  React.useEffect(() => {
-    if (isVisible && alertContainerRef.current) {
-      alertContainerRef.current.focus();
-    }
-  }, [isVisible]);
```
westwood846, line 47: Can just do this inline below.

src/General/Alert/Alert.tsx
```diff
@@ -8,59 +8,40 @@ import {
...
+  );
 
-  React.useEffect(() => {
-    if (isOpen && !isVisible) {
-      setIsVisible(true);
-    }
-    if (!isOpen && isVisible) {
-      setTimeout(() => setIsVisible(false), 300);
-    }
-  }, [isOpen, isVisible, setIsVisible]);
```
westwood846, line 40: Simplified the logic so that there's only one var that determines whether the alert is open/visible or not.

src/General/Alert/Alert.tsx
```diff
@@ -98,36 +79,37 @@ export const Alert = ({
...
-      onKeyDown={handleKeyDown(onClose)}
-      ref={alertContainerRef}
-    >
-      <AlertTypeIconContainer>
-        <AlertTypeIcon color={alertColor} />
-      </AlertTypeIconContainer>
-      <AlertContent className="alert-content">
-        <AlertMessage className="alert-message">{message}</AlertMessage>
+  return (
+    <DelayedUnmount mount={isVisible} delay={290}>
```
westwood846, line 89: Decreased to 290ms. The close animation duration is 300ms and if the unmount delay is equal to that, the alert will "flicker" back to its open position briefly before unmounting.

src/General/Alert/AlertStyle.ts
```diff
@@ -1,117 +1,87 @@
...
-  display: flex;
-  justify-content: space-between;
+  background: #c5ecfa;
+  &[data-type='success'] {
+    background: #c2e2da;
+  }
+  &[data-type='warning'] {
+    background: #fde9cc;
+  }
+  &[data-type='danger'] {
```
rosslo, line 65: This is neat!

src/General/Alert/Alert.tsx
```diff
@@ -9,58 +9,44 @@ import {
...
+  onClose: externalOnClose,
   type = 'info',
   className,
   message,
 }: Props) => {
-  const alertContainerRef = React.useRef<HTMLDivElement>();
-  const autoCloseTimeout = React.useRef<ReturnType<typeof setTimeout>>();
-  const [isVisible, setIsVisible] = React.useState<boolean>(isOpen);
+  const [internalIsVisible, setInternalIsVisible] = React.useState<boolean>(
+    true
```
rosslo, line 31: Do we need `internalIsVisible`? Was thinking if we set the default value of `isOpen` `true` can avoid having it.

src/General/Alert/Alert.tsx
```diff
@@ -9,58 +9,44 @@ import {
...
   AlertContainer,
-  AlertContent,
-  AlertMessage,
   AlertIcon,
-  AlertTypeIconContainer,
+  AlertMessage,
 } from './AlertStyle';
 
 import { Greyscale, PrimaryColor, SecondaryColor } from '../../Utils/Colors';
+import { Box } from '../../Layout/Box';
```
rosslo: eslint warning here

src/General/Alert/Alert.tsx
rosslo: I think we leave a comment for this kinda magic number. 🙏

src/Utils/DelayedUnmount.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { useEffect, useLayoutEffect, useState } from 'react';
```
rosslo: eslint warning here 🙏

src/Utils/DelayedUnmount.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+interface Props {
+  mount: boolean;
+  delay?: number;
+}
+
+export const DelayedUnmount: React.FC<Props> = ({
+  children,
+  mount,
+  delay = 300,
```
rosslo, line 11: This might be redundant since we have hard-coded it `290`

src/General/Alert/Alert.tsx
rosslo: Hmm, what if the user passes the `isOpen={true}` and `auto={300}` but doesn't pass `onClose`. The Alert won't be closed right?

src/General/Alert/Alert.test.tsx
rosslo: Haven't figured out why there is an extra `clearTimeout` called. But I found the caller should be `DelayedUnmount` when running `unmount()`.

src/General/Alert/Alert.tsx
westwood846: Correct. I think that's reasonable behavior, because the alert would be doing exactly what the user is asking: Be open.

src/General/Alert/Alert.test.tsx
westwood846: >But I found the caller should be DelayedUnmount when running unmount().
No I don't think so, that is one of the two calls that I can account for. I tried adding log messages to both effect hook cleanup functions (one in Alert and one in DelayedUnmount) and both get called only once.

src/General/Alert/Alert.tsx
westwood846: Cannot :grin: If I do that then there won't be any way to close the alert (without controlling it).

src/Utils/DelayedUnmount.tsx
westwood846: Well I was trying to keep this component general so that it can be reused in other places too.

src/General/Alert/Alert.tsx
rosslo: Oh doesn't `auto` mean the controlled alert can be automatically closed after X milliseconds without passing `onClose`? 

Btw, I was worried about this change is not backward-compatible which might break something. Have you checked that? 🙏

src/General/Alert/Alert.tsx
rosslo: Base on this https://github.com/glints-dev/glints-aries/pull/616#discussion_r589909866, I think the alert should be able to be closed without controlling it.

src/General/Alert/Alert.tsx
```diff
@@ -9,58 +9,43 @@ import {
...
+  AlertMessage,
 } from './AlertStyle';
 
 import { Greyscale, PrimaryColor, SecondaryColor } from '../../Utils/Colors';
+import { isUndefined } from 'lodash';
+import { DelayedUnmount } from '../../Utils/DelayedUnmount';
 
 export const Alert = ({
-  isOpen,
+  isOpen: externalIsVisible = true,
```
rosslo: This will break `const isVisible = isUndefined(externalIsVisible) ? internalIsVisible : externalIsVisible;`.

rosslo:
Hi @westwood846 , could you please not squash the commit during the code review? 🙏 
Because it will be easier to review your new code change based on the comments.

src/General/Alert/Alert.tsx
westwood846: There's only one usage on Employers and three on DST

Screenshot from 2021-03-10 14-48-19
Screenshot from 2021-03-10 14-48-09
Screenshot from 2021-03-10 14-47-48
Screenshot from 2021-03-10 14-46-58

westwood846: They all render the Alert unconditionally and control it's state. I don't know how to trigger those alerts so I did not check them, but I created my own in Employers like this and it works:

```
const [open, setOpen] = useState(true);
return (
  <Alert
    type="danger"
    message="this works"
    isOpen={open}
    onClose={() => setOpen(false)}
    autoClose={7000}
/>
```

So clearly it is correct to have it call `onClose` when the specified timeout triggers.

src/General/Alert/Alert.tsx
westwood846: It can, if you don't control it then the internals take over. Behaves just the same, closes after the timeout, closes when clicked, and stays open until clicked without timeout.

src/General/Alert/Alert.tsx
westwood846: Right, should not have committed that!

westwood846:
> Hi @westwood846 , could you please not squash the commit during the code review? 
> Because it will be easier to review your new code change based on the comments.

Can! Huh, GitHub does not have versions like GitLab. I was committing with that in mind.

rosslo (approved):

src/General/Alert/Alert.tsx
```diff
@@ -140,7 +126,7 @@ export interface Props {
   /** A function to close Alert. onClose will be triggered if the close icon is clicked or the Alert has ran its autoClose duration or ESC/Enter is pressed */
   onClose(): void;
```
rosslo, line 127: Since `onClose` is required here, I guess probably we don't have an uncontrolled Alert. We can create an issue to investigate it later.

PR #617, opened 2021-03-01
Textarea: Remove the row state

ritikamotwani:
The rows state basically controls the max and min height of the textarea. From 4 rows it starts expanding according to the scroll height rows and reaches a max value of 12.

To remove that state we'll need a max and min height to set the textarea component. Not sure if removing the rows is the best solution now, as most of the UI components I checked use rows for this purpose.

This solution is a bit taken from: https://codesandbox.io/s/pedantic-browser-zktd5?file=/src/TextArea.tsx

### DIFF
--- a/src/Input/Textarea/Textarea.tsx
+++ b/src/Input/Textarea/Textarea.tsx
@@ -11,6 +11,7 @@ import {
 
 const MIN_ROWS = 4;
 const MAX_ROWS = 12;
+// Rows are required to set the textarea height depending on the content added.
 
 /** Use <code>onChange</code> to listen to input changes. Use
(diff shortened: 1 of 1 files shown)


src/Input/Textarea/Textarea.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { FC, HTMLAttributes, RefObject } from 'react';
+import React, { FC, HTMLAttributes, RefObject, CSSProperties } from 'react';
```
ritikamotwani: Will remove this.

rosslo (approved):

PR #619, opened 2021-03-03
Select: Blur on outside mousedown instead of click

### LINKED ISSUE #582: Select isFocus state is stalled after select text on text input
Select isFocus state is stalled after select text on text input, so the select dropdown is not shown.

https://www.loom.com/share/16e7fca028eb4576bc17c77328e42dd5

westwood846:
This avoids a problem where selecting text in the input by dragging the mouse over the text and releasing the mouse outside the Select component closed the Select dropdown too early.

Fixes #582 

Can see here: https://5fbcb813bd13c900214bb971-vabztpcise.chromatic.com/?path=/docs/input-select--default-select

### DIFF
--- a/src/Input/Select/Select.tsx
+++ b/src/Input/Select/Select.tsx
@@ -157,8 +157,9 @@ const Select: React.FC<Props> & {
         setIsFocus(false);
       }
     };
-    document.addEventListener('click', onClickOutside, false);
-    return () => document.removeEventListener('click', onClickOutside, false);
+    document.addEventListener('mousedown', onClickOutside, false);
+    return () =>
+      document.removeEventListener('mousedown', onClickOutside, false);
   }, []);
 
   React.useEffect(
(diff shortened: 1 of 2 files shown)


src/Input/Select/Select.tsx
```diff
@@ -157,7 +157,7 @@ const Select: React.FC<Props> & {
         setIsFocus(false);
       }
     };
-    document.addEventListener('click', onClickOutside, false);
+    document.addEventListener('mousedown', onClickOutside, false);
     return () => document.removeEventListener('click', onClickOutside, false);
```
tingtingdisco: this line needed to be change to `mousedown` as well. 

This fix is so smart! Thank u @westwood846

tingtingdisco (approved):
just line 161 needed to be change as well.

src/Input/Select/Select.tsx
westwood846: Thanks :)

PR #639, opened 2021-06-18
Select: Set noOptionResult to null to hide no result option

tingtingdisco:
We have a requirement that when there is no option found in a Select component, we don't want it to display `No result found`.

https://glints.atlassian.net/browse/GM-840


westwood846 (approved):

src/Input/Select/Select.test.tsx
```diff
@@ -261,6 +263,18 @@ describe('when no results are found', () => {
   });
 });
 
+describe('when no results are found and noOptionResult is null', () => {
+  it('should display "No results found"', () => {
```
franciscrispin, line 267: Minor: Should be "should *not* display"? :sweat_smile:

PR #660, opened 2021-09-13
Tooltip: Support html content

vinhnglx:
Tooltip supports html content.

### DIFF
--- a/src/Display/Tooltip/Tooltip.tsx
+++ b/src/Display/Tooltip/Tooltip.tsx
@@ -18,6 +18,7 @@ export const Tooltip: FC<Props> = ({
   classes = {},
   children,
   text,
+  textAsString,
   position = 'top',
   ...defaultProps
 }) => {
@@ -47,12 +48,14 @@ export const Tooltip: FC<Props> = ({
     }
   };
 
+  const arialLabel = typeof text === 'string' ? text : textAsString;
+
   return (
     <TooltipContainer
       className={classNames('aries-tooltip', classes.container)}
       role="tooltip"
       aria-hidden={isShow ? 'false' : 'true'}
-      aria-label={text}
+      aria-label={arialLabel}
       // The tooltip does not close on iOS devices because of this issue https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event#Safari_Mobile
       // Adding onTouchStart and onTouchEnd as a workaround
       // On mobile, it shows the tooltip on touch and hides the tooltip when the touch is released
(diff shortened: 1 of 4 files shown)


src/Display/Tooltip/Tooltip.tsx
```diff
@@ -52,7 +52,6 @@ export const Tooltip: FC<Props> = ({
       className={classNames('aries-tooltip', classes.container)}
       role="tooltip"
       aria-hidden={isShow ? 'false' : 'true'}
-      aria-label={text}
```
westwood846, line 55: Probably don't wanna remove this without a good reason: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute

Screenshot from 2021-09-14 08-27-05
Screenshot from 2021-09-14 08-26-43

Maybe this was the reason why the Aries team didn't allow `ReactNode` before. The only solutions I can think of are:
* Get the `innerText` of `text`: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText
* Somehow allow passing in a pure string representation of the tooltip through a prop

westwood846 (changes requested):

src/Display/Tooltip/Tooltip.tsx
vinhnglx: Oh. Now I understand why we are not using `ReactNode`. Let me see what I can do.

westwood846: Yes I think this is the best we can do at the moment.

But we should still make sure that pre-existing `string` `text`s still get used for the label, for backwards-compatibility. So if `textAsString` is not given, it should use `text` for the label (should type-check first though).

vinhnglx: Alright, I just made a change.

- With string, it will be (textAsString is optional)
```javascript
<Tooltip
    position="top"
    text="Lorem Ipsum"
  >
    <Tag>
      Glints Aries
    </Tag>
</Tooltip>
```

- With html, it will be (textAsString is required)
```javascript
<Tooltip
    position="top"
    text={<b>Hello World</b>}
    textAsString="Hello World"
  >
    <Tag>
      Glints Aries
    </Tag>
</Tooltip>
```

westwood846: Hi, nice solution! Just one more thing: There's a lot of duplication between the two interfaces now. To fix that, I'd suggest you create an interface `BaseProps` which `HtmlTooltip` and `TextTooltip` extend!

westwood846 (changes requested):

westwood846 (approved):

Issue #664, opened 2021-10-02
Some Storybook Pages Render Empty With Error `children.match is not a function`


westwood846:
I think this was originally found by @vinhnglx. Probably similar to the issue https://github.com/glints-dev/glints-aries/pull/658/commits/952ed721cbd092dad79cde2591e54736f5e166c2 attempted to fix.

![Screenshot from 2021-10-02 10-17-52](https://user-images.githubusercontent.com/3341028/135700944-465185c4-6a73-46b3-a6ef-cfe3f573f44c.png)
![Screenshot from 2021-10-02 10-17-44](https://user-images.githubusercontent.com/3341028/135700946-3262907f-af21-4ea6-a3e0-8d3b31273a02.png)

westwood846:
Related issues:
* https://github.com/storybookjs/storybook/issues/15359
* https://github.com/storybookjs/storybook/issues/15932

PR #665, opened 2021-10-02
Remove all <code> elements from docs

westwood846:
I don't know why. I don't know how. `<code>` just produces an error (https://github.com/glints-dev/glints-aries/issues/664) but luckily `<Code>` does the same thing without the error.


vinhnglx:
LGTM.

vinhnglx (approved):

PR #675, opened 2021-12-08
Target React 17

### LINKED ISSUE #674: Update peerDependencies to react@^17
To prepare for the move to React 17, we should allow it as a peer dependency. I tested the upgrade in our internal projects and `glints-aries` appears to work fine.

yjwong:
Allow React 17 in peerDependencies, while keeping backwards compatibility with codebases that still use React 16. Due to a change in typings, `React.ComponentPropsWithoutRef` must now use the "type" keyword instead of an interface.

There were also some general cleanups, including upgrading Jest and testing-library and the removal of several dependencies that were no longer used.

Closes #674


src/General/Alert/Alert.test.tsx
```diff
@@ -297,15 +296,23 @@ describe('<Alert/> timeout', () => {
...
-      clearTimeoutSpy.mock.calls.length + 1
+    act(() => void unmount());
+
+    // The DelayedUnmount component, used by Alert, also calls setTimeout,
+    // in addition to the one Alert calls itself. Both timeouts should be
+    // unregistered. Other than this, the remaining setTimeout is called by
+    // the React framework (_flushCallback).
+    const actualSetTimeoutCalls = setTimeoutSpy.mock.calls.filter(
+      ([handler, timeout]) =>
+        timeout !== 0 && !String(handler).includes('_flushCallback')
```
franciscrispin, line 307: Am curious how you foud out that the remaining setTimeout was called by the framework `_flushCallback` haha

package.json
```diff
@@ -59,23 +59,22 @@
...
-    "@types/react-dom": "^16.8.4",
-    "@types/react-test-renderer": "^16.9.0",
+    "@types/react": "^17.0.37",
+    "@types/react-dom": "^17.0.11",
+    "@types/react-test-renderer": "^17.0.1",
     "@types/styled-components": "^5.1.1",
     "@types/styled-system": "^5.1.10",
-    "@typescript-eslint/eslint-plugin": "^2.21.0",
-    "@typescript-eslint/parser": "^2.21.0",
+    "@typescript-eslint/eslint-plugin": "^5.6.0",
```
franciscrispin, line 76: Just a note that I had to upgrade to node v12.22.0 to use @typescript-eslint/eslint-plugin https://github.com/typescript-eslint/typescript-eslint/blob/main/package.json#L55

But my node version (v12.20.1) might be lagging behind because I haven't ran our react codebases in a while.

franciscrispin (approved):
LGTM, thanks!

src/General/Alert/Alert.test.tsx
yjwong: Haha I hijacked the `setTimeout` function by creating a mock implementation that printed the call stack, something like:

```ts
const origSetTimeout = window.setTimeout;
const setTimeoutSpy = jest.spyOn(window, 'setTimeout').mockImplementation((handler, ms) => {
  console.log(new Error().stack);
  return origSetTimeout(handler, ms);
});
```

package.json
yjwong: Oooh, thanks. I realised we didn't really declare `engines` for the package.json too - maybe will add it in the future.

PR #684, opened 2022-09-06
Gallery: Render children directly instead of creating new image tags

vinhnglx:


src/Display/Gallery/Gallery.test.tsx
```diff
@@ -68,6 +74,14 @@ describe('<Gallery /> rendering', () => {
     );
     expect(asFragment()).toMatchSnapshot();
   });
+
+  it('should return console error when item is invalid children', () => {
```
marthayaputra, line 78: Nice, thanks for adding this test. Can we add one for rendering GlintsPicture? Or maybe we could also add a storybook too

src/Display/Gallery/Gallery.tsx
```diff
@@ -50,6 +50,16 @@ const Gallery = ({
     [visible]
   );
 
+  React.useEffect(function checkChildrenType() {
```
marthayaputra: Do we need useEffect to achieve this?

marthayaputra (approved):
LGTM

PR #691, opened 2022-12-02
Utils: @next foundation colors

ninariccimarie:
https://glints.atlassian.net/browse/SS-721

## Figma
[Branding Color](https://www.figma.com/file/J4wDmLPXT3BChWIasEWpUu/Glints-Design-System?node-id=3938%3A31072&t=My6NMmxkmtuy8mpb-4)
[Product Color](https://www.figma.com/file/J4wDmLPXT3BChWIasEWpUu/Glints-Design-System?node-id=3925%3A30153&t=My6NMmxkmtuy8mpb-4)

## Storybook update
Screenshot 2022-12-01 at 5 05 42 PM

## Nested Tree Shaking Evidence
https://glints.atlassian.net/browse/SS-721?focusedCommentId=22479

### DIFF
--- a/src/Utils/@next/colors/blue.ts
+++ b/src/Utils/@next/colors/blue.ts
@@ -1,11 +1,7 @@
-const blue = {
-  Brand: '#0BAEEC',
-  S8: '#EAF9FF',
-  S20: '#CAEEFB',
-  S54: '#6CC9EC',
-  S95: '#0BAEEC',
-  S99: '#017EB7',
-  S100: '#00568C',
-};
-
-export default blue;
+export const Brand = '#0BAEEC';
+export const S08 = '#EAF9FF';
+export const S20 = '#CAEEFB';
+export const S54 = '#6CC9EC';
+export const S95 = '#0BAEEC';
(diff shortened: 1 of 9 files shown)


marthayaputra:
@ninariccimarie , I think it's better to separate each named color to its own const for a better tree-shaking, like if my app only need 1 or 2 blue variants, I don't have to import all blue variant.

ninariccimarie:
thanks @marthayaputra I've improved the tree shaking by following this: https://github.com/webpack/changelog-v5/blob/master/README.md#nested-tree-shaking we should note that this is only compatible with webpack 5 so we must upgrade other FE services that aren't in v5 yet

marthayaputra:
> thanks @marthayaputra I've improved the tree shaking by following this: https://github.com/webpack/changelog-v5/blob/master/README.md#nested-tree-shaking we should note that this is only compatible with webpack 5 so we must upgrade other FE services that aren't in v5 yet

Oh nice work!

marthayaputra (approved):
LGTM

PR #702, opened 2023-01-06
Add Input

marthayaputra:


src/@next/Currency/Currency.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const rawValue = getRawNumber(e.currentTarget.value);
+    onChange(rawValue);
+    setFormattedValue(formatter.format(rawValue));
+  };
+
+  return (
+    <StyledCurrency
+      type="text"
+      prefix={<div>$</div>}
```
ninariccimarie: I think we can include `currency` prop and get the correct symbol from [formatToParts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/formatToParts) by providing options. based on locale

src/@next/Icon/icons/ri-search.svg
```diff
@@ -0,0 +1,10 @@
+<svg viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
ninariccimarie: can you try and check whether passing the fill prop renders the color? I think we need to remove the `fill="none"` in svg for it to work 🤔

src/@next/Currency/Currency.tsx
```diff
@@ -0,0 +1,68 @@
...
+export type CurrencyProps = Omit<
+  InputProps,
+  'type' | 'prefix' | 'onChange' | 'value'
+> & {
+  locale?: string;
+  value?: number;
+  onChange?: (value: number) => void;
+};
+
+export const Currency = ({
```
ninariccimarie: should we add `*Input` suffix to them to clearly indicate these component are Input components?

src/@next/Text/Text.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { useState } from 'react';
+import { Icon } from '../Icon';
+import { Input, InputProps } from '../Input/Input';
+
+export type TextProps = Omit<InputProps, 'type'> & { canClear?: boolean };
+
+export const Text = ({
```
ninariccimarie: I think we should also rename this since Text and Typography are often used interchangeably. maybe we can use `TextField` instead?

src/@next/Currency/Currency.tsx
marthayaputra: The problem with that is currency symbol varies, and no guarantee the style will be correct. If you see the managed candidate they put the actual currency outside of the input, I think "$" is just an indicator of money

src/@next/Icon/icons/ri-search.svg
marthayaputra: it will, I did that previously, and that technique is used in our current icons.

src/@next/Text/Text.tsx
marthayaputra: TextField is actually to use it for `form`, it is another control compose of this. But yeah, let's put Input suffix for all these

src/@next/Currency/Currency.tsx
ninariccimarie: ohh you're right, the currency is outside 👍

src/@next/Currency/Currency.tsx
marthayaputra: updated

ninariccimarie (approved):
LGTM 🚀

PR #703, opened 2023-01-06
chore: restructure @next folder

marthayaputra:


ninariccimarie (changes requested):
can you help to update the gitignore file instead of `src/General/@next/Icon/components` to `src/@next/Icon/components`?

I was thinking we should also put the non component directories (the ones that start with small letters) inside `src/@next/utilities` instead to have an organized separation.

marthayaputra:
> can you help to update the gitignore file instead of `src/General/@next/Icon/components` to `src/@next/Icon/components`?
> 
> I was thinking we should also put the non component directories (the ones that start with small letters) inside `src/@next/utilities` instead to have an organized separation.

Yeah, good idea updated

ninariccimarie (approved):
thanks! LGTM!

PR #707, opened 2023-01-10
feat: add tabs

marthayaputra:


src/@next/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,71 @@
...
+  StyledUl,
+} from './TabStyle';
+
+export type TabModel = {
+  id?: string;
+  content: React.ReactNode;
+};
+
+export type TabsProps = {
+  tabs: TabModel[];
```
ninariccimarie, line 25: can we add a description for TabModel to include what the object looks like? so it'll be easier for users to know what it expectsfrom looking at the storybook
Screenshot 2023-01-11 at 4 09 20 PM

src/@next/Tabs/Tab.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const Tab = ({ content, id, selected, onSelect }: TabProps) => {
+  return (
+    <StyledTabButton
+      data-selected={selected}
+      id={id}
+      onClick={() => onSelect()}
+    >
+      <StyledSpan>
+        <Typography as="span" variant="body1">
```
ninariccimarie, line 20: we might need to let consumer app pass their own content. since the one from Manage Candidate expects to render with a Badge as well. there's a possibility of the need to support other components as well like Icon

src/@next/Tabs/Tabs.tsx
marthayaputra: sure will add that

src/@next/Tabs/Tab.tsx
marthayaputra: Hmm this allows consumer to pass their own content, we're just creating a wrapper around it

src/@next/Tabs/Tabs.tsx
marthayaputra: updated

src/@next/Tabs/Tab.tsx
ninariccimarie: that would mean there's a Typography wrapper around their own content?

marthayaputra: Typography as `span` will be rendered as span, and on top of it we get the default font style for the children. So if you passing a text only it'll styled properly as the tab header font, if you passing a component with text (e.g badge) in it also get styled accordingly

marthayaputra: so basically we need 2 spans to generate the selected "border" style in the bottom

src/@next/Tabs/TabStyle.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  &&::after {
+    content: '';
+    position: absolute;
+    z-index: 1;
+    top: -1px;
+    right: -1px;
+    bottom: -1px;
+    left: -1px;
```
ninariccimarie: able to use shorthand `inset: -1px` ?

marthayaputra: Oh didn't know it has a shorthand, thanks! updated

ninariccimarie (approved):
LGTM 🚀

PR #709, opened 2023-01-11
feat: add Badge component

finalight:


marthayaputra:
@finalight is removable badge not part of this PR?

finalight:
> @finalight is removable badge not part of this PR?

i don't see the removable badge in the sirius web kit (https://www.figma.com/file/87fhOdF8vk1Uiy20bCqS85/Sirius-Web---UI-Kit-3.0-%5BQ4-2022%5D?node-id=51%3A498&t=LHVjOnX2q3LVCHPe-0)

plus, shopify polar has `Tag` component, which i think is more befitting to have the removable property based on our current UI

@marthayaputra

src/@next/Badge/BadgeStyle.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { B89 } from '../utilities/colors/green';
+import { B95 } from '../utilities/colors/neutral';
+import { S42, S87 } from '../utilities/colors/orange';
+import { B100 } from '../utilities/colors/red';
+import { S75 } from '../utilities/colors/yellow';
+import { space8 } from '../utilities/spacing';
+import { BadgeProps, BadgeStatusType } from './Badge';
+
+const badgeBackgroundColor = {
+  success: B89,
```
ninariccimarie: it would be better to match the type with BadgeStatusType so we get an error if there's anything missing from the types. you'll have to do something [similar](https://github.com/glints-dev/glints-aries/blob/v4/src/%40next/Typography/TypographyStyles.ts#L179) with Typography to make it work and define the BadgeStatusType differently.

src/@next/Badge/BadgeStyle.tsx
```diff
@@ -0,0 +1,35 @@
...
+  informational: S08,
+  warning: S42,
+  critical: B100,
+  promotion: S75,
+  new: S87,
+};
+
+const getBadgeBackgroundColor = (status: BadgeStatusType) => {
+  if (!(status in badgeBackgroundColor)) {
+    return B95;
```
ninariccimarie: we should probably warn user if status is not found similar to the one in Button https://github.com/glints-dev/glints-aries/blob/c539920df36a3c65a2cc3eb57d243148d6eef0eb/src/%40next/Button/ButtonStyle.ts#L108

src/@next/Badge/BadgeStyle.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return badgeBackgroundColor[status];
+};
+
+export const BadgeStyle = styled.div`
+  ${({ status }: BadgeProps) => {
+    return { 'background-color': getBadgeBackgroundColor(status) };
+  }}
+  width: fit-content;
+  padding: 2px ${space8};
+  border-radius: 20px;
```
ninariccimarie: there should be https://github.com/glints-dev/glints-aries/blob/c539920df36a3c65a2cc3eb57d243148d6eef0eb/src/%40next/utilities/borderRadius/BorderRadius.ts#L8

src/@next/Badge/BadgeStyle.tsx
ninariccimarie: can you use the `color.codeName` format (example: `Blue.S08`) so it's easier to know which color they are? we also need it in such format because of the design token mapping we'll be doing in the future.

src/@next/Badge/BadgeStyle.tsx
finalight: i don't think we should warn the issue for this case because it's default behaviour; user should expect this color if they don't pass in any attribute

src/@next/Badge/BadgeStyle.tsx
finalight: resolved

src/@next/Badge/BadgeStyle.tsx
finalight: resolved

ninariccimarie: additional discussion:
i think we should put default type to status and on the props, if there’s no passed status, we default it to `status=default` that means if you map the default type to the status to color mapping, we’ll be able to handle it

src/@next/Badge/BadgeStyle.tsx
ninariccimarie: we still would want to warn the user if they passed an incorrect status though. it would defeat the purpose of having status types if they passed a different one and they would wonder why the color is of default. it would make it easier for dev to know. (additional context regarding default type [here](https://github.com/glints-dev/glints-aries/pull/709#discussion_r1067863131))

finalight: added the console.warn

src/@next/Badge/BadgeStyle.tsx
finalight: added default status value, but its `neutral` instead of `default` since i'm following sirius UI kit

src/@next/Badge/Badge.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { Typography } from '../Typography';
+import { Neutral } from '../utilities/colors';
+import { BadgeStyle } from './BadgeStyle';
+
+export type BadgeStatusType =
```
ninariccimarie: can you replace this with https://github.com/glints-dev/glints-aries/pull/709/files#diff-938d02609f5a10f4d8c5ee533ad0b79fb3284622aa8603ef9b12d4d0a027413dR7-R15 (and remove the one from BadgeStyle)

and `export type BadgeStatusVariant = typeof BadgeStatusVariant[number];`

so both the component and the mapping refer the same type.

finalight: shifted the variant typings to the `Badge.tsx`

ninariccimarie (approved):
LGTM 🚀

PR #711, opened 2023-01-13
feat: add avatar component

finalight:


src/@next/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const AvatarSizeVariant = ['large', 'medium'] as const;
+
+export type AvatarBackgroundColorVariant =
+  typeof avatarBackgroundColorVariant[number];
+
+export type AvatarSizeVariant = typeof AvatarSizeVariant[number];
+
+export interface AvatarProps extends React.HTMLAttributes<HTMLDivElement> {
+  variant: AvatarBackgroundColorVariant;
```
ninariccimarie: this should also be optional since we should define a default based on the Figma

src/@next/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,39 @@
...
+  size?: AvatarSizeVariant;
+  initials?: string;
+}
+
+export const Avatar = ({ variant, size, initials, ...props }: AvatarProps) => {
```
ninariccimarie: based on Figma, both variant and size have default values. we can define them here instead of adding logic on the passed props below

src/@next/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,39 @@
...
+  return (
+    <AvatarStyle variant={variant} size={size ?? 'large'} {...props}>
+      <Typography
```
ninariccimarie, line 35: Screenshot 2023-01-17 at 2 56 44 PM
currently, initials goes beyond the circle but if we follow polaris, they hide the letters that goes beyond.

src/@next/Avatar/AvatarStyle.tsx
```diff
@@ -0,0 +1,88 @@
...
+  [variant in AvatarSizeVariant[number]]: number;
+} = {
+  ['large']: 40,
+  ['medium']: 32,
+};
+
+const getAvatarBackgroundColor = (variant: AvatarBackgroundColorVariant) => {
+  if (!(variant in avatarBackgroundColor)) {
+    console.warn(
+      `${variant} is not a valid variant for the backgroud color, default will be used`
```
ninariccimarie:
```suggestion
      `${variant} is not a valid variant for the background color, default will be used`
```

src/@next/Avatar/AvatarStyle.tsx
```diff
@@ -0,0 +1,88 @@
...
+  ['large']: 60,
+  ['medium']: 40,
+};
+
+const mobileAvatarSize: {
+  [variant in AvatarSizeVariant[number]]: number;
+} = {
+  ['large']: 40,
+  ['medium']: 32,
+};
```
ninariccimarie: can we do something like this for both medium and large styles:
```
const mediumAvatarSizeStyle = `
  width: 40px;
  height: 40px;
  
    width: 32px;
    height: 32px;
  }
`;
```
and map the style with the size so we eliminate the other called functions as much as we can inside the styled component.

src/@next/Avatar/Avatar.tsx
finalight: i've passed in the default value for the props

src/@next/Avatar/Avatar.tsx
finalight: i just take the first 2 characters of it, and that's it

src/@next/Avatar/AvatarStyle.tsx
finalight: fixed the typo

src/@next/Avatar/AvatarStyle.tsx
finalight: i did some refactoring of it, you can take a look

also, i've made sure that the snapshot didn't get updated to maintain the integrity of the outcome

src/@next/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export const Avatar = ({
+  variant = 'supplementary',
+  size = 'medium',
+  initials,
+  ...props
+}: AvatarProps) => {
+  return (
+    <AvatarStyle variant={variant} size={size ?? 'large'} {...props}>
```
ninariccimarie:
```suggestion
    <AvatarStyle variant={variant} size={size} {...props}>
```
we should be able to remove this because of the default value above.

src/@next/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <Typography
+        variant={size === null || size === 'medium' ? 'body1' : 'headline6'}
```
ninariccimarie:
```suggestion
        variant={size === 'medium' ? 'body1' : 'headline6'}
```
we should be able to remove this because of the default value above.

src/@next/Avatar/Avatar.tsx
ninariccimarie: @finalight was this communicated with Maeve?

finalight: nope, i assume that should be the behaviour

src/@next/Avatar/Avatar.tsx
finalight: done

src/@next/Avatar/Avatar.tsx
finalight: removed

src/@next/Avatar/Avatar.tsx
finalight: @ninariccimarie so discussed with the designer and it's fixed at 2 characters MAX, means `.slice(0,2)` should be the correct way

ninariccimarie: thanks @finalight

ninariccimarie (approved):
LGTM 🚀

PR #716, opened 2023-01-16
fix: center align text for badge for small screen

finalight:


marthayaputra (approved):

marthayaputra:
LGTM @finalight! Just that I think in the future, we better write the test as its own without abstracting it . Beside it's easier to read or debug when it fails, other than that we also can take advantage of the parallelism in playwright more if we separate them

PR #718, opened 2023-01-17
feat: add next banner

marthayaputra:


src/@next/Icon/icons/icons.ts
```diff
@@ -67,6 +67,13 @@ export const iconNames = [
   'ri-send-plane-line',
   'ri-settings-fill',
   'ri-settings-line',
+  'ri-checkbox-fill',
+  'ri-checkbox-line',
+  'ri-checkbox-indeterminate-fill',
+  'ri-checkbox-indeterminate-line',
+  'ri-search',
+  'ri-loader',
```
ninariccimarie: duplicates?

src/@next/Icon/icons/icons.ts
```diff
@@ -67,6 +67,13 @@ export const iconNames = [
...
+  'ri-checkbox-circle-fill',
```
ninariccimarie: can you move it above to be in alphabetical order?

src/@next/Icon/icons/icons.ts
```diff
@@ -139,5 +146,6 @@ export const iconsMappingComponent: { [name in IconNames]: SVGComponent } = {
   ['ri-send-plane-line']: Icons.RiSendPlaneLine,
   ['ri-settings-fill']: Icons.RiSettingsFill,
   ['ri-settings-line']: Icons.RiSettingsLine,
+  ['ri-checkbox-circle-fill']: Icons.RiCheckboxCircleFill,
```
ninariccimarie: can you also move this above to be in alphabetical order?

src/@next/Banner/index.tsx
```diff
@@ -0,0 +1 @@
+export { Banner, BannerProps } from './Banner';
```
ninariccimarie, line 1: any reason we exported BannerProps here?

src/@next/Banner/Banner.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const iconByStatus = iconNameStatusMap[status];
+  if (!iconByStatus) {
+    console.warn(`Status ${status} is not valid, default style will be used`);
+  }
+  const iconNameValue = iconName
+    ? iconName
+    : iconByStatus || iconNameStatusMap['info'];
+
+  return (
+    <div>
```
ninariccimarie: do we still need this div?

src/@next/Banner/Banner.tsx
```diff
@@ -0,0 +1,84 @@
...
+  iconName?: IconNames;
+  status?: 'success' | 'info' | 'warning' | 'critical';
+  children?: React.ReactNode;
+  action?: React.ReactNode;
+  secondaryAction?: React.ReactNode;
+  dismissable?: boolean;
+  onDismiss?: () => void;
+};
+
+const iconNameStatusMap = {
```
ninariccimarie: maybe we can type this so we don't need to cast the type of `iconNameValue`. we can use `Partial<IconNames>` maybe?

src/@next/Banner/BannerStyle.ts
```diff
@@ -0,0 +1,93 @@
+import styled from 'styled-components';
+import * as Breakpoints from '../utilities/breakpoints';
+import { borderRadius8 } from '../utilities/borderRadius';
+import { Blue, Green, Neutral, Orange, Red } from '../utilities/colors';
+
+export const StyledCloseIconWrapper = styled.div`
```
ninariccimarie, line 6: why don't we use button instead of div?

src/@next/Banner/BannerStyle.ts
```diff
@@ -0,0 +1,93 @@
...
+  cursor: pointer;
+  svg {
+    right: 0;
+    top: 0 !important;
```
ninariccimarie: would it be worth writing it this way instead? https://styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity

src/@next/Icon/icons/icons.ts
marthayaputra: Oh yes, thanks for that

src/@next/Icon/icons/icons.ts
marthayaputra: moved

src/@next/Icon/icons/icons.ts
marthayaputra: moved

src/@next/Banner/index.tsx
marthayaputra: Yeah, I think it's helpful if we export the props with components actually, probably need to do that for others later. I realized I often need third party component to type my abstracted layer of component above them.

src/@next/Banner/Banner.tsx
marthayaputra: Yeah good call!

src/@next/Banner/BannerStyle.ts
marthayaputra: Shorter styling, will need to remove some natural button attributes.  Do you see any risk with this?

src/@next/Banner/BannerStyle.ts
marthayaputra: Oh, didn't know about this, let me try. Thanks

src/@next/Banner/Banner.tsx
marthayaputra: nope, thanks!

src/@next/Banner/index.tsx
ninariccimarie: yeah, I experienced the same. then we probably need to export it in [@next/index ](https://github.com/glints-dev/glints-aries/pull/718/files#diff-5b6ecadc6bce07b9e2cf58256d9853ac20fa4c04f755c55979c70f0620032ec7R25) instead?

src/@next/Banner/BannerStyle.ts
ninariccimarie: i did read that buttons are harder to style vs divs. common risk would be accessibility if we use divs

marthayaputra: Oh, I added role "button" there to address that

src/@next/Banner/index.tsx
marthayaputra: Oh crap, I tot I did, thanks!

ninariccimarie (approved):
LGTM 🚀

PR #719, opened 2023-01-18
feat: add basic tag component

finalight:


finalight:
Note: i rename the old style from `tsx` to `ts`

src/@next/Tag/Tag.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { Neutral } from '../utilities/colors';
+import { TagStyle } from './TagStyle';
+
+export interface TagProps extends React.HTMLAttributes<HTMLDivElement> {
+  children?: React.ReactNode;
+}
+
+export const getTagChildren = (children: React.ReactNode) => {
+  if (typeof children !== 'string') {
+    console.warn('Tag children should only accept string');
```
ninariccimarie: I think this should be if the children is of type string, render it within the Typography, else, render the children. since we have an example that Tag can expect a Text with Icon components

finalight: hmm, i beg to differ on this

it would be wise at component level to determine what kind of icon to render by passing in the necessary attributes; something like what [polaris tag](https://polaris.shopify.com/components/tag) is doing

finalight: if we allow user to clump whatever they want inside the Tag component, the tag component is gonna look very weird. i think we also set a standard of how much customization the component should have

ninariccimarie: If you mean to reference Polaris, they allow more flexibility in terms of what consuming apps can render as children though (the `with custom content` example they have) which is what I'm suggesting. From the standard we currently have, we strictly type the props and check whether there's a specific composition within the component but we allow flexibility on what consuming apps can compose if the design allows them to. But of course, for the close button and links for example, we'll be controlling these.

finalight: got it @ninariccimarie 
fixed, please check again

src/@next/index.ts
```diff
@@ -1,3 +1,12 @@
+// Utilities
```
finalight, line 1: i re-order this file because it make more sense to import first, then export later

ninariccimarie (approved):
LGTM 🚀

PR #730, opened 2023-01-26
feat: Add IndexTable

ninariccimarie:


test/e2e/indexTable/indexTablePage.ts
```diff
@@ -0,0 +1,24 @@
...
+  readonly tableRow: Locator;
+  readonly moveToAction: Locator;
+
+  constructor(page: Page) {
+    super(page, '?path=/story/next-indextable--interactive');
+    this.container = page.locator('#storybook-preview-wrapper');
+    this.checkboxes = page
+      .frameLocator('internal:attr=[title="storybook-preview-iframe"i]')
+      .getByRole('checkbox');
+    this.tableRow = page
```
marthayaputra, line 17: I think this one should be done as checkboxes, return list and let the test to decide which row they wanna use

src/@next/IndexTable/index.ts
```diff
@@ -0,0 +1 @@
+export * as IndexTable from './IndexTable';
```
marthayaputra: missing export in `src/@next/index`?  Any reason to use `export as` here?

src/@next/IndexTable/components/BulkActions/BulkActions.tsx
```diff
@@ -0,0 +1,49 @@
...
+        active={active}
+        activator={activator}
+        onClose={onClose}
+      >
+        <Popover.Pane>
+          <div style={{ padding: '10px' }}>
+            <ButtonGroup>
+              <Button>Action</Button>
+              <Button icon={Suffix} iconPosition="right">
+                Move to
```
marthayaputra: I think this label should be a prop

marthayaputra (approved):

src/@next/IndexTable/components/BulkActions/BulkActions.tsx
ninariccimarie: this component is removed in favor of using polaris. thanks for spotting

src/@next/IndexTable/index.ts
ninariccimarie: my bad. i removed the `as` and also added the export

test/e2e/indexTable/indexTablePage.ts
ninariccimarie: i wasn't able to find a way to do it the same as checkboxes as there wasn't a `getByRole('tr')` or `getByRole('td')`. do you know any method?

PR #731, opened 2023-01-27
feature: enhance tag to have onRemove function

finalight:


src/@next/Tag/Tag.tsx
```diff
@@ -1,13 +1,26 @@
 import React from 'react';
+import { Icon } from '../Icon';
 import { Typography } from '../Typography';
 import { Neutral } from '../utilities/colors';
-import { TagStyle } from './TagStyle';
+import { TagContentStyle, TagRemoveContainerStyle, TagStyle } from './TagStyle';
 
-export interface TagProps extends React.HTMLAttributes<HTMLDivElement> {
+export interface RemovableTagProps
```
ninariccimarie: not sure what's the purpose of having this type? We can stick with just the `BasicTagProps` and rename it to `TagProps` instead.

```
export interface TagProps extends React.HTMLAttributes<HTMLDivElement> {
  children?: React.ReactNode;
  onRemove?: () => void;
}
```

src/@next/Tag/Tag.tsx
```diff
@@ -17,5 +30,21 @@ export const Tag = ({ children, ...props }: TagProps) => {
       children
     );
 
-  return <TagStyle {...props}>{content}</TagStyle>;
+  const removeButton = onRemove ? (
```
ninariccimarie:
```suggestion
  const removeButton = onRemove && (
```

we can do this instead of the ternary operator since onRemove is a function

src/@next/Tag/Tag.tsx
```diff
@@ -17,5 +30,21 @@ export const Tag = ({ children, ...props }: TagProps) => {
...
+      fill={Neutral.B40}
+      height={'20'}
+      onClick={() => onRemove?.(value)}
+    />
+  ) : null;
+
+  return (
+    <TagStyle {...props}>
+      <TagContentStyle onRemove={onRemove}>{content}</TagContentStyle>
+      <TagRemoveContainerStyle className={'icon-remove'}>
```
ninariccimarie: any reason for the added className?

src/@next/Tag/Tag.tsx
```diff
@@ -17,5 +34,41 @@ export const Tag = ({ children, ...props }: TagProps) => {
       children
     );
 
-  return <TagStyle {...props}>{content}</TagStyle>;
+  const removeButton = onRemove ? (
+    <TagRemoveContainerStyle
```
ninariccimarie: can we add `role="button"` to address accessibility risk since we are using div.
then i think we can remove `onMouseDown`, etc and use `:active`, etc in css instead. We will also be able to eliminate the [function](https://github.com/glints-dev/glints-aries/pull/731/files#diff-411820b74eeb16ad04767a6afdd0ed7e90abe1d6ebd043bb0bd85ac575ca5523R36-R45) inside the styled component this way. else, we can address using attributes too but i recommend above if it's possible

src/@next/Tag/Tag.tsx
finalight: i think this is resolved; i saw that it's outdated

src/@next/Tag/Tag.tsx
finalight: i've discussed with the designer bernard; we decide to remove the focused state for now

so probably the mouseDown/onMouseUp would be removed

finalight: the reason why i can't use `:active` for the class selector is because it only applies to the selected element, which in this case, is `TagRemoveContainerStyle`. The problem i'm trying to solve is that, when the user press/hover on the `TagRemoveContainerStyle` element, the parent element's css would change. 

so far in the world of css, i can't find a way to modify parent's element using css purely

src/@next/Tag/Tag.tsx
finalight: resoved

src/@next/Tag/Tag.tsx
finalight: className removed

src/@next/Tag/Tag.tsx
```diff
@@ -17,5 +32,39 @@ export const Tag = ({ children, ...props }: TagProps) => {
...
+      onBlur={() => setIsPressed(false)}
+      onMouseLeave={() => {
+        setIsPressed(false);
+        setIsHover(false);
+      }}
+      onMouseUp={() => setIsPressed(false)}
+      onMouseEnter={() => setIsHover(true)}
+    >
+      <Icon
+        name={'ri-close'}
```
ninariccimarie: please remove `{` and `}` when you pass strings

src/@next/Tag/Tag.tsx
```diff
@@ -17,5 +32,39 @@ export const Tag = ({ children, ...props }: TagProps) => {
...
+        fill={Neutral.B40}
+        height={'20'}
+        onClick={() => onRemove?.(value)}
```
ninariccimarie:
```suggestion
        onClick={onRemove}
```
i think you can just do this

src/@next/Tag/Tag.tsx
ninariccimarie: ahh I see, ok, thanks for giving me more context

src/@next/Tag/Tag.tsx
ninariccimarie: it seems this is still there

src/@next/Tag/Tag.tsx
finalight: i have this error, you got other suggestion on resolving it?

image

src/@next/Tag/Tag.tsx
finalight: okay, it's removed

src/@next/Tag/Tag.tsx
finalight: removed

src/@next/Tag/Tag.tsx
ninariccimarie: i think we could actually just remove the parameter inside the onRemove. `() => void` as from my initial suggestion. i checked polaris, they have the same typing

finalight: storybook will complain it though

image

finalight: i think this is done

ninariccimarie (approved):
LGTM 👍

test/e2e/tag/tag.spec.ts
```diff
@@ -2,17 +2,103 @@ import { test, expect, Page } from '@playwright/test';
...
+  new StoryBookPage(page, '?path=/story/next-tag--default');
 
-test('tag - standard size', async ({ page }) => {
-  const tagPage = getPage(page);
-  await tagPage.goto();
-  await expect(tagPage.container).toHaveScreenshot('tag-standard-size.png');
+const getRemoveableTagPage = (page: Page) =>
+  new StoryBookPage(page, '?path=/story/next-tag--removeable');
+
+test.describe('default tag', () => {
```
ninariccimarie: can you help to do the same as the comment here: https://github.com/glints-dev/glints-aries/pull/736#discussion_r1091595232

finalight: remove the describe test suite

now everything's flat

src/@next/Tag/TagStyle.ts
```diff
@@ -1,18 +1,66 @@
 import styled from 'styled-components';
 import { Breakpoints } from '..';
-import { borderRadius4 } from '../utilities/borderRadius';
-import { Blue } from '../utilities/colors';
+import {
+  borderRadius2,
```
ninariccimarie: we can remove since it's not used

src/@next/Tag/TagStyle.ts
```diff
@@ -1,18 +1,66 @@
...
+    padding: 0;
+  }
+
+  & svg:hover {
+    cursor: pointer;
+    background-color: ${Neutral.B95};
+    border-radius: ${borderRadiusHalf};
+  }
+
+  & svg:active {
```
ninariccimarie, line 56: can you include active state in the test?

finalight: that's the challenge i have, i am not sure how to do long press in playwright and then take screenshot

src/@next/Tag/TagStyle.ts
finalight: removed

src/@next/Tag/TagStyle.ts
finalight: added the visual test for pressed state

ninariccimarie (approved):
LGTM 🚀

PR #736, opened 2023-01-31
feat: add next Radio

ganiirsyadi:


test/e2e/radio/radio.spec.ts
```diff
@@ -0,0 +1,69 @@
+import { StoryBookPage } from '../storybookPage';
+import { Page, test, expect } from '@playwright/test';
+
+const getPage = (page: Page) =>
+  new StoryBookPage(page, '?path=/story/next-radio--interactive');
+
+test.describe('Radio - Standard Size', () => {
```
marthayaputra: Can we have a flat structure here? The reason is easier to pin point which test exactly failed

src/@next/Radio/Radio.tsx
```diff
@@ -0,0 +1,33 @@
...
+export type RadioProps = React.HTMLAttributes<HTMLInputElement> & {
+  label?: string;
+  helperText?: string;
+  disabled?: boolean;
+  name?: string;
+  checked?: boolean;
+};
+
+export const Radio = ({ ...props }: RadioProps) => {
+  const uuid = useMemo(() => uuidv4(), []);
```
marthayaputra: any reason for `useMemo` here?

test/e2e/radio/radio.spec.ts
ganiirsyadi: Can you give me an example? What is the flat structure did you mean?

src/@next/Radio/Radio.tsx
ganiirsyadi: It won't generate a new uuid for each render

src/@next/Radio/Radio.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { useMemo } from 'react';
+import { Label, LabelWrapper, RadioInput, RadioWrapper } from './RadioStyle';
+import { Typography } from '../Typography';
+import { v4 as uuidv4 } from 'uuid';
```
marthayaputra: can we use the existing `nextId` from  'react-generator-id' library?

src/@next/Radio/Radio.tsx
ganiirsyadi: I can use a `useState` and `useEffect` but using `useMemo` is a shorthand to it

src/@next/Radio/Radio.tsx
ganiirsyadi: Ah, Let me try it

src/@next/index.ts
```diff
@@ -36,6 +36,7 @@ export {
 } from './IndexTable';
 export { Pagination, SimplePagination, PaginationProps } from './Pagination';
 export { Popover, PopoverProps } from './Popover';
+export { Radio } from './Radio';
```
marthayaputra: can we also export the component props here?

test/e2e/radio/radio.spec.ts
marthayaputra: sure like this: https://github.com/glints-dev/glints-aries/blob/v4/test/e2e/alert/alert.spec.ts

we actually missed that for few components and addressing them here: https://github.com/glints-dev/glints-aries/pull/735

src/@next/index.ts
ganiirsyadi: Can

src/@next/Radio/Radio.tsx
marthayaputra: But why we even need to use them? I think this is just to assign a unique id right?

ninariccimarie:
@ganiirsyadi can we follow naming from design system? RadioButton instead of just Radio

ganiirsyadi:
> @ganiirsyadi can we follow naming from design system? RadioButton instead of just Radio

Can

src/@next/Radio/Radio.tsx
ganiirsyadi: so that we don't need to generate a new ID for every render that is useless

marthayaputra: I still don't think it needs useMemo or any hooks tho, unless it's a heavy task. Using useMemo also requires additional checks, and since we don't care about autogenerated id, it also has no value to preserve it 🤔

src/@next/Radio/Radio.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export type RadioProps = React.HTMLAttributes<HTMLInputElement> & {
+  label?: string;
+  helperText?: string;
+  disabled?: boolean;
+  name?: string;
+  checked?: boolean;
+};
+
+export const Radio = ({ ...props }: RadioProps) => {
```
ninariccimarie: can we destructure the props that were used below. it'll be easier to read and we will have lesser steps to access the prop. we are also doing similarly on the other components

src/@next/Radio/RadioStyle.ts
```diff
@@ -0,0 +1,105 @@
...
+  flex-direction: column;
+  gap: 4px;
+
+  ${({ disabled }) => {
+    console.log('disabled', disabled);
+    return (
+      disabled &&
+      `
+        span {
+          color: ${Colors.Neutral.B85} !important;
```
marthayaputra: can we avoid using !important but use this instead: https://styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity

src/@next/Radio/RadioStyle.ts
```diff
@@ -0,0 +1,105 @@
+import styled from 'styled-components';
+import { Colors } from '..';
+import { ScreenSize } from '../../Utils/StyleConfig';
+
+export const RadioWrapper = styled.div`
+  display: flex;
+  align-items: flex-start;
+  gap: 8px;
```
ninariccimarie: you can use `space8`

src/@next/Radio/RadioStyle.ts
```diff
@@ -0,0 +1,105 @@
...
+`;
+
+export const RadioInput = styled.input`
+  cursor: pointer;
+  appearance: none;
+  margin: 0;
+  width: 18px;
+  height: 18px;
+  border: 2px solid ${Colors.Neutral.B68};
+  border-radius: 50%;
```
ninariccimarie: you can use `borderRadiusHalf`

src/@next/Radio/RadioStyle.ts
```diff
@@ -0,0 +1,105 @@
...
+  transform: translateY(1.5px);
+
+  ::after {
+    content: '';
+    display: block;
+    border-radius: 50%;
```
ninariccimarie: you can use `borderRadiusHalf`

src/@next/Radio/RadioStyle.ts
```diff
@@ -0,0 +1,105 @@
...
+    width: 8px;
+    height: 8px;
+    margin: 3px;
+  }
+
+  @media (min-width: ${ScreenSize.tablet}px) {
```
ninariccimarie: we are using the one from utilities breakpoints 🙇 
ScreenSize will be deprecated next time in favor of the new design system

src/@next/Radio/RadioStyle.ts
```diff
@@ -0,0 +1,105 @@
...
+        background-color: ${Colors.Neutral.B85};
+      }
+    }
+  }
+`;
+
+export const LabelWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 4px;
```
ninariccimarie: you can also use `space4` here

src/@next/Radio/RadioStyle.ts
```diff
@@ -0,0 +1,105 @@
...
+
+type LabelProps = {
+  disabled?: boolean;
+};
+
+export const Label = styled.label<LabelProps>`
+  cursor: pointer;
+  display: flex;
+  flex-direction: column;
+  gap: 4px;
```
ninariccimarie: you can also use `space4` here

src/@next/Radio/RadioStyle.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  ${({ disabled }) => {
+    console.log('disabled', disabled);
```
ninariccimarie: need to cleanup?

src/@next/Radio/RadioStyle.ts
```diff
@@ -0,0 +1,105 @@
...
+  disabled?: boolean;
+};
+
+export const Label = styled.label<LabelProps>`
+  cursor: pointer;
+  display: flex;
+  flex-direction: column;
+  gap: 4px;
+
+  ${({ disabled }) => {
```
ninariccimarie: can you use `data-disabled` instead? we try to avoid using props inside styled components. you can check globally search it and see examples.

we avoid using props to improve performance

src/@next/Radio/index.tsx
```diff
@@ -0,0 +1 @@
+export { Radio } from './Radio';
```
ninariccimarie: can you follow the pattern on how we export in the index? this makes it easier for us to export all exports from here if we need them to be exported from `@next/index`

test/e2e/radio/radio.spec.ts
ganiirsyadi: Fixed

src/@next/Radio/Radio.tsx
ganiirsyadi: I have a personal experience where the generated ID is different between the server and client (browser) when using Next, and it produced unexpected bugs; I am afraid it will happen in the future. But if you think it's unnecessary, I will remove it. But still, in my opinion, there is no disadvantages for using the `useMemo.`

src/@next/Radio/Radio.tsx
ganiirsyadi: Updated

src/@next/index.ts
ganiirsyadi: Updated

src/@next/Radio/Radio.tsx
ganiirsyadi: Updated

src/@next/Radio/RadioStyle.ts
ganiirsyadi: Fixed

src/@next/Radio/Radio.tsx
ganiirsyadi: Fixed

src/@next/Radio/RadioStyle.ts
ganiirsyadi: Fixed

src/@next/Radio/RadioStyle.ts
ganiirsyadi: Fixed

src/@next/Radio/RadioStyle.ts
ganiirsyadi: Fixed

src/@next/Radio/RadioStyle.ts
ganiirsyadi: Fixed

src/@next/Radio/RadioStyle.ts
ganiirsyadi: Fixed

src/@next/Radio/RadioStyle.ts
ganiirsyadi: Fixed

src/@next/Radio/RadioStyle.ts
ganiirsyadi: Fixed

src/@next/Radio/index.tsx
ganiirsyadi: Fixed

src/@next/Radio/RadioStyle.ts
ganiirsyadi: Fixed

package.json
```diff
@@ -125,6 +125,7 @@
     "webpack-glob-entry": "2.1.1"
   },
   "dependencies": {
+    "@types/uuid": "^9.0.0",
```
marthayaputra: can we remove this dependency?

ganiirsyadi: Removed

src/@next/RadioButton/RadioButtonStyle.ts
```diff
@@ -0,0 +1,90 @@
...
+  ::after {
+    content: '';
+    display: block;
+    border-radius: ${borderRadiusHalf};
+    width: 8px;
+    height: 8px;
+    margin: 3px;
+  }
+
+  @media (min-width: ${large}) {
```
ninariccimarie: can you do `Breakpoints.large` instead?

ganiirsyadi: Updated

ninariccimarie (approved):
LGTM 🚀

PR #740, opened 2023-02-03
feat: add next modal

marthayaputra:


src/types/componentWithProvider.ts
```diff
@@ -0,0 +1,4 @@
+export type ComponentWithProviderProps<T> = {
```
ninariccimarie, line 1: just curious why we named it Provider? i can't really tell that it has open and close functions with the name 🤔

src/@next/Modal/useModal.tsx
```diff
@@ -0,0 +1,12 @@
+import { useContext } from 'react';
+import { ModalContext } from './ModalContext';
+
+export const useModal = () => {
+  const context = useContext(ModalContext);
+  if (!context) {
+    throw new Error(
+      'No ModalContext was provided via "ModalProvider", to use this hooks you must provide them in your app root.'
```
ninariccimarie:
```suggestion
      'No ModalContext was provided via "ModalProvider", to use this hook you must provide them in your app root.'
```

src/types/componentWithProvider.ts
marthayaputra: because it's used with Provider, it won't be limited to open and close. You have a suggestion? 🤔

ninariccimarie: you mean this would be used for combined one where we can add more apart from open and close in the future? I see it is currently used as being intersected with the Component's types which makes sense.

marthayaputra: yes, for component which are using providers

ninariccimarie (approved):
LGTM 🚀

PR #741, opened 2023-02-03
Index table styles

ninariccimarie:


src/@next/IndexTable/IndexTableStyle.ts
```diff
@@ -0,0 +1,1745 @@
...
+  .Polaris-Scrollable--hasBottomShadow {
+    box-shadow: var(--pc-scrollable-shadow-bottom);
+  }
+
+  .Polaris-Scrollable--hasTopShadow.Polaris-Scrollable--hasBottomShadow {
+    box-shadow: var(--pc-scrollable-shadow-top),
+      var(--pc-scrollable-shadow-bottom);
+  }
+
+  .Polaris-Popover {
```
marthayaputra, line 149: Do we still need Popover styles here, as they have been added in PopoverStyle?

src/@next/IndexTable/IndexTableStyle.ts
```diff
@@ -0,0 +1,1745 @@
...
+    --pc-inline-gap-lg: initial;
+    --pc-inline-gap-xl: initial;
+    gap: var(--pc-inline-gap-xs);
+    display: flex;
+    flex-wrap: var(--pc-inline-wrap);
+    align-items: var(--pc-inline-block-align);
+    justify-content: var(--pc-inline-align);
+  }
+
+  @media (min-width: 30.625em) {
```
marthayaputra: Do we need this? Also same for all media queries

ninariccimarie: ohh thanks for spotting this! yes, we can remove 👌

src/@next/IndexTable/IndexTableStyle.ts
ninariccimarie: hmm.. seems this one is breaking 🤔 
Screenshot 2023-02-03 at 4 42 13 PM

marthayaputra (approved):
LGTM

PR #744, opened 2023-02-07
feat: empty state

ninariccimarie:
- updated Storybook parameters to hide add on panel in canvas by default in order to capture bigger components
  - Affected Components to update screenshots:
    - Alert
    - IndexTable
    - Modal
    - Popover


package.json
```diff
@@ -129,7 +129,7 @@
     "classnames": "^2.2.6",
     "downshift": "^6.1.3",
     "moment": "^2.24.0",
-    "polaris-glints": "^15.4.0",
+    "polaris-glints": "^15.5.0",
```
ninariccimarie: updated because of IndexTable empty state to show headers

src/@next/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,74 @@
...
+  EmptyStateContainer,
+  EmptyStateContentContainer,
+  StyledButtonGroup,
+  StyledDescription,
+  StyledHelpText,
+  StyledImage,
+  StyledTitle,
+} from './EmptyStateStyle';
+
+export enum ImageName {
```
marthayaputra: I think we should avoid using `enum` in TypeScript and use simple object instead, here is the pitfall: https://dev.to/dvddpl/whats-the-problem-with-typescript-enums-2okj#:~:text=The%20issue%20with%20TypeScript%20enums,a%20minimum%20amount%20of%20friction.

src/@next/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,74 @@
...
+  imageName,
+  title,
+}: EmptyStateProps) => {
+  const showPrimaryButton = !!buttonPrimaryContent;
+  const showBasicButton = !!buttonBasicContent;
+  const displayButtons = showPrimaryButton || showBasicButton;
+  return (
+    <EmptyStateContainer>
+      {imageName && (
+        <StyledImage src={require(`../../../assets/image/${imageName}.png`)} />
```
marthayaputra: why do we need to use require here? 🤔 . Also could we move the assets related to EmptyState inside the folder?

src/@next/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,74 @@
...
+          </StyledTitle>
+        )}
+        {description && (
+          <StyledDescription variant="body1" color={Neutral.B40}>
+            {description}
+          </StyledDescription>
+        )}
+        {displayButtons && (
+          <StyledButtonGroup>
+            {showPrimaryButton && (
```
marthayaputra: can we derive showPrimaryButton and showBasic button from the content instead? Just like helpText here

src/@next/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export enum ImageName {
+  EMPTY_CARTON = 'empty-carton',
+  EMPTY_MAILBOX = 'empty-mailbox',
+  SAFETY_CONE = 'safety-cone',
+}
+
+export interface EmptyStateProps {
+  buttonBasicContent?: React.ReactNode;
+  buttonPrimaryContent?: React.ReactNode;
```
Afsal102: we would need callbacks (onClick) as props for basic and primary buttons? from what i can see these contents are added within the Button itself

marthayaputra: @Afsal102 wouldn't you be able to achieve that passing the whole component with onClick event handler attached?

But  @ninariccimarie , I think to strictly enforce only primary button and secondary buttons are rendered, we should accept label and event for those props and render the primary and basic in the component directly

ninariccimarie: thanks for the catch @Afsal102

src/@next/EmptyState/EmptyState.tsx
ninariccimarie: @marthayaputra I'm already deriving them from the button content. I just think it's more readable to name a const rather than inline `!!buttonPrimaryContent` for the condition

src/@next/EmptyState/EmptyState.tsx
ninariccimarie: ohh I didn't know about this. thanks!

src/@next/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,96 @@
...
+  StyledImage,
+  StyledTitle,
+} from './EmptyStateStyle';
+
+export const imageName = ['empty-carton', 'empty-mailbox', 'safety-cone'];
+export type ImageName = typeof imageName[number];
+export interface EmptyStateProps {
+  buttonBasicContent?: React.ReactNode;
+  buttonBasicLoading?: boolean;
+  buttonBasicOnClick?: React.MouseEventHandler<HTMLButtonElement>;
```
marthayaputra: Sorry @ninariccimarie I think there's a misunderstanding, we shouldn't pass ReactNode for button content as they could just pass anything which are not related to button. But we should type them as an object e.g: 
{
 label: string;
onClick: (args) => void
}

then on the component we render <Button onClick={() => basicButton.onClick()} >{basicButton.label}</Button>

I think my point to @Afsal102 is if we're passing ReactNode, we don't actually need to worry about the click event as they can just pass it as `<Button onClick={handleClick}>` themselves from the consuming app

Afsal102: @marthayaputra i am kind of confused on this statement, from the current implementation the button is already there, we can customise content of the button by passing the props, from ur suggestion if i pass a button again to the `buttonBasicContent`  it will be as follows

```
<Button> // from the existing impl
<Button> // pased via the props
</Button>
</Button>
```

would this be case(nested buttons)?

Afsal102: may be instead of just getting content of the button, how about we allow the clients to pass the whole button directly ie:

```
 {showBasicButton && (
                {buttonBasic} // we just render whats given from the props
            )}

```

ninariccimarie: @marthayaputra @Afsal102 I'm doing combination of both your initial comments. the reason why I didn't use just label was because of the design requirements.
1. we restrict to only 2 buttons, either primary or basic buttons
2. consuming app can freely pass anything inside those 2 buttons.

point 2 is the reason why i used React.ReactNode.

marthayaputra: @Afsal102 , ~~what I was saying if the type is  `ReactNode` basically you don't need an even handlers as props, because you can pass construct it in the button you passed, it's like passing a `children`. Jsx is just a sugar syntax for React.functions which are basically bunch of functions with parameters.~~

Right, my bad, I tot we passed the buttons as a whole as primary and basic

Anyway doesn't matter now as @ninariccimarie will change the type to literal objects instead of ReactNode, so now you can define the click function explicitly as prop

marthayaputra: @ninariccimarie what are the elements they can pass as a button's content? I tot from the last discussion we only limit to `primary` and `secondary` with text?

src/@next/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,96 @@
...
+  EmptyStateContainer,
+  EmptyStateContentContainer,
+  StyledButtonGroup,
+  StyledDescription,
+  StyledHelpText,
+  StyledImage,
+  StyledTitle,
+} from './EmptyStateStyle';
+
+export const imageName = ['empty-carton', 'empty-mailbox', 'safety-cone'];
```
Afsal102: @ninariccimarie we may have to narrow down the type here as currently we can pass any string value
```

export const imageName = ['empty-carton', 'empty-mailbox', 'safety-cone'] as const; 

```

src/@next/EmptyState/EmptyState.tsx
ninariccimarie: i had it clarified initially. I've tagged both of you on the thread.
I'll apply the changes based on the new requirement which is to only pass label instead of the flexibility that was initially decided.

src/@next/EmptyState/EmptyState.tsx
ninariccimarie: @Afsal102 I'm already using these as the type: https://github.com/glints-dev/glints-aries/pull/744/files#diff-7e72d5b4c46257b14b04072cd667efebe24feda83031d0275c67fcf50ff4fb6cR26
let me throw a warning if it's not found so JS can see the error since only TS files can throw the error for the typing

Afsal102: @ninariccimarie am aware of it, but even with this right u can pass any string value(not strict), but if u narrow the type down(readonly), the user will be prompted to select one of the values in `imageName` (type safety)

ninariccimarie: @Afsal102 what do you mean by narrow the type down?

Afsal102: > @ninariccimarie am aware of it, but even with this right u can pass any string value(not strict), but if u narrow the type down(readonly), the user will be prompted to select one of the values in `imageName` (type safety)

src/@next/EmptyState/EmptyState.tsx
ninariccimarie: dynamically importing now similar to Icon

src/@next/EmptyState/EmptyState.tsx
ninariccimarie: addressed @Afsal102

src/@next/EmptyState/EmptyState.tsx
ninariccimarie: updated these to pass a ButtonAction object @marthayaputra @Afsal102

marthayaputra (approved):

PR #747, opened 2023-02-10
feat: calculate input pading based on prefix  and suffix width

ganiirsyadi:


src/@next/Input/InputStyle.ts
```diff
@@ -58,12 +59,14 @@ export const StyledPrefixContainer = styled.div`
     height: 17px;
     width: 17px;
     margin: 10px 14px;
+    margin-right: ${space8};
```
ganiirsyadi: Narrower margin-right looks better imo:
image

src/@next/Input/InputStyle.ts
```diff
@@ -58,12 +59,14 @@ export const StyledPrefixContainer = styled.div`
...
     fill: ${Neutral.B40};
   }
 
   div {
     color: ${Neutral.B40};
     margin: 18px 12px;
+    margin-right: ${space4};
```
ganiirsyadi: Same here:
image

src/@next/CurrencyInput/CurrencyInput.tsx
```diff
@@ -9,13 +8,14 @@ export type CurrencyInputProps = Omit<
   locale?: string;
   value?: number;
   onChange?: (value: number) => void;
+  prefix?: string;
```
marthayaputra: can we name this currencyCode instead? Maybe we can assign the default one based on the `locale` using `Intl.NumberFormat.formatToParts` , can see `getRawNumber` function for more details. But also allowing user to override it by defining currencyCode

ganiirsyadi: The prefix will be filled with currencySymbol not currencyCode, ex: $ not USD

ganiirsyadi: Based on this Stackoverflow QnA https://stackoverflow.com/questions/50650503/get-the-currency-symbol-for-a-locale, it needs both locale and currencyCode to get the currency symbol

ganiirsyadi:
I also handled the dynamic suffix width.

src/@next/CurrencyInput/CurrencyInput.tsx
marthayaputra: Oh yeah from this prop I was under the assumption the prefix will be whatever which passed (user will need to pass "$") to display dollar sign. Ah okay, I think I should mentioned passing currencySymbol instead of currencyCode for the overriding part

src/Utils/CurrencyCodes.tsx
```diff
@@ -0,0 +1,179 @@
...
+  'STD',
+  'SYP',
+  'SZL',
+  'THB',
+  'TJS',
+  'TMT',
+  'TND',
+  'TOP',
+  'TRY',
+  'TTD',
```
marthayaputra: Why do we need this?

src/@next/CurrencyInput/CurrencyInput.tsx
```diff
@@ -47,21 +56,50 @@ export const CurrencyInput = React.forwardRef<
...
+    return currencySymbol;
+  };
+
   const [formattedValue, setFormattedValue] = useState(
     formatter.format(getRawNumber(value.toString()))
   );
 
+  let prefix: string;
+
+  if (currencySymbol) {
```
marthayaputra: can we put all this to `getCurrencySymbol` function instead with early return when condition fulfilled?

src/Utils/CurrencyCodes.tsx
ganiirsyadi: Removed it, change to try catch

src/@next/CurrencyInput/CurrencyInput.tsx
ganiirsyadi: Updated

src/@next/CurrencyInput/CurrencyInput.tsx
```diff
@@ -47,6 +55,36 @@ export const CurrencyInput = React.forwardRef<
...
   };
 
+  const getCurrencySymbol = (
+    locale: string,
+    currencyCode: string,
+    currencySymbol?: string
+  ) => {
+    if (currencySymbol) {
+      return currencySymbol;
+    } else if (currencyCode) {
```
marthayaputra: we don't need this else if section since we have inverted the if

ganiirsyadi: Updated

marthayaputra (approved):
LGTM

PR #751, opened 2023-02-16
feat: add next card

marthayaputra:


src/@next/Modal/Modal.tsx
```diff
@@ -15,20 +16,15 @@ import {
...
-};
-
 export type ModalProps = {
   isOpen?: boolean;
   header?: string;
   headerDescription?: string;
   children?: React.ReactNode;
   /** This action will be tied to primary button on footer */
-  primaryAction?: ModalAction;
+  primaryAction?: ComponentAction;
```
marthayaputra, line 25: Refactor to a common type

src/@next/Card/Card.tsx
```diff
@@ -0,0 +1,100 @@
...
+    subheading,
+    primaryAction,
+    secondaryAction,
+    actionsAlignment = 'right',
+    children,
+  }: CardProps,
+  ref
+) {
+  const headingMarkup = (
+    <Typography as="div" variant="body2">
```
ganiirsyadi, line 37: If I use `<FormattedMessage />` component here, the color will be overriden to #000. I am suggesting to add the color prop here.

Another question, is there any case that we need to change the typography for the heading? Let's say to use "body1" instead of "body2"

src/@next/Card/Card.tsx
```diff
@@ -0,0 +1,100 @@
...
+      {heading}
+    </Typography>
+  );
+
+  const subHeadingMarkup = (
+    <Typography as="div" variant="subtitle2">
+      {subheading}
```
ganiirsyadi, line 44: Same as above

src/@next/Card/Section.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Typography } from '../Typography';
+import { StyledCardSection } from './CardStyle';
+
+export const Section = ({ children }: { children: React.ReactNode }) => {
+  return (
+    <StyledCardSection className="section">
+      <Typography as="div" variant="body1">
+        {children}
```
ganiirsyadi, line 9: Same as above

src/@next/Card/Card.tsx
ganiirsyadi: Actually, I can use `formatMessage()` function to prevent the color overriding (since it return a string instead of a React Node).

marthayaputra: @ganiirsyadi I think we shouldn't design a common component with the specific use case here, need to investigate why FormatMessage overriding the color. Using prop will be hard to override

src/@next/Card/Card.tsx
marthayaputra: same as above

src/@next/Card/Card.tsx
marthayaputra: `Another question, is there any case that we need to change the typography for the heading? Let's say to use "body1" instead of "body2"`

It shouldn't, or else the heading will look different on each page

src/@next/Card/Section.tsx
marthayaputra: same comment as: https://github.com/glints-dev/glints-aries/pull/751#discussion_r1109413414

ganiirsyadi (approved):

PR #754, opened 2023-02-17
fix: inconsistent loading state and replacing with spinner

ganiirsyadi:


src/@next/IndexTable/IndexTable.tsx
```diff
@@ -8,14 +8,23 @@ import { Checkbox } from '../Checkbox';
 import { CheckboxProps } from '../Checkbox';
 import { Row } from './components/Row/Row';
 import { StyledIndexTable } from './IndexTableStyle';
+import { LoadingState } from './components/LoadingState';
+
+export interface CustomIndexTableProps
+  extends Omit<IndexTableProps, 'emptySearchTitle'> {
+  emptySearchTitle?: string;
```
ganiirsyadi, line 15: It should be optional because, now we can use emptyState prop to render empty table.

src/@next/IndexTable/IndexTable.tsx
```diff
@@ -37,8 +46,19 @@ const IndexTable = ({
       <PolarisIndexTable
         bulkActions={bulkActions}
         checkbox={renderCheckboxHeader}
-        itemCount={itemCount}
+        itemCount={loading ? 0 : itemCount}
         selectedItemsCount={selectedItemsCount}
+        emptySearchTitle={emptySearchTitle ?? ''}
+        emptyState={
```
ganiirsyadi, line 52: The strategy is to replace the emptyState when the loading prop is `true`. To make it works, I should set the itemCount = `0` too.

PR #755, opened 2023-02-17
feat: replace default loading state with spinner

ganiirsyadi:


src/@next/IndexTable/IndexTable.tsx
```diff
@@ -8,14 +8,23 @@ import { Checkbox } from '../Checkbox';
 import { CheckboxProps } from '../Checkbox';
 import { Row } from './components/Row/Row';
 import { StyledIndexTable } from './IndexTableStyle';
+import { LoadingState } from './components/LoadingState';
+
+export interface CustomIndexTableProps
+  extends Omit<IndexTableProps, 'emptySearchTitle'> {
```
marthayaputra: Any reason why don't we use the existing `emptySearchTitle `?

src/@next/IndexTable/components/LoadingState/LoadingStateStyle.ts
```diff
@@ -0,0 +1,26 @@
...
+  width: 100%;
+  height: 100%;
+  flex-direction: column;
+  gap: 16px;
+
+  p {
+    margin: 0;
+  }
+
+  @media ${Device.desktopS} {
```
marthayaputra: I think we should use Breakpoints from `@next`

src/@next/IndexTable/IndexTable.tsx
```diff
@@ -37,8 +46,20 @@ const IndexTable = ({
       <PolarisIndexTable
         bulkActions={bulkActions}
         checkbox={renderCheckboxHeader}
-        itemCount={itemCount}
+        itemCount={loading ? 0 : itemCount}
         selectedItemsCount={selectedItemsCount}
+        emptySearchTitle={emptySearchTitle ?? ''}
+        loading={false}
```
marthayaputra, line 48: I guess can remove this prop

src/@next/IndexTable/IndexTable.tsx
ganiirsyadi: Since we use emptyState there is no need to fill the emptySearchTitle anymore. It's still there but it's optional

src/@next/IndexTable/components/LoadingState/LoadingStateStyle.ts
ganiirsyadi: The table is a quite wide component, Device with width of 1024 can't show the table fully and thus I need the 1024 breakpoints. Can I add a new one or do I need to stick with 768?

src/@next/IndexTable/IndexTable.tsx
ganiirsyadi: Sorry I forgot to mention why I added this, since I implement the loading state myself, I need to make sure that the stock `loading` state won't appear. That's why I added this prop.

src/@next/IndexTable/IndexTable.tsx
marthayaputra: Oh but then we create a new one with data type `string` after omitting it 🤔 , if we don't need it, agree to omit it, but no need to create a new one in CustomIndexTableProps, also I think you need to re-export `CustomIndexTableProps` as `IndexTableProps` so the user of this component can refer to it

src/@next/IndexTable/components/LoadingState/LoadingStateStyle.ts
marthayaputra: @ganiirsyadi sure, but please discuss with Maeve, I think there will be resolution we can't see it fully, that expected, which a scrollbar will appear

ganiirsyadi: Yes a scrollbar will appear, but I need to make sure the spinner is at the center of the window (not at the center of the table, because the table width > screen width. 

Update:
It turned out, it is safe to remove the @media query and apply those styles directly. I will ask Maeve to re-test it to make sure it doesn't break on larger device.

marthayaputra: > Yes a scrollbar will appear, but I need to make sure the spinner is at the center of the window (not at the center of the table, because the table width > screen width

Hmm i think the loader should be in the center of the table, since it's a loading state for the table 🤔

ganiirsyadi: The best that I can do for now is to use breakpoint large to avoid unexpected behavior on Desktop screen when forcing the loading to be at the center of the screen.

src/@next/IndexTable/IndexTable.tsx
ganiirsyadi: I've updated it, please check if it match your expectation

marthayaputra (approved):

PR #759, opened 2023-02-21
feat: Implement clickable tag

ganiirsyadi:


ganiirsyadi:
I updated the storybook for the normal tag and regenerated the snapshots for normal tag. Previously it renders Tag with remove Icon, that should not be the case for normal Tag.

marthayaputra:
> I updated the storybook for the normal tag and regenerated the snapshots for normal tag. Previously it renders Tag with remove Icon, that should not be the case for normal Tag.

any reason removable tag screenshots were regenerated?

src/@next/Tag/Tag.tsx
```diff
@@ -51,7 +65,16 @@ export const Tag = React.forwardRef<HTMLDivElement, TagProps>(function Tag(
...
   return (
-    <TagStyle ref={ref} {...props} value={value}>
+    <TagStyle
+      ref={ref}
+      {...props}
+      value={value}
+      onClick={!disabled && onClick}
+      data-clickable={!!onClick}
+      role={!!onClick ? 'button' : undefined}
+      data-disabled={!!onClick && disabled}
```
marthayaputra: why do we need to check onClick too for disabled here?

src/@next/Tag/Tag.tsx
```diff
@@ -14,23 +14,37 @@ export interface TagProps extends React.HTMLAttributes<HTMLDivElement> {
...
+    onClick,
+    disabled,
+    ...props
+  }: TagProps,
   ref
 ) {
+  const handleTextColor = () => {
+    if (disabled) {
+      return Neutral.B85;
+    } else if (textColor) {
```
marthayaputra: there's no need else here if the first if is inverted, just an if will do

src/@next/Tag/Tag.tsx
ganiirsyadi: Hmm, yes, I should delete the onClick. I was thinking of someone passing `disabled` in non clickable tag. But I think that would not make any sense.

src/@next/Tag/Tag.tsx
ganiirsyadi: Fixed

src/@next/Tag/Tag.tsx
ganiirsyadi: Fixed

ganiirsyadi:
> > I updated the storybook for the normal tag and regenerated the snapshots for normal tag. Previously it renders Tag with remove Icon, that should not be the case for normal Tag.
> 
> any reason removable tag screenshots were regenerated?

I found that the text inside tag get shifted a bit to the top, thus I translate them 1px down so it placed at the center of the tag.

Before:
image

After:
image

marthayaputra:
> > > I updated the storybook for the normal tag and regenerated the snapshots for normal tag. Previously it renders Tag with remove Icon, that should not be the case for normal Tag.
> > 
> > 
> > any reason removable tag screenshots were regenerated?
> 
> I found that the text inside tag get shifted a bit to the top, thus I translate them 1px down so it placed at the center of the tag.
> 
> Before: image
> 
> After: image

Sorry looks like the after doesn't have any removable button?

ganiirsyadi:
> > > > I updated the storybook for the normal tag and regenerated the snapshots for normal tag. Previously it renders Tag with remove Icon, that should not be the case for normal Tag.
> > > 
> > > 
> > > any reason removable tag screenshots were regenerated?
> > 
> > 
> > I found that the text inside tag get shifted a bit to the top, thus I translate them 1px down so it placed at the center of the tag.
> > Before: image
> > After: image
> 
> Sorry looks like the after doesn't have any removable button?

Ignore the removable button, the point is on the text itself. 
Or you can just simply check from changes and use `swipe` on snapshots difference to get what I meant


https://user-images.githubusercontent.com/52077891/220536135-009fd350-485a-40c0-8106-4829e0a236b7.mov

marthayaputra:
> > > > > I updated the storybook for the normal tag and regenerated the snapshots for normal tag. Previously it renders Tag with remove Icon, that should not be the case for normal Tag.
> > > > 
> > > > 
> > > > any reason removable tag screenshots were regenerated?
> > > 
> > > 
> > > I found that the text inside tag get shifted a bit to the top, thus I translate them 1px down so it placed at the center of the tag.
> > > Before: image
> > > After: image
> > 
> > 
> > Sorry looks like the after doesn't have any removable button?
> 
> Ignore the removable button, the point is on the text itself. Or you can just simply check from changes and use `swipe` on snapshots difference to get what I meant

Have this changes been communicated with Maeve?

ganiirsyadi:
> Have this changes been communicated with Maeve?

Let me have a discussion with her

ganiirsyadi:
@marthayaputra I've discussed with Maeve, she is fine with my changes (shift the text down by 1px). It's QC approved.
https://glints.slack.com/archives/C018BD6GVS8/p1677045903420959?thread_ts=1676970121.693889&cid=C018BD6GVS8

marthayaputra (approved):

PR #760, opened 2023-02-22
feat: add next datatable

marthayaputra:


src/@next/Typography/Typography.tsx
```diff
@@ -45,7 +47,7 @@ export const Typography = ({
   }
 
   return (
-    <StyledTypography as={as} variant={variant} {...props}>
+    <StyledTypography style={style} as={as} variant={variant} {...props}>
```
marthayaputra, line 50: This change needed because seems like any attempt to override the css not allowed, we need to set `white-space: nowrap` for table cells

src/@next/Pagination/PaginationStyle.tsx
```diff
@@ -20,7 +20,7 @@ export const StyledPageButton = styled.button<PageButtonProps>`
   height: 32px;
   min-width: 32px;
   border: 0;
-  background: ${Neutral.B100};
+  background: transparent;
```
marthayaputra, line 23: Pagination button should not have a background color

src/@next/Typography/Typography.tsx
ninariccimarie: do we know why?
it's also not possible with `styled(Typography)` ?

src/@next/DataTable/TableHeader.tsx
```diff
@@ -0,0 +1,65 @@
...
+      sortDirection === 'ASCENDING' ? 'DESCENDING' : 'ASCENDING';
+    setSortDirection(direction);
+    onSort?.(direction);
+  };
+
+  const sortableCellMarkup = (
+    <Button
+      data-testid={`data-table-sort-button-${title}`}
+      icon={iconMarkup}
+      onClick={() => handleSortableClick()}
```
ninariccimarie, line 53:
```suggestion
      onClick={handleSortableClick}
```
can directly pass as this?

src/@next/DataTable/TableHeader.tsx
```diff
@@ -0,0 +1,65 @@
...
+  onSort?: (sortDirection: SortDirection) => void;
+}
+
+export const TableHeader = ({
+  sortDirection: sortDirectionProp,
+  title,
+  align = 'left',
+  onSort,
+}: TableHeaderProps) => {
+  const [sortDirection, setSortDirection] = useState(null);
```
ninariccimarie, line 21: why did we use null instead of sortDirection for it's initial value since we ask for a default direction?

src/@next/DataTable/TableCell.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export interface TableCellProps
+  extends React.TdHTMLAttributes<HTMLTableColElement> {
+  children: React.ReactNode;
+}
+
+export const TableCell = ({ align, children }: TableCellProps) => {
+  return (
+    <StyledTableCell data-text-align={align}>
+      <div>
```
ninariccimarie: do we still need the extra div?

src/@next/Typography/Typography.tsx
marthayaputra: @ninariccimarie , not sure if I did it wrongly, there's a weird `TS` error, when combining `as` and `variant` for me when I did that

src/@next/DataTable/TableHeader.tsx
marthayaputra: Hmm onClick naturally has a clickEvent argument, but in this case we don't need that, hence this

src/@next/DataTable/TableHeader.tsx
marthayaputra: because it'll be redundant with ` useEffect(() => {
    setSortDirection(sortDirectionProp);
  }, [sortDirectionProp]);`

src/@next/DataTable/TableCell.tsx
marthayaputra: yes, since anything can be inside the cell (not just text) we need this wrapper to align the content

marthayaputra: Oh I see what you meant, the Typography is already a div, will remove that

ninariccimarie (approved):
LGTM 🚀

PR #761, opened 2023-02-22
feat: add next link

ninariccimarie:


src/@next/Link/Link.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export const Link = ({
+  children,
+  external = false,
+  monochrome = false,
+  removeUnderline = false,
+  url,
+  ...props
+}: LinkProps) => {
+  const [isHovered, setIsHovered] = useState(false);
```
marthayaputra: any reason we need to define the hover state?

ninariccimarie: yes, I'm using it to define the color for the Icon: https://github.com/glints-dev/glints-aries/pull/761/files#diff-0e4ea2ddfa524b9a05d27f841fdf9909c75d4b8fd7feeafae440fe2bc899d8b4R33-R44

since we need to have them changed when it is hovered. do you have other ideas on a better way to handle it?

marthayaputra: I think we can do `&:hover svg { fill: '#color'}` since icon is the child element of the link

src/@next/Link/LinkStyle.ts
```diff
@@ -0,0 +1,34 @@
...
+  &:hover {
+    color: rgba(0, 86, 140, 0.9);
+  }
+
+  &:active {
+    color: ${Blue.S100};
+  }
+`;
+
+export const StyledIcon = styled(Icon)`
```
marthayaputra: I think we wouldn't need this too, can style them as the child of StyledLink

src/@next/Link/Link.tsx
ninariccimarie: ohhh that's better!

src/@next/Link/LinkStyle.ts
ninariccimarie: you're right! let me update it

src/@next/index.ts
```diff
@@ -38,6 +38,7 @@ export {
   IndexTableProps,
   useIndexResourceState,
 } from './IndexTable';
+export { Link } from './Link';
```
marthayaputra: oh, almost forgot, can we export LinkProps as well?

src/@next/Link/Link.tsx
ninariccimarie: addressed! please check the changes

marthayaputra (approved):

PR #764, opened 2023-02-24
feat: Spinner with label

ninariccimarie:


src/@next/Spinner/Spinner.tsx
```diff
@@ -1,25 +1,39 @@
...
+  ['large']: 'subtitle1',
+};
 
-export const Spinner = (props: SpinnerProps) => (
-  <StyledWrapper className="spinner-container">
-    <Icon name="ri-loader" {...props} />
-  </StyledWrapper>
-);
+export const Spinner = ({ label, size, ...props }: SpinnerProps) => {
+  const [iconSize, setIconSize] = useState(null);
```
marthayaputra: do we need this state? Since we don't change it locally

src/@next/Spinner/Spinner.tsx
```diff
@@ -1,25 +1,39 @@
...
+
+  useEffect(() => {
```
marthayaputra: Don't think we need an effect as well here

src/@next/Spinner/Spinner.tsx
ninariccimarie: you're right, let me remove them.

marthayaputra (approved):

chiahou (approved):
looks good to me. thanks

PR #766, opened 2023-02-24
Feature/react node card heading

ganiirsyadi:


src/@next/Card/Card.tsx
```diff
@@ -33,16 +36,20 @@ const CardComponent = React.forwardRef<HTMLDivElement, CardProps>(function Card(
...
+  const headingMarkup = PRIMITIVE_TYPES.includes(typeof heading) ? (
+    <Typography as="div" variant="subtitle1">
       {heading}
     </Typography>
+  ) : (
+    <StyledCustomHeader>{heading}</StyledCustomHeader>
   );
 
-  const subHeadingMarkup = (
+  const subHeadingMarkup = PRIMITIVE_TYPES.includes(typeof subheading) ? (
```
marthayaputra: don't think we need this condition anymore since the Typography is a `div` which can serve as a neutral wrapper

ganiirsyadi: The reason I make it conditional is because, when I delete it like this
```tsx
const headingMarkup = (
  <Typography as="div" variant="subtitle1">
    {heading}
  </Typography>
);

const subHeadingMarkup = (
  <Typography as="div" variant="subtitle2">
    {subheading}
  </Typography>
);
```
There are margins on top and bottom of the custom heading/subheading, resulting on this UI:
image

I am not sure why this happen

I created a new styled div to mitigate this issue by applying margin: 0,
```ts
export const StyledCustomHeader = styled.div`
  & > * {
    margin: 0;
  }
`;
```

ganiirsyadi: Here is my storybook component:
```tsx
const CustomHeadingSubHeadingTemplate: Story<CardProps> = args => {
  const heading = (
    <Typography variant="subtitle1" color={Red.B74}>
      Custom Heading
    </Typography>
  );

  const subheading = (
    <Typography variant="subtitle2" color={Red.B65}>
      Custom Sub Heading
    </Typography>
  );

  return (
    <Card {...args} heading={heading} subheading={subheading}>
      <Card.Section>This is a section</Card.Section>
    </Card>
  );
};
```

ganiirsyadi: Found the root cause:
image

ganiirsyadi: Do you think it is something that the user (who consume this component) need to handle themselves?

ganiirsyadi: If yes, I am good with removing the conditional (PRIMITIVE_TYPES)

marthayaputra: Looks like it's coming from the node passed to the `Card` which is `p`? I think since we can't predict what will be passed as heading, probably better for user to adjust the style while we can make sure the font style is what we want when wrapping them with Typography 🤔

ganiirsyadi: Ok if that so, I will delete this condition `PRIMITIVE_TYPES.includes(typeof subheading) ?`.

ganiirsyadi: An example to fix it on the user side,
```tsx
const CustomHeadingSubHeadingTemplate: Story<CardProps> = args => {
  const StyledTypography = styled(Typography)`
    margin: 0;
  `;

  const heading = (
    <StyledTypography variant="subtitle1" color={Red.B74}>
      Custom Heading
    </StyledTypography>
  );

  const subheading = (
    <StyledTypography variant="subtitle2" color={Red.B65}>
      Custom Sub Heading
    </StyledTypography>
  );

  return (
    <Card {...args} heading={heading} subheading={subheading}>
      <Card.Section>This is a section</Card.Section>
    </Card>
  );
};
```

ganiirsyadi: Removed

chiahou (approved):
qc passed

src/@next/Card/Card.tsx
```diff
@@ -10,12 +10,15 @@ import {
   StyledCardHeaderSection,
   StyledCardHeaderSectionHalf,
   StyledCardHeaderWrapper,
+  StyledCustomHeader,
 } from './CardStyle';
 import { Section } from './Section';
 
+const PRIMITIVE_TYPES = ['string', 'number'];
```
marthayaputra: need to remove unused var, maybe should set the lint to error for this 🤔

src/@next/Card/Card.tsx
```diff
@@ -10,12 +10,15 @@ import {
   StyledCardHeaderSection,
   StyledCardHeaderSectionHalf,
   StyledCardHeaderWrapper,
+  StyledCustomHeader,
```
marthayaputra: unused var

ganiirsyadi: Deleted

src/@next/Card/Card.tsx
ganiirsyadi: Deleted

marthayaputra (approved):

PR #771, opened 2023-02-28
feat: fix suffix won't get updated on prop change

ganiirsyadi:


src/@next/TextInput/TextInput.tsx
```diff
@@ -32,6 +32,10 @@ export const TextInput = React.forwardRef<HTMLInputElement, TextInputProps>(
       onChange?.(e);
     };
 
+    useEffect(() => {
```
ganiirsyadi, line 35: The suffix from prop only get accepted once at line 19 as a default value from useState. After that, any change from the suffix won't affect the actual rendered suffix because `suffix` prop that is passed to `<Input />` component is `suffixValue` not the one from the props.

ganiirsyadi: I added useEffect to handle the change of suffix prop that get passed to <TextInput />

ninariccimarie (approved):
LGTM 🚀 thanks for the quick fix

PR #773, opened 2023-03-01
feat: Basic Menu

ninariccimarie:


src/@next/Menu/Menu.tsx
```diff
@@ -0,0 +1,56 @@
...
+  id?: string;
+  onChange?: (selected: string[]) => void;
+  options?: Option[];
+  selected?: string[];
+  title?: string;
+}
+
+export const Menu = ({ id, onChange, options, selected, title }: MenuProps) => {
+  const handleClick = useCallback(
+    (value: string) => {
```
marthayaputra: why do we need useCallback here? 🤔

src/@next/Menu/Menu.tsx
```diff
@@ -0,0 +1,56 @@
+import React, { useCallback } from 'react';
+import { Typography } from '../Typography';
+import { Neutral } from '../utilities/colors';
+import { ListContainer, StyledMenu, TitleContainer } from './MenuStyle';
+
+interface Option {
+  active?: boolean;
```
marthayaputra: Hmm wouldn't this redundant with selected props? In what case user need to use active?

src/@next/Menu/Menu.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <TitleContainer>
+        <Typography variant="subtitle2" as="span" color={Neutral.B40}>
+          {title}
+        </Typography>
+      </TitleContainer>
+      {options.map(option => {
+        const { value, label, disabled, active, id } = option;
+        const isSelected = selected.includes(value) || active;
+        return (
+          <ListContainer key={id}>
```
marthayaputra, line 47: since id is optional, we need to add a fallback value for it

src/@next/Menu/Menu.tsx
```diff
@@ -0,0 +1,56 @@
...
+  disabled?: boolean;
+  id?: string;
+  label: React.ReactNode;
+  value: string;
+}
+export interface MenuProps {
+  id?: string;
+  onChange?: (selected: string[]) => void;
+  options?: Option[];
+  selected?: string[];
```
marthayaputra: Probably need to make it clear this will relate to `Option.value`, maybe we should rename this as `selectedValue` and add a description

marthayaputra (changes requested):

chiahou (changes requested):
It seems that the focus style doesn't work. However, I adjusted the focus style yesterday in order to align with the action list. Instead of using box-shadow, we can simply [set the border color to blue.](https://www.figma.com/file/87fhOdF8vk1Uiy20bCqS85/Sirius-Web---UI-Kit-3.0-%5BQ4-2022%5D?node-id=505%3A8061&t=ewaifkW0WpxEDcOU-4) Could you help to make this change?

src/@next/Menu/Menu.tsx
ninariccimarie: let me remove it as it's not needed here

src/@next/Menu/Menu.tsx
ninariccimarie: you're right! let me remove this too

src/@next/Menu/Menu.tsx
ninariccimarie: thanks for the catch! added fallback for id 👍

src/@next/Menu/Menu.tsx
ninariccimarie: agree! renamed and added definition

marthayaputra (approved):

chiahou (approved):
ui looks good to me. thanks!

PR #776, opened 2023-03-03
feat: fix card padding styling

ganiirsyadi:


src/@next/Card/CardStyle.ts
```diff
@@ -34,18 +34,38 @@ export const StyledCardHeaderSectionHalf = styled.div`
 `;
 
 export const StyledCardContentWrapper = styled.div`
-  padding-left: ${space4};
+  padding-bottom: ${space4};
```
ganiirsyadi, line 37: I remove padding-left and right on the wrapper. The reason is, it should't be an issue if we set the root Section to have padding-left-right = 24px, since the sub section padding is overriden

src/@next/Card/CardStyle.ts
```diff
@@ -34,18 +34,38 @@ export const StyledCardHeaderSectionHalf = styled.div`
...
+
+  @media (max-width: ${Breakpoints.large}) {
+    padding-bottom: 0px;
```
ganiirsyadi, line 40: For smaller screen, there is no need to add extra padding-bottom since the Section itself has padding-bottom = Header padding-top = 12px. So it's even for the overall padding top-bottom of the card

marthayaputra (approved):

chiahou (approved):
looks good to me. thanks!

PR #777, opened 2023-03-03
feat: pass other div props to Modal Container

ganiirsyadi:


src/@next/Modal/Modal.tsx
```diff
@@ -53,6 +53,7 @@ export const Modal = React.forwardRef<HTMLDivElement, ModalProps>(
       closeOnClickOutside,
       onClose,
       onBack,
+      ...props
```
marthayaputra, line 56: Don't you need to extend the ModalProps component with DivElement? Or did forwardRef took care of it?

ganiirsyadi: I don't think so, i will add it

ganiirsyadi: Added

src/@next/Modal/Modal.tsx
```diff
@@ -16,7 +16,7 @@ import {
   StyledModalBackButton,
 } from './ModalStyle';
 
-export type ModalProps = {
+export interface ModalProps extends React.HTMLAttributes<HTMLDivElement> {
```
ganiirsyadi, line 19: Added

marthayaputra (approved):

PR #782, opened 2023-03-08
feat: fix input value won't change

ganiirsyadi:


src/@next/CurrencyInput/CurrencyInput.tsx
```diff
@@ -35,25 +35,31 @@ export const CurrencyInput = React.forwardRef<
...
+  const formatter = useMemo(
+    () => new Intl.NumberFormat(localeValue),
+    [localeValue]
+  );
 
-  const getRawNumber = (value: string) => {
-    if (typeof value === 'number') {
-      return value;
-    }
+  const getRawNumber = useCallback(
```
ganiirsyadi, line 43: The linter gives a warning if I'm not wrapping it with useCallback because the function get called inside a useEffect

src/@next/CurrencyInput/CurrencyInput.tsx
```diff
@@ -35,25 +35,31 @@ export const CurrencyInput = React.forwardRef<
     console.warn(`Locale value of ${locale} is unsupported, "en" will be used`);
   }
 
-  const formatter = new Intl.NumberFormat(localeValue);
+  const formatter = useMemo(
```
ganiirsyadi, line 38: It becomes a chaining effect, it gives another warning that I should wrap this with useMemo because it is used inside a useCallback

PR #788, opened 2023-03-10
feat: select

ninariccimarie:


src/@next/Menu/MenuStyle.ts
```diff
@@ -71,3 +84,9 @@ export const TitleContainer = styled.div`
 export const ListContainer = styled.div`
   padding: 0 ${space8};
 `;
+
+export const StyledSections = styled.ul`
+  list-style: none;
+  padding: 0;
+  margin: ${space8} 0;
```
chiahou, line 92: vertical margins of 8px should be removed from the ListContainer

src/@next/Menu/MenuStyle.ts
chiahou, line 1: a bottom margin of 4px should be added to TitleContainer

src/@next/Select/Select.stories.tsx
```diff
@@ -0,0 +1,265 @@
...
+  {}
+);
+
+NonSearchableSingleSelect.args = {};
+
+const NonSearchableAllowMultiTemplate: Story<SelectProps> = args => {
+  return <NonSearchableMultiSelect options={options} {...args} allowMultiple />;
+};
+
+export const NonSearchableAllowMulti = NonSearchableAllowMultiTemplate.bind({});
```
chiahou: not sure why the color of placeholder is incorrect and doesn't align with other placeholders. it should be #666 instead of #2d2d2d

ninariccimarie: the cause of this one might've been the mix up with the links I shared.

chiahou (approved):
qc passed, thank you @ninariccimarie

src/@next/Menu/MenuOption.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <ListContainer>
+      <li
+        aria-disabled={disabled}
+        data-active={isSelected}
+        data-multiple={allowMultiple}
+        value={value}
+        onClick={() => handleClick(value)}
+      >
+        {renderLabelContent({ isSelected, disabled })}
```
marthayaputra: can we render all the children at once when we know multiple is allowed? Doing the `if` statement inside `renderLabelContent` seems inefficient, as it will have to do it n times items

marthayaputra: I think it's better to render different set of items on `Menu` component based on `allowMultiple `

src/@next/Menu/menuStoryHelper/PopoverMenu.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const [popoverActive, setPopoverActive] = useState(true);
+  const [selectedOptions, setSelectedOptions] = useState([]);
+
+  const togglePopoverActive = useCallback(
+    () => setPopoverActive(popoverActive => !popoverActive),
+    []
+  );
+
+  const handleSelect = ({ value }: { value: string }) => {
+    if (allowMultiple) {
```
marthayaputra: can we invert the if to reduce complexity here?

src/@next/Select/components/Activator/ActivatorSelect.tsx
```diff
@@ -0,0 +1,117 @@
...
+  placeholder,
+  value,
+  onClick,
+  onRemoveTag,
+  width,
+  ...props
+}: ActivatorSelectProps) => {
+  const activatorRef = useRef(null);
+  const hasValue = !!value;
+  const activatorContext = useSelectActivator() as ActivatorSelectContextProps;
```
marthayaputra: is it possible to avoid as here? Probably can use generic or [conditional type](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html)

src/@next/Select/components/Activator/ActivatorSelect.tsx
```diff
@@ -0,0 +1,117 @@
...
+    disabled,
+    onSelectClick,
+    hasError,
+    setWidth,
+    selectedValues,
+  } = activatorContext;
+
+  useEffect(() => {
+    setWidth(activatorRef.current.getBoundingClientRect().width);
+  }, [activatorRef, setWidth]);
```
marthayaputra: why do we need these dependencies? It seems like they're not used inside effects 🤔

src/@next/Select/selectStoryHelper/NonSearchableMultiSelect.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  const handleSelect = ({ value }: { value: string }) => {
+    if (selected.includes(value)) {
+      setSelected(selected.filter(option => option !== value));
+    } else {
+      setSelected([...selected, value]);
+    }
+  };
+
+  const removeTag = useCallback(
```
marthayaputra: why do we need useCallback here? 🤔

src/@next/Select/selectStoryHelper/NonSearchableMultiSelect.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export const NonSearchableMultiSelect = ({
+  options,
+  disabled,
+  ...args
+}: SelectProps) => {
+  const [selected, setSelected] = useState([]);
+
+  const handleSelect = ({ value }: { value: string }) => {
+    if (selected.includes(value)) {
```
marthayaputra: the if can be inverted

src/@next/Select/selectStoryHelper/Searchable.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    const filterRegex = new RegExp(value, 'i');
+    const filterOptions = options.filter((option: Option) =>
+      (option.label as string).match(filterRegex)
+    );
+    setOptions(filterOptions);
+  };
+
+  const handleSelect = ({ value }: { value: string }) => {
+    if (selectedOptions.includes(value)) {
```
marthayaputra: if can be inverted

src/@next/Select/selectStoryHelper/Searchable.tsx
```diff
@@ -0,0 +1,90 @@
...
+    }
+  };
+
+  const removeTag = useCallback(
+    tag => () => {
+      const options = [...selectedOptions];
+      options.splice(options.indexOf(tag), 1);
+      setSelectedOptions(options);
+    },
+    [selectedOptions]
```
marthayaputra: not sure if we need useCallback here

src/@next/Tag/Tag.tsx
```diff
@@ -12,7 +12,7 @@ import {
 export interface TagProps extends React.HTMLAttributes<HTMLDivElement> {
   children?: React.ReactNode;
   value?: string;
-  onRemove?: (() => void) | null;
+  onRemove?: (() => void) | null | void;
```
marthayaputra, line 15: shouldn't this be `onRemove?: () => void | null` ? or is it accepting function as an argument?

src/@next/TextInput/TextInput.tsx
```diff
@@ -5,11 +5,12 @@ import { Input, InputProps } from '../Input/Input';
 export type TextInputProps = Omit<InputProps, 'type' | 'onChange'> & {
   canClear?: boolean;
   onChange?: (value: string) => void;
+  hasError?: boolean;
 };
 
 export const TextInput = React.forwardRef<HTMLInputElement, TextInputProps>(
   function TextInput(
-    { canClear, suffix, value, onChange, ...props }: TextInputProps,
+    { canClear, suffix, value, onChange, hasError, ...props }: TextInputProps,
```
marthayaputra: I think InputProps already has `error` prop can we just use that?

marthayaputra (changes requested):

src/@next/Menu/menuStoryHelper/PopoverMenu.tsx
ninariccimarie: @marthayaputra what do you mean by invert?

src/@next/Select/components/Activator/ActivatorSelect.tsx
ninariccimarie: I'm not sure if we can do either? since we are using 2 different interfaces. 
I think this is the reason why typescript is unable to identify which one I'm currently using: https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#unions-with-common-fields

src/@next/Select/components/Activator/ActivatorSelect.tsx
ninariccimarie: not sure what you mean, they're both used in line 40?

src/@next/Tag/Tag.tsx
ninariccimarie: yes, it is accepting function

src/@next/Select/selectStoryHelper/NonSearchableMultiSelect.tsx
ninariccimarie: looks like we don't really need to use here

src/@next/Select/selectStoryHelper/Searchable.tsx
ninariccimarie: yes, you're right

src/@next/Select/components/Activator/ActivatorSelect.tsx
marthayaputra: Hmm probably something like this? https://ordina-jworks.github.io/architecture/2021/02/12/react-generic-context.html

But upon looking at the ActivatorProvider type I'm more concern on how different type they are but living in the same provider, I'm leaning forward to actually separate them into different providers altogether

src/@next/Select/components/Activator/ActivatorSelect.tsx
marthayaputra: Oh right my bad, didn't see the setWidth. I guess it's strange because usually ref.current is used to escape the dependencies, does the exhaustive depth linter give you a warning if you remove activatorRef?

src/@next/Select/components/Activator/ActivatorSelect.tsx
ninariccimarie: thanks for the insights, Putra. I agree with separating them. I've addressed the separation

src/@next/Menu/menuStoryHelper/PopoverMenu.tsx
ninariccimarie: I've addressed the inversions. please check

src/@next/Select/Select.tsx
```diff
@@ -9,41 +9,57 @@ import {
...
+  width: string;
 }
 
 export const Select = ({
-  activator,
   allowMultiple = false,
   disabled,
   hasError,
   helpText,
+  inputValue,
```
marthayaputra: maybe we can group searchable related into one prop?

marthayaputra (approved):
LGTM

PR #792, opened 2023-03-13
feat: combobox option list scrollable

ninariccimarie:


src/@next/Combobox/components/OptionList/OptionListStyle.ts
```diff
@@ -5,11 +5,14 @@ import { space4, space8 } from '../../../utilities/spacing';
...
 interface OptionListContainerProps {
   textInputWidth?: number;
+  height?: string;
 }
 
 export const OptionListContainer = styled.div<OptionListContainerProps>`
   width: ${props => props.textInputWidth}px;
+  height: ${props => props.height};
   padding: ${space8} 0;
+  overflow-y: auto;
```
marthayaputra: hmmm I think Popover has scrollable style, just that I forgot to include it, can we manage that there? Or else will  need to do for every popoever children 🤔 
image

ninariccimarie: @marthayaputra I see! thanks! let me check that 👀

ninariccimarie: @marthayaputra I've applied your suggestion

src/@next/Combobox/comboboxStoryHelper/ComboboxMultuSelect.tsx
ninariccimarie, line 1: just moved this from the `Combobox.stories.tsx` file. no need to correct the file name since we are combining this with Select.

chiahou (approved):
qc passed, thanks!

skurnia001 (approved):
LGTM

PR #805, opened 2023-03-31
Feature: IndexTable checkbox hover state

ninariccimarie:


src/@next/Checkbox/Checkbox.tsx
```diff
@@ -76,10 +80,14 @@ export const Checkbox = React.forwardRef<HTMLDivElement, CheckboxProps>(
         aria-disabled={disabled}
         data-error={hasError}
         data-spacing={isPadded}
+        data-in-table={inTable}
       >
-        <StyledRow>
+        <StyledRow data-in-table={inTable}>
```
marthayaputra: I'm not sure if this sustainable when we have another need to put checkbox in another control, wouldn't it better to put it on IndexTable's checkbox?

src/@next/Checkbox/Checkbox.tsx
```diff
@@ -40,6 +42,8 @@ export const Checkbox = React.forwardRef<HTMLDivElement, CheckboxProps>(
       hasError,
       helpText,
       isPadded = true,
+      isHovered = false,
```
marthayaputra: Why do we need this? 🤔

marthayaputra (changes requested):

src/@next/Checkbox/Checkbox.tsx
ninariccimarie: you're right, I should move this to IndexTable instead

src/@next/Checkbox/Checkbox.tsx
ninariccimarie: it's to trigger hover when Checkbox table cell is hovered. Basically, it widens the area of point to trigger Checkbox hover. it's one of the requirements for the hover state

marthayaputra: Oh ic, can't we rely on native html hover behavior?

ninariccimarie: we can't trigger checkbox hover with just html when the table cell is hovered though?

marthayaputra (approved):
LGTM

PR #811, opened 2023-04-06
Feature: Plain and Destructive Plain buttons

ninariccimarie:


src/@next/Button/DestructivePlainButtonStyle.tsx
```diff
@@ -0,0 +1,51 @@
+import styled from 'styled-components';
```
marthayaputra: Does the file need to be tsx? 🤔

src/@next/Button/PlainButtonStyle.tsx
```diff
@@ -0,0 +1,51 @@
+import styled from 'styled-components';
```
marthayaputra: same, does it need to be tsx?

src/@next/Button/DestructivePlainButtonStyle.tsx
ninariccimarie: ah. no they don't let me update that

marthayaputra (approved):

PR #812, opened 2023-04-11
Feature: Skeleton base

ninariccimarie:


src/@next/Skeleton/components/SkeletonText/SkeletonText.tsx
```diff
@@ -0,0 +1,40 @@
...
+      ? Array.from({ length: rows }, () => variant)
+      : variant;
+
+  const renderTexts = () => {
+    return [...Array(rows)].map((_, index) => {
+      return (
+        <StyledSkeletonText
+          key={`skeleton-text-${index}`}
+          width={widthArray[index]}
+          variant={variantArray[index]}
```
marthayaputra: what will happen if the no of rows != label variant?

src/@next/Skeleton/components/SkeletonImage/SkeletonImageCircle.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+import { StyledSkeletonImageCircle } from './SkeletonImageStyle';
+
+export interface SkeletonImageCircleProps {
+  size?: string;
+}
+
+export const SkeletonImageCircle = ({ size }: SkeletonImageCircleProps) => {
```
marthayaputra: should we have a default size here?

src/@next/Skeleton/components/SkeletonImage/SkeletonImageSquare.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+import { StyledSkeletonImageSquare } from './SkeletonImageStyle';
+
+export interface SkeletonImageSquareProps {
+  height: string;
+  width: string;
+}
+
+export const SkeletonImageSquare = ({
+  height,
```
marthayaputra: should we have default width and height?

marthayaputra (changes requested):

src/@next/Skeleton/components/SkeletonText/SkeletonText.tsx
ninariccimarie: it won't render the row. we can ask for a default for this one as well? for the width, it defaults to 100%.

src/@next/Skeleton/components/SkeletonImage/SkeletonImageCircle.tsx
ninariccimarie: agree! asked Rebecca for a default. this will be at 48px

src/@next/Skeleton/components/SkeletonImage/SkeletonImageSquare.tsx
ninariccimarie: agree! asked Rebecca for a default. this will have default height=80px and width=80px

src/@next/Skeleton/components/SkeletonText/SkeletonText.tsx
```diff
@@ -0,0 +1,41 @@
...
+      ? Array.from({ length: rows }, () => variant)
+      : variant;
+
+  const renderTexts = () => {
+    return [...Array(rows)].map((_, index) => {
+      return (
+        <StyledSkeletonText
+          key={`skeleton-text-${index}`}
+          width={widthArray[index]}
+          variant={variantArray[index] || defaultVariant}
```
marthayaputra: won't this throw an exception if the variant array size < row? Also true for width array, maybe we can make rows prop type as this? 

`rows: { width: string, variant: Variant}[]` ?

Then we can have a default value for width and variant if row is define but not completely

ninariccimarie: no it won't. the value will be undefined so for width, it won't apply any in the styles so it's using 100% while for variant, it'll use defaultVariant value

ninariccimarie: I'll follow the suggested type for rows above since it'll be clearer as well to identify the width and variant for each row. thanks!

marthayaputra (approved):
LGTM

PR #814, opened 2023-04-27
feat: searchable single select

ninariccimarie:


chiahou (commented):
@ninariccimarie Can we hide the ClearButton when the input box is focused regardless of whether an option has been selected or not? The clear button should be used to clear the selected option instead of letters, since users are most likely to press the delete or backspace key to remove letters during typing. By doing so we can avoid confusion and also to avoid the bug that can be reproduced as follows:
1. Select "Cambodia"
2. Click the input box and type "Ma"
3. Click the clear button
4. Click the input box again
5. Expected result: See the whole list/ Actual result: Only the options contain "Ma" appear

src/@next/Select/components/SearchableSelectInput/SearchableSelectInputStyle.ts
```diff
@@ -0,0 +1,83 @@
...
+  grid-column: 1 / 1;
+  grid-row: 1 / 1;
+
+  font-family: ${NotoSans}, sans-serif;
+  font-style: normal;
+  font-weight: 400;
+  font-size: 16px;
+  line-height: 150%;
+
+  color: ${Neutral.B18};
```
chiahou, line 69: Can we turn this color to the one for placeholder ${Neutral.B40} when the input box is focused before text is entered? To be more precise:
1. An option is selected, we see the selected value in B18
2. The input box is focused, we see the selected value in B40
3. Users start typing, the selected value is replaced by the letters users type in

chiahou (approved):
all good. qc passed

src/@next/Select/Select.tsx
```diff
@@ -61,21 +66,53 @@ export const Select = ({
 }: SelectProps) => {
   const [popoverActive, setPopoverActive] = useState(false);
   const [optionListHeight, setOptionListHeight] = useState('');
+  const [menuOptions, setMenuOptions] = useState(options);
+  const [inputValue, setInputValue] = useState(searchableProps?.inputValue);
```
marthayaputra: same as above

src/@next/Select/Select.tsx
```diff
@@ -61,21 +66,53 @@ export const Select = ({
...
+
+  const [searchableSelectState, setSearchableSelectState] =
+    useState<SearchableSelectState>({
+      showSelected: false,
+      showInput: true,
+      showPlaceholder: true,
+    });
+
+  const updateSearchableSelectState = useCallback(
```
marthayaputra: any reason for useCallback, same as others

src/@next/Select/Select.tsx
```diff
@@ -111,17 +157,18 @@ export const Select = ({
   }, [listHeight, scrollable]);
 
   const activator = () => {
-    if (searchableProps) {
-      const { inputValue, onInputChange } = searchableProps;
+    if (searchable || searchableProps) {
       return (
         <ActivatorTextInputContext.Provider
```
marthayaputra: why do we still need providers here when we know what is the activator based on searchableProp?

src/@next/Select/Select.tsx
```diff
@@ -134,47 +181,53 @@ export const Select = ({
...
+            )}
+          </ActivatorWrapper>
+        }
+        onClose={handleClose}
+        autofocusTarget="none"
+        preventFocusOnClose
+        fullWidth
+      >
+        <Popover.Pane height={optionListHeight}>
+          <ActivatorTextInputContext.Provider
```
marthayaputra: Why do we need text provider here but not select? 🤔

src/@next/Select/Select.tsx
```diff
@@ -43,24 +42,54 @@ export interface SelectProps {
...
-  scrollable,
+  scrollable = false,
   sections,
   selectedValues,
   width,
 }: SelectProps) => {
   const [popoverActive, setPopoverActive] = useState(false);
   const [optionListHeight, setOptionListHeight] = useState('');
+  const [menuOptions, setMenuOptions] = useState(options);
+  const [inputValue, setInputValue] = useState(
```
marthayaputra, line 67: will this work if user programatically changed the value?

src/@next/Select/Select.tsx
```diff
@@ -43,24 +42,54 @@ export interface SelectProps {
...
+      showSelected: false,
+      showInput: true,
+      showPlaceholder: true,
+    });
+
+  const updateSearchableSelectState = (newState: SearchableSelectState) => {
+    setSearchableSelectState(newState);
+  };
+
+  const updateInputValue = useCallback(
```
marthayaputra: can we confirm if we still need the useCallback? Same for others

src/@next/Select/Select.tsx
ninariccimarie: it won't work but we'll be removing this in a separate PR after this one. to handle the input state internally. currently I just added a TODO comment beside the `searchableProps` prop and we'll help to check and replace if it's already been used in employers.

the current way for searchable multi select to update input value is to also pass `searchableProps?.onInputChange`. we also already have it on the storybook code docs

src/@next/Select/Select.tsx
ninariccimarie: thanks for the catch. this should be removable as well

marthayaputra (approved):

PR #816, opened 2023-05-02
chore: wraps open and close functions in useCallback hook

Afsal102:
- current these function reference changes everytime the component renders /re-renders which causes infinite looping issues when used in side effects(useEffect), wrapping it in useCallback fixes the issue of the reference been changed rapidly


ninariccimarie (approved):
LGTM

PR #819, opened 2023-05-05
feat: Badge attention color and Button with Badge

ninariccimarie:


src/@next/Badge/Badge.tsx
```diff
@@ -1,22 +1,23 @@
...
 import { Typography } from '../Typography';
-import { Neutral } from '../utilities/colors';
-import { BadgeStyle } from './BadgeStyle';
+import { BadgeStyle, badgeTextColor } from './BadgeStyle';
 
 const badgeStatusVariant = [
   'neutral',
   'success',
-  'informational',
+  'information',
```
ninariccimarie, line 8: need to update Badges that used these statuses in product services

chiahou (approved):
all good, thanks

src/@next/Button/ButtonStoriesSyle.ts
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components';
+import { Badge } from '../Badge';
+
+export const StyledBadge = styled(Badge)`
```
marthayaputra: Do we leave the creation of floating badge to the product squad? 🤔

ninariccimarie: yes @marthayaputra ? don't we want more flexibility?

marthayaputra: Hmm because this is part of Button stories, one might assume this is part of the buttons? But from what  I see here it's not supposed to be one

ninariccimarie: as discussed in slack, we'll create a separate page for the floating badge

marthayaputra (approved):

PR #826, opened 2023-05-15
indexTable bulkAction change visual test to assertion

michac789:


test/e2e/indexTable/indexTable.spec.ts
```diff
@@ -24,26 +24,65 @@ test('IndexTable - bulk action', async ({ page }) => {
   const indexTablePage = new IndexTablePage(page);
   await indexTablePage.goto();
 
+  let isCheckboxCorrect: boolean, tempText: string | null;
```
ninariccimarie: can we rename this to `isCheckboxValuesCorrect` and `selectAllTableHeadText` so it's easier to know what they are from the variable names.

test/e2e/indexTable/indexTable.spec.ts
```diff
@@ -24,26 +24,65 @@ test('IndexTable - bulk action', async ({ page }) => {
...
+  isCheckboxCorrect = await indexTablePage.verifyCheckboxValues(
+    new Array(5).fill(false)
+  );
+  expect(isCheckboxCorrect).toBe(true);
+
   await indexTablePage.checkboxes.first().click();
```
ninariccimarie: can you create a separate locator for the first checkbox and name it as `selectAllCheckbox` maybe so we understand what the first checkbox is. can you also assert for when it's `indeterminate` state other than the `checked` and `unchecked` states. we need this since we removed the screenshots and this is the only checkbox that has this state in this component.

test/e2e/indexTable/indexTable.spec.ts
```diff
@@ -24,26 +24,65 @@ test('IndexTable - bulk action', async ({ page }) => {
...
 
-  await expect(indexTablePage.canvas).toHaveScreenshot(
-    'indextable-bulk-action-selected-row.png'
-  );
+  isCheckboxCorrect = await indexTablePage.verifyCheckboxValues([
+    false,
+    true,
+    true,
+    true,
+    true,
```
ninariccimarie, line 60: i'm expecting the last checkbox should have been false as well. since previously this was checked, clicking it should've updated the checked state to false

ninariccimarie: can you double check this as well it seems the assertion is false positive and could affect the other assertions using the same function. the `selectAllTableHeadText` assertion below is correct which is why this seems to be false positive

test/e2e/indexTable/indexTable.spec.ts
```diff
@@ -24,26 +24,65 @@ test('IndexTable - bulk action', async ({ page }) => {
...
+  expect(tempText).toContain('Reject');
+  expect(tempText).toContain('Move to');
 
   await indexTablePage.moveToAction.click();
   await expect(indexTablePage.actionList).toBeVisible();
 
-  await expect(indexTablePage.canvas).toHaveScreenshot(
-    'indextable-bulk-action-action-list.png'
-  );
+  isCheckboxCorrect = await indexTablePage.verifyCheckboxValues([
```
ninariccimarie: i don't think you need lines 71-85 but verify there are 4 options from the bulk action list

test/e2e/indexTable/indexTable.spec.ts
```diff
@@ -24,26 +24,65 @@ test('IndexTable - bulk action', async ({ page }) => {
   const indexTablePage = new IndexTablePage(page);
   await indexTablePage.goto();
 
+  let isCheckboxCorrect: boolean, tempText: string | null;
+  isCheckboxCorrect = await indexTablePage.verifyCheckboxValues(
+    new Array(5).fill(false)
+  );
+  expect(isCheckboxCorrect).toBe(true);
```
ninariccimarie: can you assert that the bulk action is not visible initially?

ninariccimarie (changes requested):

test/e2e/indexTable/indexTable.spec.ts
michac789: sure!

test/e2e/indexTable/indexTable.spec.ts
michac789: sure!

test/e2e/indexTable/indexTable.spec.ts
michac789: > i'm expecting the last checkbox should have been false as well. since previously this was checked, clicking it should've updated the checked state to false

Yes, the last checkbox is `false`, which is on line 53. I also thought that the order should be based on the DOM, but not sure why it turns out that the order does not follow the DOM (pretty sure it is reversed there).
image
In that picture, the very first checkbox (which shows 3 selected candidates) is `checked` (so it should be `true` as indicated on line 57), the difference is only on the input-id (so there is a minus sign there instead of tick), maybe I could also have a test for that hmm.

Let me try to check for a moment and update you again.

test/e2e/indexTable/indexTable.spec.ts
michac789: okay! Is verifying the 4 texts (assessment, interviewing, offered, hired) enough?

michac789: I did this for now, let me know when you review it again if it is not sufficient

test/e2e/indexTable/indexTable.spec.ts
michac789: checked and fixed

test/e2e/indexTable/indexTable.spec.ts
michac789: yes! what I am not sure is how to check for `indeterminate` state... when I saw the checkbox by inspecting (and also see the accessibility tab), it seems the checkbox can only be checked or not checked, not sure if there are some specific properties that I should check to consider that as `intedeterminate`

ninariccimarie: can you try something like this: https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-have-attribute
```ts
// indeterminate checkbox
await expect(indexTablePage. selectAllCheckbox).toHaveAttribute('aria-checked', 'mixed');
```

ref: https://github.com/glints-dev/glints-aries/blob/v4/src/%40next/Checkbox/Checkbox.tsx#L59

ninariccimarie (approved):
LGTM 🚀 thanks!
it seems the base branch is out of date. Do update the base branch and rebase before merging

PR #828, opened 2023-05-17
Carousel

ninariccimarie:


src/@next/Carousel/CarouselStyle.ts
```diff
@@ -0,0 +1,144 @@
...
+  ${props => indicatorPositionStyleMapping[props.indicatorPosition]}
+`;
+
+const indicatorTypeStyleMapping = {
+  ['dot']: ``,
+  ['line']: `
+    width: 16px;
+    height: 6px;
+    border-radius: ${borderRadius16};
+    background: ${Neutral.B40};
```
chiahou, line 130: Hi @ninariccimarie There should be an opacity of 0.5 applied to the background.
image

src/@next/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,150 @@
...
+      setActiveIndex(index);
+    },
+    [childrenLength]
+  );
+
+  const handlePreviousClick = () => {
+    updateActiveIndex(activeIndex - 1);
+  };
+
+  const handleNextClick = () => {
```
chiahou, line 76: @ninariccimarie I've noticed that when the slide transitions from the last to the first image, or vice versa, it goes through all the images in between instead of immediately jumping to the other end. Could you help to improve the transition logic between the first and last images?

src/@next/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,150 @@
...
+    <CarouselContainer
+      aspectRatio={
+        aspectRatio ? `${aspectRatio?.width}/${aspectRatio?.height}` : null
+      }
+      height={height}
+      width={width}
+      onMouseEnter={handleMouseEnter}
+      onMouseLeave={handleMouseLeave}
+    >
+      <SlideContainer
```
chiahou, line 112: @ninariccimarie the slider doesn't seem to be responsive [as desired](https://nimb.ws/BmPxwt) with the indicators staying in the middle and left/right arrow remaining fixed on the two ends

src/@next/Carousel/Carousel.tsx
ninariccimarie: as discussed, we will work on this for improvements

chiahou (approved):
qc passed, thanks

src/@next/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,150 @@
...
+        ref={slideRef}
+        aspectRatio={
+          aspectRatio ? `${aspectRatio?.width}/${aspectRatio?.height}` : null
+        }
+        height={height}
+        width={width}
+      >
+        <ItemsContainer activeIndex={activeIndex}>{children}</ItemsContainer>
```
marthayaputra: shouldn't this be `childrenList`?

src/@next/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,150 @@
...
+  width,
+  indicatorPosition = 'outer',
+  indicatorType = 'line',
+}: CarouselProps) => {
+  const slideRef = useRef(null);
+  const [activeIndex, setActiveIndex] = useState(0);
+  const [hideArrows, setHideArrows] = useState(autoRotate);
+
+  const childrenList = React.Children.toArray(children).filter(child =>
+    React.isValidElement(child)
```
marthayaputra, line 51: Do we only allow images here? If so, is it possible to create a stricter filter?

ninariccimarie: @marthayaputra we allow any

marthayaputra (approved):

PR #831, opened 2023-05-19
Add side sheet component

michac789:
1st commit:
Add simple SideSheet component

Update (what's changing) from 2nd commit for SideSheet component:
- Scrolling > make sure contents behind are not scrollable, added stories to demonstrate functionality
- Backdrop > when clicked trigger the onClose callback
- Storybook > code now appear in 'docs' section, added 2 more samples including the one with tab (based on Figma)
- Code > renamed left and right button to basic and primary as requested
- Code > reorder imports (import from external libraries are on top, from local files are below ordered alphabetically)
- Styling > remove unneeded hover effect
- Styling > fixed bad styling in small screen (typo: used 100vh instead of 100vw for width on small screen)
- Styling > on large screen side sheet will be placed on the right
- Styling > added transition animation (fade in and fade out for 0.2 seconds)
- Styling > removed margin for content (children of this component) *(this is needed to allow tabs creation based on Figma)

Update (3rd commit):
Updated the animation. Now:
- When the open sidesheet button is clicked, the backdrop will fade for 0.2sec from opacity 0 to 1; the sidesheet itself will also fade with the backdrop and move from transformX(100%) to the intended position
- Animation in reverse when the close button is pressed or the backdrop is pressed: sidesheet will move to the right for 0.2seconds, backdrop and sidesheet will also fade out from opacity 1 to 0 for 0.2seconds

Update (4th commit):
- Stylings Modification Based On Review

Update (5th commit):
- Added test cases (visual)


src/@next/SideSheet/SideSheet.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export const Interactive = Template.bind({});
+
+Interactive.args = {
+  header: 'Title',
+  headerDescription: 'Description',
+  children: 'Content',
+  showBackButton: true,
+  showCloseButton: true,
+};
```
michac789: not really needed, will delete this or modify (I'm aware there are some irrelevant args, bcoz I copy pasted it earlier from Modal stories implementation

src/@next/SideSheet/SideSheet.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { SideSheetProps } from './SideSheet';
+import { SideSheet } from './SideSheet';
+import { Typography } from '../Typography';
+import { BaseContainer } from '../../Layout/GlintsContainer/GlintsContainer';
+
+export default {
+  title: '@next/SideSheet',
+  component: SideSheet,
+  decorators: [
+    Story => <BaseContainer>{Story()}</BaseContainer>,
```
michac789, line 19: Code does not seem to appear in storybook:
image

Am I missing something?

src/@next/SideSheet/SideSheetStyle.ts
```diff
@@ -0,0 +1,71 @@
...
+export const StyledButtonContainer = styled.div`
+  cursor: pointer;
+  margin-right: 8px;
+  svg {
+    width: 24px;
+    fill: ${Neutral.B40};
+
+    &:hover {
+      fill: ${Neutral.B18};
+    }
```
michac789: will delete this

src/@next/SideSheet/SideSheet.stories.tsx
ninariccimarie: you can add something like this: https://github.com/glints-dev/glints-aries/blob/v4/src/%40next/Carousel/Carousel.stories.tsx#L39-L66

src/@next/SideSheet/SideSheet.tsx
```diff
@@ -0,0 +1,113 @@
...
+} from './SideSheetStyle';
+import { Typography } from '../Typography';
+
+export interface SideSheetProps {
+  isOpen: boolean;
+  title: string;
+  description?: string;
+  showHorizontalLine?: boolean;
+  children?: React.ReactNode;
+  leftButtonAction?: ComponentAction;
```
ninariccimarie: normally we call this type of button, Basic and the other one is Primary. so you can use basicButtonAction and primaryButtonAction

src/@next/SideSheet/SideSheetStyle.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export const StyledSideSheetContainer = styled.div`
+  position: absolute;
+  display: flex;
+  flex-direction: column;
+  width: 480px;
+  height: 90vh;
+  background: ${Neutral.B100};
+  box-shadow: 0px 26px 80px rgba(0, 0, 0, 0.2), 0px 0px 1px rgba(0, 0, 0, 0.2);
+  overflow: hidden;
```
michac789: TODO for me:
- make side sheet shown to the right on large screen
- add animation based on figma

src/@next/SideSheet/SideSheetStyle.ts
```diff
@@ -0,0 +1,86 @@
...
+      opacity: 1;
+    }
+  }
+
+  @media (max-width: ${Breakpoints.large}) {
+    width: 100vw;
+  }
+`;
+
+export const StyledSideSheetHeader = styled.div`
```
chiahou, line 81: align-item should be set to top with 4px of top and bottom margins added to the back and close buttons

src/@next/SideSheet/SideSheetStyle.ts
```diff
@@ -0,0 +1,86 @@
...
+  background-color: #eeeeee;
+  margin: 0px 0px;
+  border: none;
+`;
+
+export const StyledSideSheetFooter = styled.div`
+  margin-top: auto;
+`;
+
+export const StyledButtonGroupContainer = styled.div`
```
chiahou, line 111: margin should be 16px 20px

src/@next/SideSheet/SideSheetStyle.ts
```diff
@@ -0,0 +1,86 @@
...
+    width: 24px;
+    fill: ${Neutral.B40};
+  }
+`;
+
+export const StyledRightContainer = styled.div`
+  margin-left: auto;
+`;
+
+export const StyledHorizontalLine = styled.hr`
```
chiahou, line 100: Incorrect color used here. #eeeeee doesn't exist in our color scheme. This color should be replaced with Neutral-B95

src/@next/SideSheet/SideSheetStyle.ts
```diff
@@ -0,0 +1,86 @@
+import styled from 'styled-components';
+import { Neutral } from '../utilities/colors';
```
chiahou, line 2: Incorrect color used for text in header and body. #000 should be replaced with Neutral-B18

chiahou (approved):
looks good to me! thank you @michac789

src/@next/SideSheet/SideSheet.tsx
```diff
@@ -0,0 +1,156 @@
...
+      const enableScroll = () => (document.body.style.overflow = 'unset');
+      isOpen ? (document.body.style.overflow = 'hidden') : enableScroll();
+      return () => {
+        enableScroll();
+      };
+    }, [isOpen]);
+
+    // fade out effect for 0.2seconds when closed
+    const handleClose = () => {
+      const wrapperDiv = document.getElementById('side-sheet-wrapper-unique');
```
marthayaputra: Is it not possible doing this via css + react state?

src/@next/SideSheet/SideSheet.tsx
```diff
@@ -0,0 +1,156 @@
...
+        }
+      }, 10);
+    };
+
+    return (
+      <>
+        {isOpen && (
+          <Portal>
+            <StyledSideSheetWrapper
+              id="side-sheet-wrapper-unique"
```
marthayaputra: I think it's better to use className instead of Id here

src/@next/SideSheet/SideSheet.tsx
michac789: let me try, I think I can do something like:
`sidesheet.closed {...` for the css

marthayaputra (approved):

PR #833, opened 2023-05-23
Bar component enhancement

michac789:


src/@next/Bar/BarStyle.ts
```diff
@@ -2,14 +2,20 @@ import styled from 'styled-components';
...
   flex-direction: row;
-  padding: 20px 0;
+  padding: ${props => (props.smallScreen ? '8' : '12')}px 0;
   width: 100%;
   position: absolute;
-  height: 90px;
+  min-height: ${props => (props.smallScreen ? '70' : '90')}px;
+  height: fit-content;
 
   @media (max-width: ${Breakpoints.large}) {
```
chiahou: @michac789 Given that we are using min-height here, I believe we should remove the height attribute to prevent content overflow.

chiahou (approved):
qc passed, thanks!

src/@next/Bar/Bar.stories.tsx
```diff
@@ -13,7 +13,10 @@ export default {
   component: Bar,
   decorators: [
     Story => (
-      <BaseContainer style={{ height: '200px' }}>{Story()}</BaseContainer>
+      <>
```
ninariccimarie, line 17: i think you can use instead: 
`[withGlintsPortalContainer]`

ninariccimarie:
not really sure about the diff for the other snapshots that were updated. when swiping, there doesn't seem to be any difference? 🤔 can you help to verify this @michac789

michac789:
> not really sure about the diff for the other snapshots that were updated. when swiping, there doesn't seem to be any difference? 🤔 can you help to verify this @michac789

preety sure I changed the spacings a bit based on the Figma (new), so I expected minor change (just a slight space) but yeah tbh not really visible hmm, let me check real quick and will update you again


Edit:
@ninariccimarie 
So it turns out yeah the snapshots are exactly the same 🤣 
Thanks for pointing that out 😄 
On small screen, there is a subtle difference. On large screen however, if the content is short there is no difference at all. What was being modified on large screen was the padding on the vertical axis, however due to the minimum height property it does not affect anything since the content is short.

*Edit2: I reverted that commit, is it okay? If not okay perhaps I can rebase then pick the commit (not sure if it will work but I can try). or worst case I can create a new branch and commit everything again*

src/@next/Bar/Bar.stories.tsx
michac789: @ninariccimarie 

edit:
This works:
```
  decorators: [
    Story => (
      <>
        <BaseContainer style={{ height: '200px' }}>{Story()}</BaseContainer>
      </>
    ),
    withGlintsPortalContainer,
  ],
```


~~using that decorator only caused this issue:~~
image

~~currently it is like this:~~
image
~~(I need to put the customStyle height of 200px)~~

michac789: @ninariccimarie ~~do you want me to modify the decorator so it can take in an argument of customStyle to be passed on to the div instead, or just leave it as it is right now (but I don't use that decorator)?~~

michac789: will settle with this for now, lmk if u have anything else

ninariccimarie (approved):
LGTM 👍 thanks

PR #837, opened 2023-05-25
Select with async options

ninariccimarie:


chiahou (approved):
looks good to me, thank!

farhanabi (approved):

src/@next/Select/Select.tsx
```diff
@@ -93,13 +94,33 @@ export const Select = ({
...
+
+  useEffect(() => {
+    if (inputValue != '' && options.length > 0) {
+      setPopoverActive(true);
+    }
+
+    if (inputValue === '' && options.length < 1) {
+      setPopoverActive(false);
+    }
+  }, [inputValue, options]);
```
marthayaputra: can we destruct the length from options here so we can use it in dependency array instead of the whole `options`?

test/e2e/select/select.spec.ts-snapshots/select-async-searchable-single-select-filter-options-no-match-chromium-linux.png
marthayaputra, line 1: Should we add one more scenario when the input value is blank, no popup should appear?

ninariccimarie: yep we should. i just updated the tests

marthayaputra (approved):

PR #838, opened 2023-05-29
Banner created fixed variant

michac789:


src/@next/Banner/Banner.stories.tsx
```diff
@@ -43,3 +43,90 @@ WithButtons.args = {
...
+      >
+        {defaultText}
+      </Banner>
+      Etiam vel mauris vehicula, bibendum elit id, rutrum sapien. Fusce
+      efficitur dolor vel lorem gravida lacinia. Donec at malesuada diam. Nullam
+      fermentum arcu at ligula ultricies sagittis. Sed erat dui, tempus in
+      luctus ac, venenatis et enim. Cras sed sollicitudin sem. Etiam id quam
+      vitae nisl lacinia sollicitudin vitae suscipit massa. Nullam quis viverra
+      nisl. Integer tempus hendrerit elementum. Maecenas pharetra risus metus.
+      Interdum et malesuada fames ac ante ipsum primis in faucibus. Cras dictum
```
michac789, line 90: I noticed that we have several lorem ipsum texts like this used for stories in various other components. Rather than writing something like this and make the code a bit dirty, how about if we make utility function to generate lorem ipsum text (e.g., just need to call `generateLoremIpsum(300)`, where 300 is the number of words).

A downside is it might take a while (O(n)) where n is the number of words to do this. (e.g., store lorem as a list, slice the list and join it together as strings with space). What do you think?

src/@next/Banner/Banner.tsx
```diff
@@ -9,9 +9,10 @@ import {
   StyledBannerTitle,
   StyledBannerTitleContainer,
   StyledCloseIconWrapper,
+  StyledFixedCloseIconWrapper,
 } from './BannerStyle';
 
-export type BannerProps = {
+export interface BannerProps extends React.HTMLAttributes<HTMLDivElement> {
```
michac789, line 21: Before, if we passed an extra prop (e.g., style), it caused a typescript error. This is unrelated to my functionality, but I think needed to prevent getting those unnecessary typescript warnings.

src/@next/Banner/Banner.tsx
```diff
@@ -21,7 +22,8 @@ export type BannerProps = {
   secondaryAction?: React.ReactNode;
   dismissable?: boolean;
   onDismiss?: () => void;
-};
+  type?: 'static' | 'fixed' | 'fixed-noicon';
```
michac789: I wasn't sure if this is the best thing... so based on the specification:
- type can be static (default)
- type can be fixed, and **only** for fixed type  there is an option to set the icon as optional

For now this is my approach

src/@next/Banner/Banner.tsx
```diff
@@ -58,37 +61,60 @@ export const Banner = React.forwardRef<HTMLDivElement, BannerProps>(
...
+          {action &&
+            (secondaryAction !== 'undefined' ? (
               <ButtonGroup>
                 {action}
                 {secondaryAction}
               </ButtonGroup>
+            ) : (
+              { action }
             ))}
         </StyledBannerContentContainer>
```
michac789: preety sure this was buggy before (?)

Before, if you have both `action` and `secondaryAction`, you can only see the `action` and not the `secondaryAction`. So essentially anytime action is defined, then render action; otherwise if it is not defined and secondary action is defined then render both, which I think is a wrong logic. Fixed this issue.

src/@next/Banner/Banner.tsx
```diff
@@ -58,37 +61,60 @@ export const Banner = React.forwardRef<HTMLDivElement, BannerProps>(
...
+              </StyledCloseIconWrapper>
+            )}
+          </StyledBannerTitleContainer>
+        ) : (
+          type === 'fixed' && (
+            <StyledBannerContentContainer data-type={type}>
+              <Icon name={iconNameValue} />
+            </StyledBannerContentContainer>
+          )
+        )}
```
michac789: Just wanted to ask for suggestions...

Do you think it is confusing to see the logic like this (and please refer to the `BannerStyle` file too)? Was contemplating creating an entirely different styled div and make the jsx seperate too for the fixed one, but it made a few things repeated (though can be more readable and less confusing to edit fixed or static seperately, as the static banner is quite distinct too in terms of stylings compared to the fixed one)

e.g.,
```
if (type === "static") return <>former code</>
else return <>here I create new styled div with all the fixed stylings</>
```

src/@next/Banner/BannerStyle.ts
```diff
@@ -72,6 +72,27 @@ export const StyledBanner = styled.div`
     fill: ${Red.B93};
   }
 
+  &[data-type='fixed'],
+  &[data-type='fixed-noicon'] {
+    position: fixed;
```
michac789: so in Figma it was written:
> Position - Fixed Below Header

I was about to set `top: m px;` where m is the height of the header (was about to ask Maeve which header she meant or how tall is it), but I realized it made the component nonadaptive to different codebases if we do that...
so I didn't set the top property, but instead, if you want let's say make the banner below the header that is 50px tall, then you have to put in extra style as a prop instead to modify that manually

lmk what u think, or if anyone has other suggestions

src/@next/Banner/Banner.tsx
marthayaputra: Are those 2 types entirely different component or just the style is differ?

src/@next/Banner/Banner.tsx
marthayaputra: Hmmm wouldn't  undefined is valid value in jsx? It just won't render anything right? Is there a test cases for rendering action and secondAction?

src/@next/Banner/Banner.tsx
michac789: the component are the same but the ordering (and some of the child component) could be a bit different too (which makes me think if I should seperate them) 

For example, before it was A B C D
And now I want to make it B C D A, with some extra rules (conditionals) whether it exists or not

what annoys me is that there are many conditionals there of whether it is static or fixed or fixed-noicon, I feel it made the code a bit more confusing compared to before, wdyt? is this okay?

src/@next/Banner/Banner.tsx
marthayaputra: If I read the jira ticket here: https://glints.atlassian.net/browse/SS-1057, it seems like change behavior from previously static to fixed, do we still need static? Also what's the `fixed-noicon` for? 🤔

michac789: Figma link: https://www.figma.com/file/87fhOdF8vk1Uiy20bCqS85/Sirius-Web---UI-Kit-3.0-%5BQ4-2022%5D?type=design&node-id=4633-49953&t=mHqK1ifLLMEo7v3B-0

I think they still need the static banner and was just requesting a new variant (fixed) banner. In the Figma above, on left bottom it is written there:
> Only the fixed banner can set the icon as optional

So I assume for static banner, there should always be an icon (yes, the icon prop is optional, but there is a default icon value); while for this one, I assume they want an option to not have the icon completely

src/@next/Banner/Banner.tsx
michac789: I saw the visual tests, preety sure no tests involving 2 actions are there yet;

Yes, undefined should be valid. But the logic was something like this:
```
<StyledBannerContentContainer>
    {action ||
      (secondaryAction && (
        <ButtonGroup>
          {action}
          {secondaryAction}
        </ButtonGroup>
      ))}
</StyledBannerContentContainer>
```

cmiiw, I think with that logic:
- If action is not undefined, render action
- If action is undefined and if secondaryAction is not undefined, render button group

The problem here is that if action and secondaryAction is both defined, it will just render action and not the button group with both actions.

src/@next/Banner/Banner.tsx
michac789: Perhaps after this, I'll also be creating the visual test cases for the banner with two actions too

src/@next/Banner/Banner.tsx
michac789: Perhaps after this, I'll also be creating the visual test cases for the banner with two actions too

src/@next/Banner/Banner.tsx
marthayaputra: @michac789 , but why do we need that as an option tho? From the `bannerStyle` file they have the same style.

src/@next/Banner/Banner.tsx
marthayaputra: Ohhhh you're right, because the the secondary action statement is not part of the condition, good catch, I think we should hoist the conditions like

```
const hasActions = action || secondaryAction;

<StyledBannerContentContainer>
    {hasActions && (
        <ButtonGroup>
          {action}
          {secondaryAction}
        </ButtonGroup>
      )}
</StyledBannerContentContainer>
```

src/@next/Banner/Banner.tsx
marthayaputra: Ah ic, yeah better to separate them and use early return for readability

src/@next/Banner/Banner.tsx
marthayaputra: I'm more incline to have a prop called `showIcon` or something

src/@next/Banner/Banner.tsx
michac789: Needed for line 87 to 91 in `Banner.tsx`, and that's it actually, you're right it is not used anywhere in `BannerStyle.tsx`

```
          type === 'fixed' && (
            <StyledBannerContentContainer data-type={type}>
              <Icon name={iconNameValue} />
            </StyledBannerContentContainer>
          )
```

Essentially, `fixed-noicon` is just `fixed` but without any icon 🤣 
It's different than not giving the `icon` prop though (I can't just check if `icon` prop is undefined or not). The default behavior is that if we don't give an icon prop, a default icon will be assigned depending on the `status` prop.

Was thinking to create another boolean prop (e.g., `hasIcon`), but since based on Figma this should only be possible for fixed type only, so yeah I did this instead.

src/@next/Banner/Banner.tsx
michac789: It is used in `Banner.tsx` here only: (line 87-91)

```
          type === 'fixed' && (
            <StyledBannerContentContainer data-type={type}>
              <Icon name={iconNameValue} />
            </StyledBannerContentContainer>
          )
```

Essentially, it is the same as fixed but it has no icon, that's all 🤣 

At first I also wanted to create a boolean prop `showIcon`, but the Figma says only fixed banner can has no icon, nd the original (static) type should always has an icon. So I settled with those 3 options, but maybe separate it into 2 props would be better (?) 

Note: It is different from checking whether the `icon` prop is given or not. By default, an icon will be given based on the `status` prop if `icon` prop is not given.

chiahou (commented):
@michac789 Some colors and spacing are inconsistent with the design and the buttons should go to the next line on small screens (max-width: 768px). Could you take a look at the values I [highlighted in pink on Figma](https://www.figma.com/file/87fhOdF8vk1Uiy20bCqS85/Sirius-Web---UI-Kit-3.0-%5BQ4-2022%5D?type=design&node-id=5532-54447&t=JuLDhY1Rif0aGj3G-4)?

src/@next/Banner/Banner.tsx
marthayaputra: I think we can make `showIcon` default to `true` here as that's a default behavior and shouldn't be  rendered when it's  `false`. If user set no icon for `static` then I think they should be consciously doing that. WDYT?

michac789: sounds good!

src/@next/Banner/BannerStyle.ts
marthayaputra: I think we should have a default gap from header by default, but I suggest to give a class name for the section which user can override easier later

michac789: So, do you mean having some default value (let's say top: 50px), and just let the user pass in extra style prop (e.g., style={{ top: "20px" }}) in case they want to override the default one? something like this is okay? (if yes, then I'll confirm with Maeve the height of the header)

src/@next/Banner/Banner.tsx
michac789: okay will made another commit soon to actually separate these 2 (it's currently still combined and yes they're so messy and has lower readability)

src/@next/Banner/BannerStyle.ts
```diff
@@ -72,20 +72,86 @@ export const StyledBanner = styled.div`
...
+    border: 0;
+    border-radius: 0;
+    left: 0;
+    right: 0;
+    height: 56px;
```
chiahou: I see the banner measuring 100px in height. Seemingly, it is being overridden by the min-height attribute which is set to 100px.
image

src/@next/Banner/BannerStyle.ts
```diff
@@ -97,7 +163,26 @@ export const StyledBannerTitle = styled.div`
...
+  &[data-type='static'] {
+    button {
+      margin-top: 10px;
+    }
+  }
+`;
+
+export const StyledFixedCloseIconWrapper = styled.div`
+  cursor: pointer;
+  color: ${Neutral.B40};
```
chiahou: I still see the close button in different colors for different types of banners
image
image

src/@next/Banner/FixedBannerStyle.tsx
```diff
@@ -0,0 +1,116 @@
+import styled from 'styled-components';
```
marthayaputra: Do we need  it to be `.tsx`?

src/@next/Banner/BannerStyle.ts
marthayaputra: We don't need to expose the style, what I meant is user can override them if they want with styled-component from their side, as long as the element is given a proper selector

src/@next/Banner/Banner.tsx
```diff
@@ -53,43 +64,87 @@ export const Banner = React.forwardRef<HTMLDivElement, BannerProps>(
...
+              </StyledCloseIconWrapper>
+            )}
+          </StyledBannerTitleContainer>
+          <StyledBannerContentContainer>
+            <Typography as="div" variant="body1">
+              {children}
+            </Typography>
+          </StyledBannerContentContainer>
+          <StyledBannerContentContainer>
+            {action &&
```
marthayaputra: Is this not working? https://github.com/glints-dev/glints-aries/pull/838#discussion_r1209072975

src/@next/Banner/Banner.tsx
```diff
@@ -53,43 +64,87 @@ export const Banner = React.forwardRef<HTMLDivElement, BannerProps>(
...
           <Typography as="div" variant="body1">
             {children}
           </Typography>
-        </StyledBannerContentContainer>
-        <StyledBannerContentContainer>
-          {action ||
-            (secondaryAction && (
+        </StyledFixedBannerContentContainer>
+        <StyledFixedBannerButtonContainer data-noicon={showIcon ? '' : 'true'}>
+          {action &&
```
marthayaputra: same, is this not owring? https://github.com/glints-dev/glints-aries/pull/838#discussion_r1209072975

src/@next/Banner/Banner.tsx
```diff
@@ -53,43 +64,87 @@ export const Banner = React.forwardRef<HTMLDivElement, BannerProps>(
       ? iconName
       : iconByStatus || iconNameStatusMap['info'];
 
+    if (type === 'static')
+      return (
```
marthayaputra, line 79: Can we extract the duplicated codes into functions?

src/@next/Banner/FixedBannerStyle.tsx
michac789: ah no because we don't need any jsx there, okay will rename that to .ts only, similar to the other style files

src/@next/Banner/Banner.tsx
michac789: oh the problem with that is I need to check if action is given but secondaryAction is not given, I'll just display `{action}` without the buttonGroup, so it is not working as I intended

Is the current code not readable or not clean enough?
The current logic that I intended is:
- If action is defined:
- - If secondaryAction is defined: show button group with both actions
- - Else: show only action

src/@next/Banner/Banner.tsx
michac789: same reason as above

src/@next/Banner/Banner.tsx
michac789: umm which one is duplicated? because although similar, the ones with static and fixed are quite different; not only the ordering, but also it uses different styled div and the fixed ones has some additional data-XX attribute given, finally I refactor the css for fixed banner on a separate file so hopefully it is more readable

src/@next/Banner/FixedBannerStyle.tsx
michac789: renamed that

src/@next/Banner/Banner.tsx
marthayaputra: Any reason why you need to get rid of button group when only action present? Is it breaking the style? I think ternary is quite hard to read esp in the jsx part

src/@next/Banner/Banner.tsx
marthayaputra: for example the actions rendering parts

src/@next/Banner/BannerStyle.ts
marthayaputra: I don't think we need to expose the style, but just put some className to the part which users can override later

chiahou (approved):
qc passed, @michac789 please move forward, thanks

src/@next/Banner/Banner.tsx
michac789: Let me check soon, was concern that there might me some additional unintended spacings if I use button group... but I think you're right that we can still use the button group even there is only one button, okay will change this soon

michac789: I'll just make it like this, I hope it's readable enough now:
```
    const actionComponent = (
      <>
        {action && (
          <ButtonGroup>
            {action}
            {secondaryAction && secondaryAction}
          </ButtonGroup>
        )}
      </>
    );
```

src/@next/Banner/Banner.tsx
```diff
@@ -53,43 +64,82 @@ export const Banner = React.forwardRef<HTMLDivElement, BannerProps>(
       ? iconName
       : iconByStatus || iconNameStatusMap['info'];
 
+    const actionComponent = (
+      <>
+        {action && (
```
marthayaputra: what if I only define secondary action but not action? 🤔

michac789: no buttons will be rendered for now

michac789: because I think you're not supposed to use secondaryAction if you don't give action in the first place, or wait a minute I have an alternative:

marthayaputra: @michac789 I don't think there's a validation logic stopping users from doing that?

michac789:
```
      <>
        {(action || secondaryAction) && (
          <ButtonGroup>
            {action && action}
            {secondaryAction && secondaryAction}
          </ButtonGroup>
        )}
      </>
```

So if you:
Define neither action nor secondaryAction -> do not show any button group
Define action only -> show button group with action
Define secondaryAction only -> show button group with secondaryAction
Define both action and secondaryAction -> show button group with both action and secondaryAction

wdyt @marthayaputra ?

marthayaputra: I think can just let action  & secondary action render without condition to avoid more complexity
   ```
   <>
        {(action || secondaryAction) && (
          <ButtonGroup>
            { action}
            {secondaryAction}
          </ButtonGroup>
        )}
      </>
```

marthayaputra: IIRC undefined and null will be rendered as nothing with react, can verify this with the visual tests

michac789: yup" you're right I just checked it... I wasn't really aware of this before, thanks

marthayaputra (approved):

PR #841, opened 2023-06-05
Button change icon size from 16px to 20px

michac789:
https://glints.atlassian.net/browse/SS-1071


michac789:
New PR: https://github.com/glints-dev/glints-aries/pull/848

Closed to rename branch name to feature/XXX, so there can be a playwright preview

PR #844, opened 2023-06-05
Banner fix text overflow and right padding issue

michac789:
- make font weight shown in playwright to 600 as requested
- fixed banner on small screen now supports multiple line description (note: only stylings for fixed banner, on small screen, and when description exceed one line, is changed; otherwise it should remain the same with what Maeve has approved)


src/@next/Banner/Banner.tsx
```diff
@@ -123,7 +123,11 @@ export const Banner = React.forwardRef<HTMLDivElement, BannerProps>(
             <Icon name={iconNameValue} />
           </StyledFixedIconWrapper>
         )}
-        <StyledFixedBannerContentContainer data-noicon={showIcon ? '' : 'true'}>
+        <StyledFixedBannerContentContainer
+          data-noicon={showIcon ? '' : 'true'}
```
marthayaputra: I think we don't need ternary operator here, just 

`data-noicon={!showIcon}`

or better yet can change to

`data-show-icon={showIcon}`

Applicable to other data attributes

michac789: I think it only applies for boolean right? will change the one for `showIcon` and `dismissable`

michac789: @marthayaputra done

marthayaputra: @michac789 I think you missed out this one 🤔 
`
data-has-action={action || secondaryAction ? 'true' : ''}`

src/@next/Banner/Banner.tsx
```diff
@@ -123,12 +123,16 @@ export const Banner = React.forwardRef<HTMLDivElement, BannerProps>(
             <Icon name={iconNameValue} />
           </StyledFixedIconWrapper>
         )}
-        <StyledFixedBannerContentContainer data-noicon={showIcon ? '' : 'true'}>
+        <StyledFixedBannerContentContainer
+          data-show-icon={showIcon}
+          data-has-action={action || secondaryAction ? 'true' : ''}
+          data-dismissable={dismissable}
+        >
           <Typography as="div" variant="body1">
```
chiahou: hi @michac789 according to Rebecca, the text style here should be changed to body2

src/@next/Banner/FixedBannerStyle.ts
```diff
@@ -37,14 +37,16 @@ export const StyledFixedBanner = styled.div`
...
     svg {
       margin-top: 0px;
     }
 
     &[data-nobutton='true'] {
-      height: 52px;
+      min-height: 52px;
     }
   }
 `;
```
chiahou, line 52: According to Rebecca's comment on Slack, the icon size should be 24x24 instead of 20x20 
image

src/@next/Banner/FixedBannerStyle.ts
```diff
@@ -76,23 +78,30 @@ export const StyledFixedBannerContentContainer = styled.div`
...
+      padding-bottom: 48px;
+    }
+
+    &:not([data-show-icon='true']) {
+      padding-left: 0px;
     }
   }
 `;
 
 export const StyledFixedBannerButtonContainer = styled.div`
```
chiahou, line 98: hi @michac789 I don't think we can set the position of the ButtonContainer to absolute, because we do not know how much content we are going to display in a fixed banner and the content will inevitably break into multiple lines on small screens. On top of that, we would like to see a 16px gap between the text and buttons. Otherwise, we will see an overlap like this in the screenshot.
image

michac789: Let me check... so my idea was that the height of the banner depends on the height of text (add padding bottom for the text container), then put the botton with absolute position from the bottom of the banner (not top)... let me check, i think it worked previously

src/@next/Banner/Banner.tsx
michac789: Maybe I'll make it to:
`!!(action || secondaryAction)`

Thought that since it is not boolean I should use ternary, but I think I can just use the snippet above

src/@next/Banner/FixedBannerStyle.ts
michac789: hmm for me it is displayed properly in storybook:
image


Haven't changed anything (still using the storybook link that I created last Monday):
image


Can you check if you're referring to the correct storybook link?
(Storybook Preview: https://615abff875b8f5004aa277ac-fybmupupdy.chromatic.com/) -> based on my latest commit last Monday (05 June 2023)


I think it should work on small screen because:
- Banner has min-height, if the content is large then the height depends on the content (description), I added padding-bottom of 48px if there is an action (instead of 12px because the extra 36px is for the button; note that there is still 16px padding on the banner container itself, so the distance between the bottom border of content div to the bottom border of the banner div is (12 + 36 + 16)px)
- The action is located 16px absolute from the bottom of the page, it should take the extra 36px I have left 

image

Please verify and lmk if somehow it doesn't work on your side!

src/@next/Banner/Banner.tsx
michac789: changed @marthayaputra

marthayaputra (approved):

src/@next/Banner/FixedBannerStyle.ts
chiahou: somehow it works well on my end now, but could you help to increase the padding bottom by 4px, as the gap should be 16px instead of 12px @michac789 thanks!

michac789: @chiahou done

chiahou (approved):
qc passed

PR #848, opened 2023-06-05
Button change icon size from 16px to 20px

michac789:
Jira Link: https://glints.atlassian.net/browse/SS-1071


chiahou (approved):
qc passed, thanks

PR #850, opened 2023-06-07
TextArea new component

michac789:
Jira Link: https://glints.atlassian.net/browse/SS-1072


src/@next/TextArea/TextAreaStyle.ts
```diff
@@ -0,0 +1,83 @@
...
+import { Neutral, Red } from '../utilities/colors';
+import { NotoSans } from '../utilities/fonts';
+
+export const StyledTextAreaContainer = styled.div`
+  position: relative;
+  width: 100%;
+  max-width: 520px;
+`;
+
+export const StyledTextArea = styled.textarea`
```
chiahou: @michac789 Currently, the overflowing text may overlap with the character counter, which is not in line with the design. As the row is set to 3, we want to avoid the text breaking into the 4th line.
image

src/@next/TextArea/TextAreaStyle.ts
```diff
@@ -0,0 +1,83 @@
...
+  background: ${Neutral.B100};
+  box-sizing: border-box;
+  border: 1px solid ${Neutral.B68};
+  border-radius: ${borderRadius4};
+  padding: 8px 0px 25px 8px;
```
chiahou: Actually, the paddings of the text area should be 8px 12px, with two elements placed vertically inside
image

src/@next/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,63 @@
...
+          value={value}
+          rows={rows}
+          onChange={handleChange}
+          data-error={isError}
+          data-disabled={disabled}
+          disabled={disabled}
+          {...props}
+        />
+        <StyledWordCountContainer data-disabled={disabled} data-error={isError}>
+          <Typography as="span" variant="caption">
```
chiahou: @michac789 could you please change caption to overline?

src/@next/TextArea/TextAreaStyle.ts
michac789: I used absolute positioning for the character count, but yea you're right it caused the problem above that text might overlap with the character counter because of this... fixing this issue now...

src/@next/TextArea/TextArea.tsx
michac789: sure, changing it now...

src/@next/TextArea/TextAreaStyle.ts
```diff
@@ -0,0 +1,92 @@
...
+import { NotoSans } from '../utilities/fonts';
+
+export const StyledTextAreaContainer = styled.div`
+  position: relative;
+  width: 100%;
+  max-width: 520px;
+
+  border: 1px solid ${Neutral.B68};
+  border-radius: ${borderRadius4};
+  padding: 8px 12px 25px 12px;
```
michac789, line 16: equivalent of 8px padding-x and 12px padding-y; the 25px is 12px for padding bottom and 17px height to put the character count (which I used absolute positioning as I feel it's easier)

src/@next/TextArea/TextArea.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const Interactive = Template.bind({});
+Interactive.args = {
+  placeholder: 'Please enter...',
+  disabled: false,
+  rows: 3,
+  maxLength: 60,
+  width: 520,
+};
+
+const WithPopulatedValueTemplate: Story<TextAreaProps> = args => {
```
ninariccimarie: I think there's no need to create this example so you can just remove this.

src/@next/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,106 @@
...
+        width={width}
+        onClick={handleContainerClick}
+      >
+        <StyledTextArea
+          ref={setRefs}
+          value={value}
+          rows={rows}
+          width={width}
+          onChange={handleChange}
+          data-disabled={disabled}
```
ninariccimarie: you won't need this since text area has disabled props already

src/@next/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,106 @@
...
+        />
+        {!!maxLength && (
+          <StyledWordCountContainer
+            data-disabled={disabled}
+            data-error={isError}
+          >
+            <Typography
+              as="span"
+              variant="overline"
+              style={{ fontSize: '12px' }}
```
ninariccimarie: you don't need this since overline is already 12px.

src/@next/TextArea/TextAreaStyle.ts
```diff
@@ -0,0 +1,98 @@
...
+export const StyledTextArea = styled.textarea<TextAreaProp>`
+  background: ${Neutral.B100};
+  box-sizing: border-box;
+  border: none;
+
+  font-family: ${NotoSans}, sans-serif;
+  font-style: normal;
+  font-weight: 400;
+  font-size: 16px;
+  line-height: 150%;
```
ninariccimarie: you can export `body1` from `TypographyStyles.ts` and import it here inline instead to apply a much more accurate styles for this. Just rename the import name so it's easy to know that it's the body1 typography styles applied. it'll be easier to maintain as well since we don't have to bother maintaining this separately in case there are changes with Typography styles in the future.

ninariccimarie (changes requested):
Don't forget to add an export in `src/@next/index.ts` as well

src/@next/TextArea/TextArea.stories.tsx
michac789: @ninariccimarie  just want to inform that removing this will also remove one snapshot from test case (the one with populated value); I created this example actually just purely for testing (as in Figma this variant is shown)

As the `value` prop is used in useState I can't just add query params like `&args=value:abcde` to get the snapshot of populated text area, lmk if u have other alternative of testing this, or for now I'll remove the snapshot

ninariccimarie: i see. you can add a test to fill the textArea. https://playwright.dev/docs/api/class-elementhandle#element-handle-fill

src/@next/TextArea/TextAreaStyle.ts
```diff
@@ -0,0 +1,94 @@
...
+    background: ${Neutral.B95};
+  }
+`;
+
+export const StyledTextArea = styled.textarea<TextAreaProp>`
+  background: ${Neutral.B100};
+  box-sizing: border-box;
+  border: none;
+
+  ${variantCssMapping['body1']}
```
ninariccimarie: just exporting and importing `body1` would be faster.
Do rename it on import: `Import { body1 as typographyBody1 } from '../Typography/TypographyStyles';`

chiahou (approved):
qc passed, thanks

src/@next/TextArea/TextAreaStyle.ts
michac789: hmm do you want me to add `export` to every texts... because currently they are not exported

michac789: okay for now I'll just add export to all text styles

ninariccimarie (approved):
LGTM 🚀

PR #852, opened 2023-06-13
Upload: create new component

michac789:
https://glints.atlassian.net/browse/SS-1080


src/@next/Upload/Upload.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const LoadingTemplate: Story<UploadProps> = args => {
+  const [file, setFile] = useState<File | null>(null);
+
+  return (
+    <>
+      <Upload {...args} file={file} setFile={setFile} loading={true} />
+    </>
+  );
+};
```
michac789: actually, I shouldn't even need to make `loading` as a prop here... I just make this so that the designers can see the loading component for QC...

imo though, the loading component is not that relevant (almost always never be seen or you just see a 0.05sec flash of it)

src/@next/Upload/UploadStyle.ts
```diff
@@ -0,0 +1,142 @@
...
+    top: 17.5px;
+    svg {
+      fill: ${Blue.S99};
+      height: 18px;
+      width: 18px;
+    }
+  }
+`;
+
+export const StyledUploadedImage = styled.img`
```
chiahou: @michac789 for the loaded image, can we set the object-fit to "cover" but perhaps expose this property to allow customization. Ideally, we do not want to see the uploaded image being distorted or cut and thus I'd like to make it flexible.
image

src/@next/Upload/Upload.stories.tsx
michac789: will remove exposing loading as prop as qc is completed

michac789: nvm im not removing it as it's still used for testing... when testing, I can't see loading state hmm.... it automatically goes to image state, even when I'm not using await or adding any time delays

chiahou (approved):
all good, thanks!

src/@next/Icon/icons/ri-attachment-fill.svg
ninariccimarie, line 1: @michac789 can you reexport this? I think it's not on the correct format.

src/@next/Upload/Upload.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+  component: Upload,
+  decorators: [Story => <BaseContainer>{Story()}</BaseContainer>],
+  argTypes: {},
+} as Meta;
+
+const Template: Story<UploadProps> = args => {
+  const [file, setFile] = useState<File | null>(null);
+
+  return (
+    <>
```
ninariccimarie: no need for this wrapper

src/@next/Upload/Upload.tsx
```diff
@@ -0,0 +1,157 @@
...
+  StyledErrorContainer,
+  StyledErrorIconContainer,
+  StyledErrorTextContainer,
+} from './UploadStyle';
+import { Spinner } from '../Spinner';
+import { Typography } from '../Typography';
+
+export interface UploadProps {
+  file: File | null;
+  setFile: React.Dispatch<any>;
```
ninariccimarie: it's best practice to pass functions instead of setStates for separation of concerns and maintenance. it would be less coupled with the consuming app

src/@next/Upload/Upload.tsx
```diff
@@ -0,0 +1,157 @@
...
+      if (file) reader.readAsDataURL(file);
+    }, [file]);
+
+    const handleClick = () => {
+      if (fileInputRef.current) {
+        fileInputRef.current.click();
+      }
+    };
+
+    const MAX_FILE_SIZE = 5; // in MB
```
ninariccimarie: NIT: can you move this at the top before the useStates?

src/@next/Upload/Upload.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+    const uploadedImageComponent = (
+      <StyledUploadedImage src={attachmentUrl} objectFit={objectFit} />
+    );
+
+    return (
+      <>
+        <StyledInvisibleInput
+          type="file"
+          accept="image/png, image/jpeg"
```
ninariccimarie: based on the requirements:
> The upload trigger is used to upload a single image, such as a profile photo or company logo. Only image formats are supported, including PNG, JPG, GIF, SVG, BMP, and TIFF. 

it should allow other image formats. you can do:
`accept="image/*"` instead

src/@next/Upload/Upload.tsx
```diff
@@ -0,0 +1,157 @@
...
+        data-type="initial"
+        data-error={!!error}
+        data-testid="upload-container"
+      >
+        <StyledIconContainer>
+          <Icon name="ri-add" />
+        </StyledIconContainer>
+        <StyledTextContainer>
+          <Typography as="span" variant="button">
+            Upload
```
ninariccimarie: do keep in mind that we use i18n, it's better to expect a ReactNode for this and use "Upload" as default

src/@next/Upload/Upload.tsx
```diff
@@ -0,0 +1,157 @@
...
+    );
+
+    const loadingComponent = (
+      <StyledUploadContainer data-type="loading">
+        <StyledIconContainer data-type="loading">
+          <Spinner />
+        </StyledIconContainer>
+        <StyledTextContainer>
+          <Typography as="span" variant="overline">
+            loading...
```
ninariccimarie: same comment regarding i18n

ninariccimarie (changes requested):

src/@next/Icon/icons/ri-attachment-fill.svg
michac789: sorry I removed that instead, due to change of specs, that icon is not being used

ninariccimarie (approved):
LGTM 👍

PR #853, opened 2023-06-14
PasswordInput: create new component

michac789:
https://glints.atlassian.net/browse/SS-1087?atlOrigin=eyJpIjoiMWRhNGY0YWE4M2I4NDgxOGE3MGQ4ZTlhZDA1NTExMmQiLCJwIjoiaiJ9


src/@next/PasswordInput/PasswordInputStyle.ts
```diff
@@ -0,0 +1,8 @@
+import styled from 'styled-components';
+
+export const StyledSuffixContainer = styled.div`
+  height: 17px;
```
chiahou: @michac789 size of a suffix icon should be 20px x 20px with 8px of paddings on the left and right
image

chiahou (commented):
@michac789 according to the specs, we should allow users to see the last typed character briefly as they type on mobile devices exclusively

src/@next/PasswordInput/PasswordInputStyle.ts
```diff
@@ -1,8 +1,12 @@
 import styled from 'styled-components';
 
 export const StyledSuffixContainer = styled.div`
-  height: 17px;
+  height: 20px;
+  transform: translateX(4px);
```
michac789, line 5: the suffix by default has 12px padding right and 4px padding left, I can't really change that directly, so I just use transformation here... it should get me 8px padding left and 8px padding right now

michac789:
> @michac789 according to the specs, we should allow users to see the last typed character briefly as they type on mobile devices exclusively

this one is a bit complicated, still have some problem doing it, maybe I'll try on next sprint :)

chiahou (approved):
looks good to me, thanks!

marthayaputra (approved):
LGTM

PR #858, opened 2023-06-21
chore: use React.ReactNode as the type for modal header property

Afsal102:


ninariccimarie (approved):
LGTM. do update the pr before merging

PR #861, opened 2023-06-22
Add TikTok icon

roshnet:
Icon name: _ri-tiktok-line_


roshnet:
Merging due to urgency, can be reviewed later.

PR #864, opened 2023-06-27
Feat: Show initial value when it is available

farhanabi:


chiahou (approved):
qc passed, thanks

src/@next/Select/components/SearchableSelectInput/SearchableSelectInput.tsx
```diff
@@ -140,10 +140,20 @@ export const SearchableSelectInput = forwardRef<
...
       return;
     }
 
     setShowClear(false);
+    updateSearchableSelectState({
+      showSelected: false,
+      showInput: true,
+      showPlaceholder: true,
+    });
   }, [selectedValue]);
```
marthayaputra: @farhanabi there's seem to be a warning about useEffect dependencies here

marthayaputra (changes requested):

src/@next/Select/components/SearchableSelectInput/SearchableSelectInput.tsx
farhanabi: Any idea how to solve this? If I add `updateSearchableSelectState` to the dependency array, it will do a lot of unnecessary render. I tried wrapping `updateSearchableSelectState` with `useCallback` in the parent component when it initializes, the rendering keeps happening.

marthayaputra: Does it need to be run under useEffect? The function seems to depend on `selectedValue` or `showClear`?

marthayaputra (changes requested):

marthayaputra (approved):

PR #865, opened 2023-06-28
MobileModal: implement modal style for mobile device

michac789:
Note: test case failed because of loading time... it is a bit inconsistent, still trying to fix this now, but shouldn't impact QC


chiahou (approved):
looks good to me, thanks!

PR #866, opened 2023-06-28
Feature/refactor-text-area

Afsal102:


src/@next/TextArea/TextArea.tsx
```diff
@@ -20,82 +19,80 @@ export type TextAreaProps = Omit<
...
-      }
-    };
+  const transformedMaxLength: number = maxLength || undefined;
 
-    useEffect(() => {
-      setCharCount(value.length);
-    }, [value]);
+  const hasMaxLengthEnforced = Boolean(
+    transformedMaxLength && transformedMaxLength > 0
+  );
```
michac789: Nit: `const hasMaxLengthEnforced = transformedMaxLength > 0`

Afsal102: should be fixed in the latest commit,
tl;dr the reason for adding this was that typescript will show linter errors as this value could be undefined too (but i don't get that warning on this codebase though)

chiahou (approved):
looks good to me, thanks!

src/@next/TextArea/TextArea.tsx
```diff
@@ -20,82 +19,93 @@ export type TextAreaProps = Omit<
...
-    const localRef = useRef<HTMLTextAreaElement>(null);
-    const setRefs = (node: HTMLTextAreaElement) => {
-      localRef.current = node;
-      if (typeof ref === 'function') {
-        (ref as (instance: HTMLTextAreaElement | null) => void)(node);
-      } else if (ref) {
-        (ref as MutableRefObject<HTMLTextAreaElement | null>).current = node;
-      }
-    };
+  const transformedMaxLength: number = maxLength || undefined;
```
ninariccimarie: do you still need this? if you don't pass maxLength, won't the value already be undefined?

michac789: I think he needs that when maxLength is 0

Afsal102: @ninariccimarie reason for adding this was to avoid confusion, for some reason when i hover on the maxLength it always shows as `maxLength: number` instead of `maxLength?:number` 
which leads to confusion as to why its always defined(the same happens for transformedMaxlLength), let me know if we show remove this

ninariccimarie: @Afsal102 you can remove this. I verified that when you start typing to pass maxLength in TextArea component, it will show you the correct type.
Screenshot 2023-07-05 at 11 49 59 AM

engineers can verify through the TextAreaProps as well from the TextareaHTMLAttributes

Afsal102: @ninariccimarie but when u are working inside the TextArea implementation, this doesnt show it like this! nevertheless  i agree with the second point

ninariccimarie (approved):
LGTM! thanks

PR #868, opened 2023-06-30
Pagination: add new pagination type

davidzzz:


ninariccimarie (changes requested):
the first things i can spot:
1. borders are not present
2. there shouldn’t be a gap in between the 2 if there’s no label in between
3. support for the optional label is missing

src/@next/Pagination/PaginationStyle.tsx
```diff
@@ -58,6 +63,30 @@ export const StyledSimplePaginationButton = styled(StyledPageButton)`
   pointer-events: none;
 `;
 
+export const StyledBorderPaginationButton = styled(StyledPageButton)`
+  border: 1px solid ${Neutral.B68};
```
chiahou: incorrect color. should be #D4D5D8 (Neutral.B85)

src/@next/Pagination/PaginationStyle.tsx
```diff
@@ -58,6 +63,30 @@ export const StyledSimplePaginationButton = styled(StyledPageButton)`
...
+
+  &:first-of-type {
+    border-radius: ${borderRadius2} 0 0 ${borderRadius2};
```
chiahou: incorrect border radius. should be 4px

src/@next/Pagination/PaginationStyle.tsx
```diff
@@ -58,6 +63,30 @@ export const StyledSimplePaginationButton = styled(StyledPageButton)`
...
+  }
+
+  &:last-of-type {
+    border-radius: 0 ${borderRadius2} ${borderRadius2} 0;
```
chiahou: incorrect border radius. should be 4px

src/@next/Pagination/PaginationStyle.tsx
```diff
@@ -58,6 +63,30 @@ export const StyledSimplePaginationButton = styled(StyledPageButton)`
...
 `;
 
+export const StyledBorderPaginationButton = styled(StyledPageButton)`
+  border: 1px solid ${Neutral.B68};
+
+  &:first-of-type {
+    border-radius: ${borderRadius2} 0 0 ${borderRadius2};
+  }
+
+  &:last-of-type {
```
chiahou, line 80: please add either `border-right-style: none` to the previous button OR `border-left-style:none` to the next button

src/@next/Pagination/paginationUtils.ts
```diff
@@ -9,6 +11,14 @@ export type PaginationProps = {
   onPageChanged?: (currentPage: number) => void;
 };
 
+export type IconOnlyPaginationProps = {
```
chiahou: the label property on Storybook is not working and thus I cannot see how a label between buttons looks like. Could you help to fix that so that I can check the style?

src/@next/Pagination/IconOnlyPagination.tsx
```diff
@@ -0,0 +1,35 @@
...
+  ref
+) {
+  return (
+    <IconOnlyPaginationStyledNav ref={ref}>
+      <StyledBorderPaginationButton disabled={hasPrevious} onClick={onPrevious}>
+        <Icon name="ri-arrow-m-left-line" />
+      </StyledBorderPaginationButton>
+      {label && (
+        <StyledLabel>
+          <Typography as="div" variant="body1">
```
chiahou: we're using "caption" for the label here and the color should be #2d2d2d `Neutral-B18`
image

src/@next/Pagination/PaginationStyle.tsx
```diff
@@ -58,6 +63,30 @@ export const StyledSimplePaginationButton = styled(StyledPageButton)`
...
+    background: ${Blue.S08};
+  }
+
+  &:disabled {
+    background: ${Neutral.B95};
+  }
+`;
+
+export const StyledLabel = styled.div`
+  padding: ${space4};
```
chiahou: label should be vertically centered with 12px of paddings on the left and right
image

src/@next/Pagination/paginationUtils.ts
davidzzz: yeah, forgot to handle the label property after changing the label type. will fix this one

src/@next/Pagination/PaginationStyle.tsx
davidzzz: just tried this. this works out if there is no label given. but, if a label is given, it's obvious that either the right-side of prev button or the left-side of next button will miss a border. the result will be like this image below
image

src/@next/Pagination/IconOnlyPagination.tsx
davidzzz: removed typography tag as the label type is changed to ReactNode

src/@next/Pagination/PaginationStyle.tsx
chiahou: Can we disable the "border-right-style" or "border-left-style" effect when label is given? @davidzzz

chiahou: btw @davidzzz we should not apply both "border-right-style" and "border-left-style" to the buttons. My suggestion was to pick one instead.

chiahou: @davidzzz wait! I found a way to achieve the desired outcome with css. give me a sec. let me test that out for the one with label first

davidzzz: I will check more on this whether we can disable the "border-right-style" or "border-left-style" effect when label is given. will be a bit trickier to pull it off

src/@next/Pagination/PaginationStyle.tsx
```diff
@@ -58,6 +63,33 @@ export const StyledSimplePaginationButton = styled(StyledPageButton)`
   pointer-events: none;
 `;
 
+export const StyledBorderPaginationButton = styled(StyledPageButton)`
+  border: 1px solid ${Neutral.B85};
```
chiahou: @davidzzz please change the property from `border` to `outline` with value remaining the same and set the `column-gap` to 1px. And remove the 'border-right-style' and 'border-left-style' for the first-of-type and last-of-type

src/@next/Pagination/IconOnlyPagination.tsx
chiahou: @davidzzz the text color should be neutral-b18, it's still missing from the code

src/@next/Pagination/PaginationStyle.tsx
```diff
@@ -15,6 +15,12 @@ export const SimplePaginationStyledNav = styled(StyledNav)`
...
 
+export const IconPaginationStyledNav = styled(StyledNav)`
+  gap: 0px;
+  align-items: center;
+  column-gap: 1px;
+`;
+
 export const StyledPageButton = styled.button<PageButtonProps>`
   cursor: pointer;
   height: 32px;
```
chiahou, line 26: @davidzzz just realized that the size is incorrect either. The buttons should be 36px x 36px instead of 32px x 32px

chiahou (approved):
qc passed, thanks

src/@next/Pagination/paginationUtils.ts
```diff
@@ -9,6 +11,14 @@ export type PaginationProps = {
   onPageChanged?: (currentPage: number) => void;
 };
 
+export type IconPaginationProps = {
+  label?: React.ReactNode;
+  hasNext?: boolean;
```
ninariccimarie: can we use `disableNext` and `disablePrevious` with a default of `false`? `has` seems like it should not be there but what we want is for it to be disabled.

ninariccimarie (changes requested):
can you add visual tests for this?

ninariccimarie (approved):
LGTM! thanks!

PR #870, opened 2023-07-03
Carousel: toggle indicator and arrows

ninariccimarie:


src/@next/Carousel/Carousel.tsx
```diff
@@ -30,8 +30,10 @@ export interface CarouselProps {
   width?: string;
   indicatorPosition?: 'top' | 'bottom' | 'left' | 'right' | 'outer';
   indicatorType?: 'dot' | 'line' | 'slider';
-  /** When to show arrows*/
-  showArrow?: 'always' | 'hover' | 'never';
+  /** Sets whether to show indicator */
+  showIndicator?: boolean;
```
marthayaputra: can we use `indicatorType` for this instead? If it's falsy we don't show it

ninariccimarie: ~~yep, good idea~~
hmmm but thinking about it, that means we can't have a default for `indicatorType` and we'll always need to ask them for this prop

chiahou (approved):
looks good to me, thanks

src/@next/Carousel/Carousel.tsx
marthayaputra: I think it's okay to have default value just assign it to null if not needed and the prop is optional anyway right?

src/@next/Carousel/Carousel.tsx
```diff
@@ -138,7 +136,7 @@ export const Carousel = ({
           </NextArrow>
         </ArrowsContainer>
       )}
-      {showIndicator && (
+      {indicatorType != null && (
```
marthayaputra: probably we can check if it's falsy instead of specifically check for `null`,  coz since it's optional, devs can pass `undefined` as well

marthayaputra (approved):

PR #871, opened 2023-07-04
Card: add optional action props on header

michac789:


src/@next/Button/Button.tsx
```diff
@@ -3,7 +3,7 @@ import { Spinner } from '../Spinner/Spinner';
 import { Typography } from '../Typography';
 import { StyledButton } from './ButtonStyle';
 
-export type ButtonSize = 'default' | 'slim' | 'large';
+export type ButtonSize = 'default' | 'slim' | 'large' | 'verySlim';
```
michac789, line 6: image
I found no styles that can give the wanted spacings for the button (between the label and the border), so I had to create this... wdyt? maybe also document this in Figma?

the spacing is: padding 5px horizontal 2px vertical for large screen, 0px for small screen

src/@next/Button/ButtonStyle.ts
```diff
@@ -46,10 +47,24 @@ min-width: 67px;
...
+  padding: 5px ${space2};
+
+  @media (max-width: ${Breakpoints.large}) {
+    padding: 0px 0px;
+  }
+
+  p {
+    margin-block-start: 0px;
+    margin-block-end: 0px;
+  }
```
michac789, line 60: I encountered some weird vertical spaces affecting Typography component inside button, something like this:
image

It says 'user agent stylesheet', when I researched it seems to be some default stylings by the browser, so I just overwrite those

src/@next/Card/CardStyle.ts
```diff
@@ -15,8 +15,8 @@ export const StyledCardContainer = styled.div`
 export const StyledCardHeaderWrapper = styled.div`
   display: flex;
   padding: ${space24} ${space24} 20px;
-  flex-wrap: wrap;
-  flex-direction: column;
+  flex-wrap: nowrap;
+  flex-direction: row;
```
michac789, line 19: tbh I am concerning about this... if someone put a few div as the header content (custom header), the default behavior changes, the div will be in row instead of column

src/@next/Button/Button.tsx
chiahou: @michac789 I've just checked the design file and there are some issues with the `button group` & `button` settings on Figma. But it looks like the problems exist on storybook as well. The button group component is somehow removing the padding of plain buttons?
image
 Can we just use the slim size instead? I'll make time today to fix the issues on Figma to see if we need to make adjustments to Aries.

chiahou: @michac789 I've just checked the design file and there are some issues with the `button group` & `button` settings on Figma.  Actually, we don't need three different sizes for plain buttons... Anyway, instead of creating 'verySlim', can we adjust the spacing of 'slim' as shown in the screenshot? 
image

chiahou: And then we can use slim buttons with 4px gap (now it's 8px) @michac789

michac789: @chiahou sure, so now it has 4px padding in all directions for large screen, how about the spacing for the small screen?

michac789:
closed: changed to https://github.com/glints-dev/glints-aries/pull/873

PR #874, opened 2023-07-06
fix: Remove erratic behavior from Carousel

roshnet:
https://glints.atlassian.net/browse/DS-9


chiahou (approved):
looks good to me, thanks

PR #880, opened 2023-07-11
feat: upgrade storybook

finalight:


src/@next/Bar/Bar.stories.tsx
```diff
@@ -217,12 +217,12 @@ const customActionGroupTemplate: Story<BarProps> = args => {
...
 
-export const customActionGroup = customActionGroupTemplate.bind({});
-customActionGroup.parameters = {
+export const CustomActionGroup = customActionGroupTemplate.bind({});
+CustomActionGroup.parameters = {
   docs: {
     source: {
       code: `
-      const customActionGroup = (
+      const CustomActionGroup = (
```
finalight, line 225: storybook plugin dictates that PascalCase should be used

src/@next/Combobox/comboboxStoryHelper/ComboboxMultuSelect.tsx
```diff
@@ -87,7 +87,7 @@ export const ComboboxMultiSelect = ({
   }, [isSearchEmpty, options]);
 
   return (
-    <div style={{ maxWidth: '500px' }}>
+    <div style={{ maxWidth: 500 }}>
```
finalight, line 90: i need to use non-unit number else it will have problem rendering this storybook

src/@next/Select/components/Activator/ActivatorSelect.tsx
```diff
@@ -76,13 +77,14 @@ export const ActivatorSelect = ({
     if (hasSelectedValues) {
       return (
         <TagsContainer>
-          <StyledTag
+          <Tag
+            style={{ width: 28, marginRight: space8 }}
```
finalight, line 81: somehow it will have error rendering this page if we use styled component of `Tag` in this page

so i have to not use styled component instead assign a style like the above to the tag

src/@next/Select/components/Activator/ActivatorStyle.ts
```diff
@@ -59,10 +57,10 @@ export const TagsContainer = styled.div`
   align-items: center;
 `;
 
-export const StyledTag = styled(Tag)`
-  height: 28px;
-  margin-right: ${space8};
-`;
+// export const StyledTag = styled(Tag)`
```
finalight, line 60: keeping this because of the above

src/Input/Textarea/Textarea.stories.tsx
```diff
@@ -17,8 +17,8 @@ Interactive.args = {
   label: 'Description',
 };
 
-export const removeFloatingLabelIsTrue = Template.bind({});
-removeFloatingLabelIsTrue.args = {
+export const RemoveFloatingLabelIsTrue = Template.bind({});
```
finalight, line 20: storybook plugin dictates that PascalCase should be used

src/@next/Select/components/Activator/ActivatorSelect.tsx
marthayaputra: Hmmm can we investigate this more? I don't think we should update the implementations to cater storybook upgrade

src/@next/Select/components/Activator/ActivatorStyle.ts
marthayaputra: same comment as above

PR #881, opened 2023-07-12
Card: component actions disabled

ninariccimarie:
- allows adding logic to disabling of action buttons

card-disabled-actions-chromium-linux


marthayaputra (approved):
LGTM

PR #883, opened 2023-07-12
Card: ComponentAction extend ButtonProps

ninariccimarie:


michac789 (changes requested):
hmm I think you also need to pass the ComponentAction other attributes as a prop to the button, maybe something like: (untested)
```
<PlainButton
  onClick={headerPrimaryAction.action}
  size="slim"
  {...headerPrimaryAction} // this line here
>
  {headerPrimaryAction.label}
</PlainButton>
```

ninariccimarie:
oh yeah @michac789 forgot about that. thanks for the catch

michac789 (approved):
LGTM!

Maybe just a note: since we're extending componentAction which is used by other components too, imo it's great if we can pass the componentAction as a prop on buttons in other components too (e.g., Bar, SideSheet, etc., even though they're not used now)

Someone might be confused if for example, they see that they can pass `disabled` in `ComponentAction` as a prop in `Bar` component, but nothing changes on the component

ninariccimarie:
thanks @michac789 ! I've created this jira ticket as a to do: https://glints.atlassian.net/browse/DS-12

PR #886, opened 2023-07-13
Tab: remove unwanted vertical margin

michac789:


chiahou (commented):
@michac789 could you also help to remove `margin-bottom:1px` for `<button>` inside `<li>`. The margin is not required, since designers will define the gap between tabs and content below on a case-by-case basis

michac789:
> @michac789 could you also help to remove 'margin-bottom:1px' for inside
> 
> * ? The margin is not required, since designers will define the gap between tabs and content below on a case-by-case basis

@chiahou what `margin-bottom` are you referring to hmm? I don't see any margin-bottom in TabStyle, no margin bottom here:
image

chiahou (commented):
qc passed, thanks

chiahou (approved):

PR #890, opened 2023-07-17
Typography: change caption to subtitle2, update caption & button style

michac789:


michac789:
refer to https://github.com/glints-dev/glints-aries/pull/893

PR #893, opened 2023-07-17
Typography: update typography styles

michac789:
https://glints.atlassian.net/browse/DS-6


michac789:
replaced here https://github.com/glints-dev/glints-aries/pull/896/files

PR #894, opened 2023-07-18
fix: set loading to be false if no file provided

ganiirsyadi:


michac789 (changes requested):
Can you change the line that you added to:
```
else if (!loading) setIsLoading(false);
```
If loading=true is given (tbh don't think this is even being used lol), you will always see the loading spinner. Your logic makes it so that even if loading=true, the loading state will be set to false there. And that's the reason that a test case is failing.

Another alternative is actually remove the loading prop entirely (if no one uses it) and remove the corresponding test case there.

ganiirsyadi:
> Can you change the line that you added to:
> 
> ```
> else if (!loading) setIsLoading(false);
> ```
> 
> If loading=true is given (tbh don't think this is even being used lol), you will always see the loading spinner. Your logic makes it so that even if loading=true, the loading state will be set to false there. And that's the reason that a test case is failing.
> 
> Another alternative is actually to remove the loading prop entirely (if no one uses it) and remove the corresponding test case there.

If I added that line, then my problem will still exist, since the `loading` value is `true` in my case. Steps: I upload the SVG image, based on my validation it's failing, hence the `file` props I passed to the Upload component will be null. Though, the loading is already set to `true` due to `handleChange` function that was triggered.

finalight:
can see there's a snapshot test failure, would need that to be updated somehow

michac789 (approved):
lgtm! 

note: can consider removing the loading prop entirely in the future if it is useless

ganiirsyadi:
> lgtm!
> 
> note: can consider removing the loading prop entirely in the future if it is useless

And supporting `imageUrl`

finalight:
is this okay?

image

ganiirsyadi:
> is this okay?
> 
> image

Fixed

PR #896, opened 2023-07-18
Typography: change caption to subtitle2, update caption and button

michac789:


chiahou (approved):
looks good to me, thanks

PR #897, opened 2023-07-19
feat: add file copy line svg

michac789:


src/@next/Icon/icons/ri-file-copy2-line.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 17 17">
```
ninariccimarie: can you check the export? it seems the wrong size. this should be 24 by default when you export

michac789: fixed

ninariccimarie (approved):

PR #899, opened 2023-07-19
Tooltip: option to show tooltip when clicked instead of hover

michac789:


src/@next/Tooltip/Tooltip.tsx
```diff
@@ -154,6 +157,9 @@ export const Tooltip = ({
   content,
   preferredPosition = defaultPosition,
   zIndex,
+  clickable = false,
+  timeout = 1000,
+  onClick,
```
ninariccimarie, line 162: i think we can just use this to know whether it's clickable or not.

michac789: you mean use `onClick` to determine whether it is clickable or not?
if that's the case, not sure if it's possible, I tried and it didn't work, the truth value (e.g., `!onClick`) will be false regardless I pass a function or not...

Maybe isn't that the reason why in other components you need something similar like this?
image
(there is separate `onBack` and `showBackButton` props here, it could've been merged into one, but I think the reason it's not is because it does not work, unless I'm passing `ComponentAction` that it worked)

also, if we determine using that, it's harder to test and play around in glints-aries (you can't just toggle the `onClick` function in storybook unlike boolean values... lmk if you have alternatives, thanks!

ninariccimarie (approved):

src/@next/Tooltip/TooltipStyle.ts
```diff
@@ -131,6 +131,42 @@ export const StyledTooltip = styled.div<{ zIndex?: number }>`
     bottom: ${nonCentralArrowMargin};
   }
 
+  &[data-clickable='true'] {
```
chiahou: @michac789 can we apply the animations to the hover event as well to maintain consistency in the tooltip's behavior?

src/@next/Tooltip/TooltipStyle.ts
```diff
@@ -131,6 +131,42 @@ export const StyledTooltip = styled.div<{ zIndex?: number }>`
...
+    animation: fade-in-tooltip 400ms ease-in forwards;
```
chiahou: [tooltip.webm](https://github.com/glints-dev/glints-aries/assets/12844836/71f1f28e-2f60-434f-a75a-7e8081dead9d) ->Not sure what is causing this movement of the tooltip? Could it be due to the animation?

src/@next/Tooltip/TooltipStyle.ts
michac789: oh sure I thought it is just for the clickable, wait a moment

src/@next/Tooltip/TooltipStyle.ts
michac789: I thought you intended that by having the transform: scale(...) on the animation CSS you give me? is that not intended?

chiahou: nope haha, okay I think maybe we should just get rid of transform:scale for now

michac789: or did you mean to actually enlarge the tooltip or make it smaller by that scale instead of moving it?

michac789: okayy will get rid of it first now

chiahou: @michac789 yes, coz we are not setting transform to `translate()` but `scale()` only, so it should have remained its position.

chiahou (approved):

PR #900, opened 2023-07-20
Feature/tooltip clickable extension

michac789:


src/@next/Tooltip/Tooltip.tsx
```diff
@@ -147,7 +147,7 @@ export type TooltipProps = {
   zIndex?: number;
   clickable?: boolean; // if true, tooltip will be shown on click instead of hover
   timeout?: number; // how long tooltip will still be shown after clicked or when not hovered anymore
-  onClick?: () => void; // onClick will be called when tooltip is clicked or hovered
+  onAction?: () => void; // onAction will be called when tooltip is clicked or hovered out
```
ninariccimarie: I just realised, I forgot to mention this in previous PR but can you put the comments in this format:
```ts
 /** description of prop here */
prop: propType;
```

so it'll appear as description in the storybook

src/@next/Tooltip/Tooltip.tsx
```diff
@@ -247,39 +247,45 @@ export const Tooltip = ({
...
-    }, timeout);
+      }
+      setIsActive(true);
+      timeoutRef.current = setTimeout(() => {
+        setAnimate(true);
+        timeoutRef.current = setTimeout(() => {
+          setIsActive(false);
+          setAnimate(false);
+        }, 370); // animation is 400ms, make this slightly lower to prevent visual glitch
+      }, timeout);
```
ninariccimarie: sorry i didn't saw the specs for animations. If we'll have a static time, we won't need the timeout prop and these lines. you can apply the fade in and out in styles instead.

src/@next/Tooltip/Tooltip.tsx
```diff
@@ -247,16 +250,25 @@ export const Tooltip = ({
...
-  const handleMouseEnter = () => setIsActive(true);
+  const handleMouseEnter = () => {
+    if (!clickable) setIsActive(true);
+  };
   const handleMouseLeave = () => {
-    setIsActive(false);
-    handleClick();
+    if (!clickable) {
+      setIsActive(false);
+      handleAnimation();
```
ninariccimarie, line 259: you won't be needing this after moving the animations to the styles

src/@next/Tooltip/Tooltip.tsx
michac789: the animation is static, the timeout is just how long the icon will be there after we click it before it proceeds showing the fade out animation and disappear

ninariccimarie: or you can pass the time on how long should it be present in the animations instead or compute for it

src/@next/Tooltip/Tooltip.tsx
michac789: hmm I need to set animate state to true indicating it's time for fade out effect... the animation is handle on the styles, this is just to update the state (whether to animate fade out or not)...

ninariccimarie (approved):
approving but asking for code improvements on next release

PR #907, opened 2023-07-31
chore: prefix sidesheet animation

marthayaputra:


michac789 (approved):
approved but I'll just leave this comment:

what does `ga` in `ga-fade-in2` and `ga-fade-out2` stand for?
I'd prefer to have something like `sidesheet-fade-in` and `sidesheet-fade-out` instead, to make it clear the animation is for sidesheet and other components or other animation won't use that name again... sorry my bad earlier for using `fade-in2` and `fade-out2`, I think it was a bad naming as it's prone to cause conflicts and confusion for the other devs

marthayaputra:
> approved but I'll just leave this comment:
> 
> what does `ga` in `ga-fade-in2` and `ga-fade-out2` stand for? I'd prefer to have something like `sidesheet-fade-in` and `sidesheet-fade-out` instead, to make it clear the animation is for sidesheet and other components or other animation won't use that name again... sorry my bad earlier for using `fade-in2` and `fade-out2`, I think it was a bad naming as it's prone to cause conflicts and confusion for the other devs
It's glints aries, hmm I saw project like antd prefix the animation that way. One of the disadvantage of naming it specific to component is, will be awkward if want to reuse it

marthayaputra:
> approved but I'll just leave this comment:
> 
> what does `ga` in `ga-fade-in2` and `ga-fade-out2` stand for? I'd prefer to have something like `sidesheet-fade-in` and `sidesheet-fade-out` instead, to make it clear the animation is for sidesheet and other components or other animation won't use that name again... sorry my bad earlier for using `fade-in2` and `fade-out2`, I think it was a bad naming as it's prone to cause conflicts and confusion for the other devs

it stands for glints-aries, I think in a library animation should be consistent hence the name. I think we can revisit this later when there's another animation

PR #908, opened 2023-07-31
feat: enhance Upload component

ganiirsyadi:
Changes:
- [UPDATE] `loading` prop can always control the loading state of the component, allowing async upload function to enable the loading while uploading
- [NEW] `disableDefaultValidation` prop, since every use case has a different need, I need to be able to disable default validation from the component and control it from my project
- [NEW] `accept` prop, I need to be able to control the subType of images file that I want to accept, ex: only JPG files.

### DIFF
--- a/src/@next/Upload/Upload.tsx
+++ b/src/@next/Upload/Upload.tsx
@@ -22,6 +22,8 @@ export interface UploadProps {
   objectFit?: 'fill' | 'contain' | 'cover' | 'none' | 'scale-down';
   uploadMessage?: React.ReactNode;
   loadingMessage?: React.ReactNode;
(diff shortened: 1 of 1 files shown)


src/@next/Upload/Upload.tsx
```diff
@@ -50,8 +54,9 @@ export const Upload = React.forwardRef<HTMLInputElement, UploadProps>(
         setAttachmentUrl(reader.result as string);
         setIsLoading(false);
       };
+      reader.onloadstart = () => setIsLoading(true);
```
ganiirsyadi, line 57: The internal loading should only start if the reader start loading the image.

michac789 (approved):
LGTM

PR #909, opened 2023-07-31
Typography: download font poppins and noto sans

michac789:
Simply use appropriate font family and weight for everything.

TODO: fix all failing test cases first, DO NOT MERGE


.storybook/preview-head.html
```diff
@@ -0,0 +1,17 @@
...
+<html>
+  <head>
+    <!-- need to import this -->
+    <link
+      rel="stylesheet"
+      href="https://fonts.googleapis.com/css2?family=Poppins:wght@500;600;700&display=swap"
+    />
+    <link
+      rel="stylesheet"
+      href="https://fonts.googleapis.com/css2?family=Noto+Sans:wght@500;600;700&display=swap"
```
chiahou: @michac789 I think we need to add 400 here for noto sans, since body1 and subtitle2 are using regular font weight, which is 400

src/@next/Typography/Typography.tsx
```diff
@@ -1,4 +1,9 @@
 import React, { ElementType, ReactNode } from 'react';
+import '@fontsource/noto-sans/400.css';
```
michac789, line 2: importing on `index.ts` for some reason didn't work...
importing here actually applies this to all of the component in storybook

PR #913, opened 2023-08-04
feat: update index table typography to subtitle2

marthayaputra:
Majority of the font in Index Table is subtitle2, setting this as the default Typography


test/e2e/indexTable/indexTable.spec.ts
```diff
@@ -102,6 +102,7 @@ test('IndexTable - empty state', async ({ page }) => {
 test('IndexTable - scrollable', async ({ page }) => {
   const indexTablePage = new IndexTablePage(page);
 
+  page.setViewportSize({ width: 500, height: 500 });
```
michac789, line 105: just wondering why you need to set such width and height for this test?

michac789 (approved):
LGTM, just one line that I don't really understand

test/e2e/indexTable/indexTable.spec.ts
marthayaputra: To make it obvious it's scrollable

chiahou (approved):

PR #914, opened 2023-08-07
EmptyState: allow custom image

ninariccimarie:


src/@next/EmptyState/EmptyState.tsx
```diff
@@ -24,6 +24,8 @@ export interface EmptyStateProps {
   fullWidth?: boolean;
   helpText?: string;
   imageName?: ImageName;
+  /** Renders a custom image */
```
Afsal102: do we need to have a union here? we would erxpect either customImage | imageName to be present only

src/@next/EmptyState/EmptyState.tsx
```diff
@@ -46,6 +49,7 @@ export const EmptyState = ({
   return (
     <EmptyStateContainer>
       {imageName && <StyledImage src={imageMapping[imageName]} />}
+      {customImage}
```
Afsal102: this seems to be unstlyed? do we allow parent to control the size?

src/@next/EmptyState/EmptyState.tsx
```diff
@@ -36,20 +37,35 @@ export const EmptyState = ({
...
+  }
+
+  if (imageName) {
+    console.warn(
+      'imageName will soon be deprecated. Please use image prop instead'
+    );
   }
 
+  const renderImage = () => {
+    if (typeof image === 'string') {
```
Afsal102: this is still not backwards compatible, u need to check if imageName is present and then do the following as well

src/@next/EmptyState/EmptyState.tsx
```diff
@@ -23,7 +23,10 @@ export interface EmptyStateProps {
   description?: React.ReactNode;
   fullWidth?: boolean;
   helpText?: string;
+  /** will soon be @deprecated. Please use image prop instead */
   imageName?: ImageName;
```
Afsal102: can we use discriminated unions?

src/@next/EmptyState/EmptyState.tsx
```diff
@@ -34,18 +37,35 @@ export const EmptyState = ({
...
   imageName,
+  image,
   title,
 }: EmptyStateProps) => {
   const displayButtons = !!primaryButtonAction || !!basicButtonAction;
 
-  const isValidImageName = imageNames.includes(imageName);
-  if (imageName && !isValidImageName) {
-    console.warn(`imageName "${imageName}" is not a valid Image Name.`);
+  const isValidImageName = imageNames.includes(image as ImageName);
```
Afsal102: can we have the check in line number 58 too? else it might break since we just typecast the image to string blindly

src/@next/EmptyState/EmptyState.tsx
```diff
@@ -63,6 +94,7 @@ export const EmptyState = ({
               <PrimaryButton
                 loading={primaryButtonAction?.loading}
                 onClick={primaryButtonAction?.onClick}
+                className="empty-state-primary-button"
```
davidzzz, line 107: we can add `icon={primaryButtonAction?.icon}` here, so that the icon can show up if we pass the icon, unless `className` here somehow already provides it. maybe can add test to see how it looks

src/@next/EmptyState/EmptyState.tsx
```diff
@@ -71,14 +103,19 @@ export const EmptyState = ({
               <Button
                 loading={basicButtonAction?.loading}
                 onClick={basicButtonAction.onClick}
+                className="empty-state-basic-button"
```
davidzzz, line 117: same here with the above thread, but use `basicButtonAction`

src/@next/EmptyState/EmptyState.tsx
ninariccimarie: ah! thanks for the catch. I forgot to pass the props 😓

src/@next/EmptyState/EmptyState.tsx
```diff
@@ -34,26 +33,58 @@ export const EmptyState = ({
...
+      console.warn(`image "${imageName}" is not a valid Image Name.`);
+    }
+  }
+
+  if (typeof image === 'string' && !imageNames.includes(imageName)) {
+    console.warn(`image "${imageName}" is not a valid Image Name.`);
   }
 
+  const renderImage = () => {
+    if (imageName || typeof image === 'string') {
```
Afsal102: can we do a check check if image is present, currently we only do a typeof check this is gonna be true even when we pass an empty string

Afsal102 (approved):

Afsal102:
LGTM!

PR #917, opened 2023-08-08
feat: fix searchable select not disabled properly

Afsal102:
when the field is disabled with a default value, the clear icon is still showing and the user is able to modify the text in the input,
this is fixed by hiding the clear button when the field is disabled and passing the disabled prop correctly to the text input


ninariccimarie (approved):
LGTM! thanks for the fix

PR #925, opened 2023-08-10
IndexTable: allow overwriting of scrollbar styles

michac789:


michac789:
cancelled due to alternative solution found

PR #926, opened 2023-08-15
DataTable: allow reactnode as datatable heading

michac789:
Reason: need to be able to pass reactnode for header for this (the tooltip, or if it is accompanied with other icons)
image

### DIFF
--- a/src/@next/DataTable/DataTable.tsx
+++ b/src/@next/DataTable/DataTable.tsx
@@ -19,8 +19,11 @@ import { TableRow } from './TableRow';
 
 export type SortDirection = 'ASCENDING' | 'DESCENDING';
 
-export type TableHeading = React.ThHTMLAttributes<HTMLTableColElement> & {
-  title?: string;
+export type TableHeading = Omit<
+  React.ThHTMLAttributes<HTMLTableColElement>,
+  'title'
+> & {
+  title?: React.ReactNode;
   id?: string;
   defaultSortDirection?: SortDirection;
 };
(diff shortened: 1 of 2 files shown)


src/@next/DataTable/DataTable.tsx
```diff
@@ -19,8 +19,11 @@ import { TableRow } from './TableRow';
 
 export type SortDirection = 'ASCENDING' | 'DESCENDING';
 
-export type TableHeading = React.ThHTMLAttributes<HTMLTableColElement> & {
-  title?: string;
+export type TableHeading = Omit<
+  React.ThHTMLAttributes<HTMLTableColElement>,
+  'title'
+> & {
+  title?: React.ReactNode;
```
ninariccimarie, line 26: I think it’s best to include the comment for headings prop in the `DataTableProps`. Can you mention that if the title is of ReactNode type, they must provide the id for sorting to be able to work correctly?

apart from that, can you include a console.warn with the same logic? 👆

michac789: image
@ninariccimarie something like this is fine for the comment?

michac789: image
and this for the console.warn

ninariccimarie: you can wrap it in `/** comment here */` to make it show in the storybook. example: https://github.com/glints-dev/glints-aries/blob/f37a6a716932f485f9d4f717a1192599fe66d7bc/src/%40next/Tabs/Tabs.tsx#L12

ninariccimarie (approved):
LGTM! thanks!

PR #931, opened 2023-08-18
DataTable typography and icons updates

ninariccimarie:


chiahou (commented):
@ninariccimarie it appears that there is excessive empty space considering the smaller font size we are currently using. I wonder if the spacing should be defined here or not? Can it be customized at a product level?

ninariccimarie:
@chiahou If it would make more sense to redefine the spacing as well, then we can redefine from here cause it would be troublesome to redefine spacing from the product side every time we use DataTable

chiahou:
@ninariccimarie I see. Then can we adjust the paddings of each cell from `16px` to `8px 16px`. And one more thing, can we remove the `vertical align: top` or maybe we can expose this attribute

chiahou (approved):
lgtm, thx!

michac789 (approved):
lgtm

PR #933, opened 2023-08-21
DataTable: make emptyState prop optional

michac789:
Simply make the `emptyState` prop in DataTable optional for usage in MTS.
When data is an empty list and `emptyState` is not given, simply do not render any rows nor show the empty state.


ninariccimarie (approved):
LGTM

PR #934, opened 2023-08-21
SideSheet: remove footer div if no action is given

michac789:
image
if no action (buttons) are given, do not show the footer div as it caused styling issues like this


ninariccimarie (approved):
LGTM

PR #935, opened 2023-08-21
Housekeeping: change action to onClick for all ComponentAction

michac789:
simply change `action` to `onClick` for all ComponentAction used in any components, as it will soon be deprecated

### DIFF
--- a/src/@next/Bar/Bar.tsx
+++ b/src/@next/Bar/Bar.tsx
@@ -79,16 +79,16 @@ export const Bar = React.forwardRef<HTMLDivElement, BarProps>(function Bar(
           ) : (
             <ButtonGroup>
               {tertiaryAction && (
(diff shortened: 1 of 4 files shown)


ninariccimarie (commented):
the current implementation in the FE still use the action prop. do you have a follow up work in replacing them or was it communicated with the other engineers? if it's not urgent for other components, let's update the one we need for now unless you have time to update the components in the other FEs and request for reviews or we already communicated the update to the other engineers.

michac789:
> the current implementation in the FE still use the action prop. do you have a follow up work in replacing them or was it communicated with the other engineers? if it's not urgent for other components, let's update the one we need for now unless you have time to update the components in the other FEs and request for reviews or we already communicated the update to the other engineers.

oh yes you're right, okay let me close the PR for now, thanks for reminding this

PR #936, opened 2023-08-21
Select: add customActivatorIcon prop to customize icon

michac789:
Not sure if this is the best way or not...

Simply allow the select icon to be customized (not just the default `ri-arrow-m-down-line`)
image
in this case I need `ri-arrow-m-down-fill`


ninariccimarie (approved):
LGTM

PR #944, opened 2023-08-23
fix: Update placeholder color to Neutral.B68

franciscrispin:
https://glints.atlassian.net/browse/WDS-39

Created a theme constant so changing such styling is not a pain in the future and also to enforce consistency with the design system. 

Want to avoid going down the ThemeProvider route yet as I think that is overkill for this fix. Let me know if there's a better approach!


src/@next/Input/InputStyle.ts
```diff
@@ -104,7 +105,7 @@ export const StyledInput = styled.input<InputProps>`
   height: 36px;
 
   &::placeholder {
-    color: ${Neutral.B40};
+    color: ${theme.placeholder};
```
ninariccimarie: @franciscrispin can we directly change the color instead? since we are not using this kind of pattern in the code 🙇

franciscrispin: Yep can! But was just concerned it will make things harder to change in the future (which was the case when had to dig through all the affected placeholders haha). Wdyt? @ninariccimarie

franciscrispin: @ninariccimarie removed the theme constant!

ninariccimarie: thanks for raising this concern! I'll take this into consideration as a pain point for when we need to update

ninariccimarie (approved):
LGTM

PR #951, opened 2023-08-30
Select: prop to trigger action when select is clicked

michac789:
Need a way to know when select is clicked (opened) for MTS:
> when select component is opened, disable scroll of the layout container


ninariccimarie (approved):
LGTM

PR #955, opened 2023-09-07
Tag: add contentType props, adjust bg color and icon size

michac789:


src/@next/Tag/TagStyle.ts
```diff
@@ -37,14 +54,22 @@ export const TagIconWrapper = styled.div`
...
   padding: 0;
 
+  &[data-content-type='success'] {
+    background-color: ${Green.B89};
+  }
+  &[data-content-type='warning'] {
+    background-color: ${Orange.S21};
+  }
+
   &[data-clickable='true'] {
```
chiahou, line 79: @michac789 both clickable and removeable icons should not be affected by the color adjustments and remain blue

src/@next/Tag/TagStyle.ts
```diff
@@ -74,10 +99,17 @@ export const TagStyle = styled.div<TagProps>`
 
   @media (max-width: ${Breakpoints.large}) {
     font-size: 12px;
```
chiahou: @michac789 since we are using `subtitle2` for text, I wonder why there is a need for specifying font-size here? Additionally, the font family seems incorrect on storybook

src/@next/Tag/TagStyle.ts
```diff
@@ -74,10 +99,17 @@ export const TagStyle = styled.div<TagProps>`
...
+
+    & svg {
```
chiahou, line 103: The size of 18px x 18px for mobile view is not reflected on storybook

src/@next/Tag/TagStyle.ts
```diff
@@ -74,10 +99,17 @@ export const TagStyle = styled.div<TagProps>`
...
+      width: 18px;
+      height: 18px;
+    }
```
chiahou: The size of 18px x 18px for mobile view is not reflected on storybook

src/@next/Tag/TagStyle.ts
michac789: removed that (old code), btw font family is incorrect because the pr here: https://github.com/glints-dev/glints-aries/pull/909 has not been merged; there was an issue with testing, me & Nina tried updating but there was a problem, so in the meantime the correct font family are not used in entire glints-aries

src/@next/Tag/TagStyle.ts
michac789: the 18px x 18px seems correct on my side 🤔

michac789: image

chiahou (approved):

src/@next/Tag/TagStyle.ts
```diff
@@ -16,9 +17,25 @@ export const TagContentStyle = styled.span<TagContentProps>`
...
+export const TagContentWrapper = styled.div`
+  padding: 0 ${space4};
+`;
+
 export const TagRemoveContainerStyle = styled.div<TagRemoveContainerProps>`
-  padding: ${space4};
+  padding: ${space4} ${space4} ${space4} 0;
   display: flex;
+
+  svg {
```
ninariccimarie: this will affect all children that are svgs. Can you add a className on the Icon instead and apply the style based on it?

same with the Icon in the Tag content

ninariccimarie (approved):
LGTM

PR #956, opened 2023-09-12
feat: add Authentication Input component

marthayaputra:
Component to be used as OTP for MFA


src/@next/AuthenticationInput/AuthenticationInputStyle.ts
```diff
@@ -0,0 +1,13 @@
+import styled from 'styled-components';
+import { space4 } from '../utilities/spacing';
```
chiahou: @marthayaputra As the component in Figma is incorrectly wrapped twice, the 4px you saw was incorrect. If you click into the second layer, you'll see 8px, so please help to change `space4` to `space8`.

src/@next/AuthenticationInput/AuthenticationInput.tsx
```diff
@@ -0,0 +1,116 @@
...
+    newInputValues[index] = value;
+    setInputValues(newInputValues);
+    const nextInputIndex = index + 1;
+
+    if (nextInputIndex <= numberOfInputs - 1) {
+      setCurrentIndex(nextInputIndex);
+    }
+  };
+
+  const handlePaste = (event: ClipboardEvent) => {
```
Afsal102, line 61: suggest to use a useCallback here

src/@next/AuthenticationInput/AuthenticationInput.tsx
```diff
@@ -0,0 +1,116 @@
...
+    }
+    const newInputValues = [...inputValues];
+    newInputValues[index] = value;
+    setInputValues(newInputValues);
+    const nextInputIndex = index + 1;
+
+    if (nextInputIndex <= numberOfInputs - 1) {
+      setCurrentIndex(nextInputIndex);
+    }
+  };
```
Afsal102, line 59: suggest to use a useCallback here

src/@next/AuthenticationInput/AuthenticationInput.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const [currentIndex, setCurrentIndex] = useState(0);
+  const inputArray = new Array(numberOfInputs).fill('');
+
+  useEffect(() => {
+    if (inputValues.some(v => !Number.isInteger(v))) {
+      onChange?.(null);
+      return;
+    }
+    onChange?.(inputValues.join(''));
+  }, [inputValues, onChange]);
```
Afsal102: this onChange could cause infinite looping issue if the callback provided by the user keeps changing its reference on every render? should we fix the same here instead of expecting the user/client to pass a callback which does not change its reference every time?

src/@next/AuthenticationInput/AuthenticationInputSingle.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { useEffect, useRef } from 'react';
+import { NumberInput } from '../NumberInput';
+
+interface AuthenticationInputSingleProps
+  extends Omit<React.HTMLAttributes<HTMLInputElement>, 'onChange'> {
+  index: number;
+  value: number | '';
+  shouldFocus?: boolean;
+  error?: boolean;
+  disabled?: boolean;
```
Afsal102, line 10: can we change the cursor pointer when its disabled? for better semantics

src/@next/AuthenticationInput/AuthenticationInputSingle.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { useEffect, useRef } from 'react';
+import { NumberInput } from '../NumberInput';
+
+interface AuthenticationInputSingleProps
+  extends Omit<React.HTMLAttributes<HTMLInputElement>, 'onChange'> {
+  index: number;
+  value: number | '';
```
Afsal102, line 7: i cannot understand this, why do we have an empty string here? cant we just the string type?

src/@next/AuthenticationInput/AuthenticationInput.tsx
marthayaputra: I don't think we need that right? Doesn't seem to be expensive operation

src/@next/AuthenticationInput/AuthenticationInput.tsx
marthayaputra: same as above

src/@next/AuthenticationInput/AuthenticationInputSingle.tsx
marthayaputra: the value can be empty string or a number, since our OTP is a number

src/@next/AuthenticationInput/AuthenticationInput.tsx
Afsal102: the function seems to get created on every render unnecessarily if we don't have any deps with the props in this function and it does not need to recreate, i think its best we can wrap in a callback, but i am fine with keeping it as it is too, we can optimize incase of any issues

src/@next/AuthenticationInput/AuthenticationInputSingle.tsx
marthayaputra: done

src/@next/AuthenticationInput/AuthenticationInput.tsx
marthayaputra: good point, use ref to handle the exhaustive deps warning

src/@next/AuthenticationInput/AuthenticationInputSingle.tsx
```diff
@@ -0,0 +1,70 @@
...
+      {...rest}
+      maxLength={1}
+      onFocus={e => {
+        e.currentTarget.select();
+        onIndexChanged?.(index);
+      }}
+      inputRef={ref}
+      value={value}
+      onChange={e => {
+        console.log({ currentTarget: e.currentTarget.value });
```
ninariccimarie: cleanup for console.log

src/@next/AuthenticationInput/AuthenticationInputSingle.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const ref = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (shouldFocus) {
+      ref?.current?.focus();
+      ref?.current?.select();
+    }
+  }, [shouldFocus]);
+
+  console.log({ singleValue: value });
```
ninariccimarie: cleanup for console.log

src/@next/AuthenticationInput/AuthenticationInputStyle.ts
marthayaputra: Fixed

src/@next/AuthenticationInput/AuthenticationInput.tsx
marthayaputra: Hmm useCallback and useMemo actually has a downside, unless it's an expensive operation the rerender really doesn't make a difference, it only make it hard to read. Here is an interesting take: https://kentcdodds.com/blog/usememo-and-usecallback

src/@next/AuthenticationInput/AuthenticationInput.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  const onChangeRef = useRef(onChange);
+
+  useEffect(() => {
+    if (inputValues.some(v => !Number.isInteger(v))) {
+      onChangeRef.current?.(null);
+      return;
+    }
+    onChangeRef.current?.(inputValues.join(''));
+  }, [inputValues]);
```
ninariccimarie: what are these for? 🤔

marthayaputra: It'll always return null unless all field are filled, can be used for validation in the consumer app

src/@next/Input/InputStyle.ts
```diff
@@ -44,6 +44,7 @@ export const StyledContainer = styled.div<InputProps & PreffixSuffixWidthProps>`
     border: 1px solid ${Neutral.B85};
     background: ${Neutral.B95};
     color: ${Neutral.B85};
+    cursor: pointer;
```
Afsal102: sorry, this should be not-allowed instead of a pointer

src/@next/AuthenticationInput/AuthenticationInput.tsx
marthayaputra: added a documentation for `onChange` method

src/@next/AuthenticationInput/AuthenticationInput.tsx
```diff
@@ -0,0 +1,124 @@
...
+  forwardedRef,
+  ...rest
+}: AuthenticationInputProps) => {
+  const [inputValues, setInputValues] = useState(
+    new Array<number | ''>(numberOfInputs).fill('')
+  );
+  const [currentIndex, setCurrentIndex] = useState(0);
+  const inputArray = new Array(numberOfInputs).fill('');
+
+  const onChangeRef = useRef(onChange);
```
Afsal102: we need to change the refs function on every render, as just incase the onChange handler changes we need that to reflect in the component as well, else we would be calling the outdate on change here


onChangeRef.current = onChange;

src/@next/Input/InputStyle.ts
marthayaputra: OMG lol, what am I drinking, thanks!

ninariccimarie (approved):
LGTM

chiahou (approved):

PR #959, opened 2023-09-19
Collapse Component

michac789:
https://glints.atlassian.net/browse/WDS-35


chiahou (commented):
@michac789 overall, it looks good to me. just wanted to confirm one thing with you, although the elements inside collapse item are split into headerLeft and headerRight, am I correct to say that I can rearrange icons and badges freely?

michac789:
> @michac789 overall, it looks good to me. just wanted to confirm one thing with you, although the elements inside collapse item are split into headerLeft and headerRight, am I correct to say that I can rearrange icons and badges freely?

yes that's right, not just icon and badges, but anything else you want (for example, tooltip, tag, etc.)
you can arrange freely what you want to put on the left side and what do you want on the right side (if you want any)
here is a sample code for the 8th example in the storybook, you can simply say:
- I want a badge to be on the left (next to title)
- I want an icon, followed by another badge on the right of header
```
  <Collapse.Item
    title="Sample Title 8"
    headerLeft={<Badge>Neutral</Badge>}
    headerRight={
      <>
        <Icon name="ri-more" />
        <Badge>AnotherBadge</Badge>
      </>
    }
    defaultCollapsed={false}
    indicator="right"
  >
```

chiahou (approved):
lgtm, thanks!

src/@next/Collapse/CollapseItem.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+export interface CollapseItemProps {
+  /** Content to be displayed when the item is not collapsed, 8px padding applied by default */
+  children: React.ReactNode;
+  /** Title for each collapse item on the header, simply can pass in a string */
+  title: React.ReactNode;
+  /** Extra components to be displayed on the left side of heading, on the right of indicator (if it is left-aligned) and header (title), flex row with gap of 4px is applied */
+  headerLeft?: React.ReactNode;
+  /** Extra components to be displayed on the right side of heading, on the left of indicator (if it is right-aligned), flex row with gap of 4px is applied */
+  headerRight?: React.ReactNode;
```
ninariccimarie, line 23: I don't think we really need this since title is already of type React.ReactNode.

src/@next/Collapse/Collapse.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { CollapseItem, CollapseItemProps } from './CollapseItem';
+import { CollapseComponentContainer } from './CollapseStyle';
+
+export interface CollapseProps extends React.HTMLAttributes<HTMLDivElement> {
+  /** If true, there is outer 1px border with 4px border radius, default is true */
+  hasBorder?: boolean;
+  /** The children should all be 'Collapse.Item' components */
+  children?: React.ReactElement<CollapseItemProps>[];
+  /** Arrow indicator to be put on the very left or right of the header (or none), default is left, applies to all children collapse items */
+  indicator?: 'left' | 'right' | 'none';
```
ninariccimarie: can we use `expandPosition` instead? I think it's clearer this way in context of a Collapse component

src/@next/Collapse/Collapse.tsx
```diff
@@ -0,0 +1,41 @@
...
+    return (
+      <>
+        <CollapseComponentContainer
+          ref={ref}
+          data-border={hasBorder}
+          {...props}
+        >
+          {React.Children.map(children, child => {
+            const childIndicator =
+              (child.props as { indicator?: 'left' | 'right' | 'none' })
```
ninariccimarie: I think you can use Pick here. Pick<CollapseProps, 'indicator'>. that way, we will always be referring directly to the component's prop.

do check my component above regarding indicator

test/e2e/collapse/collapse.spec.ts-snapshots/collapse-no-indicator-chromium-linux.png
ninariccimarie, line 1: the screenshot says no indicator but Sample Title 8 seems to still have it?

michac789: yes because I overwrote that in the code, I made it so that you can say no indicator on Collapse to apply it to the entire Collapse Item (without writing one by one on each item), but you can specifically indicate in a CollapseItem to have for example a right indicator (overriding the default behavior from the parent Collapse div)

src/@next/Collapse/CollapseItem.tsx
michac789: yea but it will be harder for people to add items on left or right;
if there is no `headerLeft` and `headerRight`, you need to manually create the row flex container, with 2 divs inside (one for items at left, and the other for right), since the sample in Figma show some badges & icons used on left & right, I made it like this so that it is easier;
also, I put the `title` inside Typography directly with the appropriate color (depending on disabled or not)

please reply again if you still want me to remove `headerLeft` and `headerRight`

src/@next/Collapse/Collapse.tsx
michac789: I just followed the specs in Figma which is `indicator`
Also, you can expand the CollapseItem anywhere on the header (not just on the arrow), so it's not really the position where we can expand the CollapseItem

please reply again if you still want me to rename this

src/@next/Collapse/Collapse.tsx
michac789: changed to:
```
  const childIndicator =
    (child.props as Pick<CollapseProps, 'indicator'>)?.indicator ||
    indicator;
```
thanks for the feedback

src/@next/Collapse/Collapse.tsx
ninariccimarie: I see. I understand your point. then can we use indicatorPosiition instead? I'm suggesting this since indicator in a different component we have means what type of indicator it will render adding position in our case would be clearer

test/e2e/collapse/collapse.spec.ts-snapshots/collapse-no-indicator-chromium-linux.png
ninariccimarie: i see. it would be better to separate 'none' state of the indicator if it's meant to define the position. it would be much clearer if you have a showIndicator prop for it instead. it is to separate their concern

ninariccimarie (approved):
LGTM! thanks

PR #970, opened 2023-10-03
Banner: fix showIcon bug not working in static banner

michac789:


src/@next/Banner/Banner.tsx
```diff
@@ -84,9 +84,9 @@ export const Banner = React.forwardRef<HTMLDivElement, BannerProps>(
           {...props}
         >
           <StyledBannerTitleContainer>
-            <Icon name={iconNameValue} />
+            {showIcon && <Icon name={iconNameValue} />}
             {title && (
-              <StyledBannerTitle>
+              <StyledBannerTitle data-show-icon={showIcon}>
```
ninariccimarie, line 89: it seems like you can apply this to StyledBanner itself since you're putting the condition on each of the children? can you check?

michac789: If `showIcon` is `true`, I don't want to apply it here: `{showIcon && <Icon name={iconNameValue} />}`
That is one of the children, so I can't apply it to the parent div directly, otherwise I'll get something like this:
image

ninariccimarie: what I also meant was, based on Figma, it will always have the padding with or without the Icon. We should retain that on the parent but apply the spacing of the Icon to itself instead to isolate it from the other children.

I think the structure of the components should also be the same with the fixed banner:
```js
<StyledBanner>
  <Icon>
  <Content>
    {children}
    {actions}
  </Content>
  <CloseIcon>
</StyledBanner>
```

In this case, there'll be no overlapping of the Icon and the content and you'll be able to apply the exact spacing value defined.

michac789: > it will always have the padding with or without the Icon

thanks, let me confirm with the designer first about this
image
not sure if the screenshot here is intended or is it a mistake

michac789: > always have the padding with or without the Icon

asked designer, it is not intended

ninariccimarie (approved):

PR #974, opened 2023-10-09
Popover: Set z-index higher

farhanabi:
Set the Popover's z index higher, so it can be used inside the modal


src/@next/Popover/PopoverStyle.ts
```diff
@@ -145,7 +145,7 @@ export const StyledPopover: any = createGlobalStyle`
 
 .Polaris-PositionedOverlay {
   position: absolute;
-  z-index: 400;
+  z-index: 999;
```
ninariccimarie, line 148: @farhanabi can we add a prop for a zIndexOverride instead? This will probably break some components that are using Popover.

PR #986, opened 2023-11-02
Steps: add new component

michac789:


chiahou (commented):
@michac789 Just realized that I made a mistake in design specs. The dot type shouldn't be the mobile design for normal type. They have different usages. I've already rectified[ the specs. ](https://www.figma.com/file/87fhOdF8vk1Uiy20bCqS85/Aries---UI-Kit-3.0-%5BQ4-2022%5D?type=design&node-id=9709-49740&mode=design&t=Wy8E2sqUsMzdMZaE-4) Apologies for my oversight and thanks for building the dot version. We're going to need it very soon.

src/@next/Steps/StepStyle.ts
```diff
@@ -0,0 +1,88 @@
+import styled from 'styled-components';
+import { space4, space12, space16 } from '../utilities/spacing';
+import { Breakpoints } from '..';
+import { Blue, Neutral, Red } from '../utilities/colors';
+
+export const StepItemWrapper = styled.div`
```
chiahou, line 13: gaps between elements inside a Step is 12px on both large and small screens

src/@next/Steps/Step.tsx
```diff
@@ -0,0 +1,56 @@
...
+            </Typography>
+          )}
+          {variant === 'pending' && (
+            <Typography as="div" color={Neutral.B40}>
+              {index}
+            </Typography>
+          )}
+        </CircleDiv>
+        <Typography
+          as="div"
```
chiahou, line 63: text colors are missing

src/@next/Steps/Step.tsx
```diff
@@ -0,0 +1,56 @@
...
+  ref
+) {
+  return (
+    <>
+      <StepItemWrapper ref={ref}>
+        <CircleDiv data-variant={variant}>
+          {variant === 'completed' && <Icon name="ri-check" />}
+          {variant === 'error' && <Icon name="ri-close" />}
+          {variant === 'processing' && (
+            <Typography as="div" color={Neutral.B100}>
```
chiahou: subtitle2 or caption text styles should be applied to index

chiahou (commented):
I see cursor turn into edit state when I mouse over a text. Can you help to check if cursor is set to 'pointer' when they are clickable, and set to 'default' when not

src/@next/Steps/Step.tsx
```diff
@@ -0,0 +1,76 @@
...
+} from './StepStyle';
+import { Icon } from '../Icon';
+import { Neutral } from '../utilities/colors';
+
+export interface StepProps {
+  /** Styles of step, predetermined from parent component, or you can overwrite this */
+  variant?: 'pending' | 'completed' | 'processing' | 'error';
+  /** Label given to the step component  */
+  label?: string;
+  hasLine?: boolean;
```
ninariccimarie: you should be able to do this in the styles without using a prop to identify whether to render the line or not. you can identify whether to display it or not by knowing the last child (Steps.Step) and not display it from the parent (Steps).

src/@next/Steps/Step.tsx
```diff
@@ -0,0 +1,76 @@
...
+  /** Step number to be shown, by default it's 1,2,3,... from the parent component, or you can overwrite this */
+  index?: number;
+  isDot?: boolean;
```
ninariccimarie: can we use: `type?: 'normal' | 'dot'` to match the one from figma

src/@next/Steps/Step.tsx
```diff
@@ -0,0 +1,76 @@
...
+    hasLine = true,
+    index = 0,
+    isDot = false,
+    clickable = false,
+    handleClick,
+  }: StepProps,
+  ref
+) {
+  return (
+    <>
```
ninariccimarie: can you add a classname for this?

src/@next/Steps/StepStyle.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+`;
+
+export const CircleDiv = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 28px;
+  width: 28px;
+  border-radius: 50%;
```
ninariccimarie: you can use `borderRadiusHalf`

src/@next/Steps/Step.tsx
```diff
@@ -0,0 +1,76 @@
...
+        ref={ref}
+        data-dot={isDot}
+        data-clickable={clickable}
+        onClick={() => handleClick(index)}
+      >
+        <CircleDiv data-variant={variant} data-dot={isDot}>
+          {variant === 'completed' && <Icon name="ri-check" />}
+          {variant === 'error' && <Icon name="ri-close" />}
+          {variant === 'processing' && (
+            <Typography as="div" variant="caption" color={Neutral.B100}>
```
ninariccimarie: please use span to

src/@next/Steps/StepStyle.ts
```diff
@@ -0,0 +1,102 @@
...
+      fill: ${Neutral.B100};
+    }
+  }
+
+  &[data-dot='true'] {
+    height: 10px;
+    width: 10px;
+
+    > div,
+    > svg {
```
ninariccimarie: it's hard to know what this is

src/@next/Steps/Steps.tsx
```diff
@@ -0,0 +1,62 @@
...
+  /** List of indexes (1-indexed) to be marked as error */
+  errorSteps?: number[];
+  /** Step components as child are required (1-indexed) */
+  children?: React.ReactElement<StepProps>[];
+  /** If true, display dot only; default is false; automatically passed to all children */
+  isDot?: boolean;
+  /** If true, steps are clickable; default is false; automatically passed to all children */
+  clickable?: boolean;
+  /** Callback function when step is clicked, index of the step is passed as an argument */
+  handleClick?: (index: number) => void;
```
ninariccimarie: can you remove these for now to avoid confusion on what're working features for Steps component

src/@next/Steps/Steps.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export interface StepsProps extends React.HTMLAttributes<HTMLDivElement> {
+  /** Step index (1-indexed) marked as processing */
+  currentStep: number;
+  /** List of indexes (1-indexed) to be marked as error */
+  errorSteps?: number[];
+  /** Step components as child are required (1-indexed) */
+  children?: React.ReactElement<StepProps>[];
+  /** If true, display dot only; default is false; automatically passed to all children */
+  isDot?: boolean;
```
ninariccimarie: same comment above to use type instead

src/@next/Steps/Steps.tsx
```diff
@@ -0,0 +1,62 @@
...
+            : index + 1 === currentStep
+            ? 'processing'
+            : index + 1 < currentStep
+            ? 'completed'
+            : 'pending';
+          const childVariant =
+            (child.props as Pick<StepProps, 'variant'>)?.variant || variant;
+          const childIndex =
+            (child.props as Pick<StepProps, 'index'>)?.index || index + 1;
+          return React.cloneElement(child, {
```
ninariccimarie: can we try to avoid using this method as there is a warning from React docs: https://react.dev/reference/react/cloneElement can check other alternatives instead?

src/@next/Steps/Step.tsx
michac789: add className here?
```
      <StepItemWrapper
        ref={ref}
        data-dot={type === 'dot'}
        data-clickable={clickable}
        onClick={() => handleClick(index)}
        className="step-item-wrapper"
      >
```

michac789: ^added that, but not sure if it's what you want

src/@next/Steps/Step.tsx
chiahou: @michac789 text colors for pending and completed state is Neutral-B40 (#666)

src/@next/Steps/StepStyle.ts
michac789: can you elaborate more on this? I searched and didn't see any `borderRadiusHalf` CSS properties 🤔

src/@next/Steps/StepStyle.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+`;
+
+export const VerticalLineWrapper = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  padding: ${space4} 0;
+  height: 64px;
+  width: 28px;
```
chiahou, line 85: @michac789 vertical line doesn't centrally align with the circle div on small screens (24px)

src/@next/Steps/StepStyle.ts
michac789: I made `circle-content` className, hopefully it's clearer;
basically when it's dot type, there shouldn't be anything inside of the circle (the numbers & icons)

src/@next/Steps/Steps.tsx
michac789: > I see cursor turn into edit state when I mouse over a text. Can you help to check if cursor is set to 'pointer' when they are clickable, and set to 'default' when not

I didn't implement the clickable yesterday evening when I made the PR but Maeve said that (implying she need the clickable feature too and it's also written in Figma?) ^^^
So on my 2nd commit I added the clickable prop 🤔 
Possible to sync and update me again on the final verdict? thanks

src/@next/Steps/StepStyle.ts
ninariccimarie: you can import and use `borderRadiusHalf` from utils

src/@next/Steps/Steps.tsx
ninariccimarie: @chiahou can we remove these parts in the code first? so it's easier for dev to know these aren't implemented in the code yet

chiahou (approved):

src/@next/Steps/Steps.tsx
chiahou: @ninariccimarie ok np

src/@next/Steps/Step.tsx
ninariccimarie: can you add another one for the React Fragment itself? probably convert it to a div

src/@next/Steps/StepStyle.ts
ninariccimarie: NIT: can you use `circle-icon` instead

ninariccimarie:
LGTM thanks

ninariccimarie (approved):

PR #993, opened 2023-11-15
SideSheet: fix scrollable content

ninariccimarie:
### Before
Kapture 2023-11-15 at 22 26 15

### After
Kapture 2023-11-15 at 22 25 49


michac789 (approved):
tried it on localhost and works fine, just wonder why it's overflow-x instead of y haha but so long as it's fine (pretty sure overflow-y auto works too), also if possible can add example for overflow but if not also no problem

ninariccimarie:
@michac789 oops. thanks for spotting that, i've changed it

PR #996, opened 2023-12-11
feat: Allow non-string children to options

farhanabi:


src/Input/Select/Select.noTypeChecking.test.tsx
```diff
@@ -46,7 +46,7 @@ describe('<Select /> can still work when receives non-string option children or
     const selectInput = queryByRole('combobox');
     userEvent.type(selectInput, '1');
     const filteredOptions = queryAllByTestId('option');
-    expect(filteredOptions.length).toEqual(1);
+    expect(filteredOptions.length).toEqual(3);
```
ninariccimarie, line 64: @farhanabi how is it this changed?

farhanabi: Basically, I removed the filtering that is happening on non-string options because otherwise it will not show anything when the children are not string, even though we passed the options. We could manipulate this by just changing the options passed directly.

src/Input/Select/Select.noTypeChecking.test.tsx
```diff
@@ -9,35 +9,50 @@ import userEvent from '@testing-library/user-event';
...
-  test('non-string option children, search filtering', async () => {
+  test('non-string and non-number option children, search filtering', async () => {
     const Component = () => (
-      <Select label="non-string option children">
-        <Select.Option value="one">{1}</Select.Option>
-        <Select.Option value="two">{2}</Select.Option>
-        <Select.Option value="three">{3}</Select.Option>
+      <Select label="non-string and non-number option children">
+        <Select.Option value="one">
+          <span>1</span>
```
ninariccimarie, line 48: i think you forgot to remove the span blocks here

src/Input/Select/Select.noTypeChecking.test.tsx
```diff
@@ -46,17 +61,26 @@ describe('<Select /> can still work when receives non-string option children or
     const selectInput = queryByRole('combobox');
     userEvent.type(selectInput, '1');
     const filteredOptions = queryAllByTestId('option');
-    expect(filteredOptions.length).toEqual(1);
+    expect(filteredOptions.length).toEqual(3);
```
ninariccimarie, line 64: after removing the span blocks, this should be 1 now instead of 3

src/Input/Select/Select.noTypeChecking.test.tsx
farhanabi, line 48: Oh no, it is intended. I want to add a component inside the option

src/Input/Select/Select.noTypeChecking.test.tsx
```diff
@@ -46,17 +61,26 @@ describe('<Select /> can still work when receives non-string option children or
...
     expect(asFragment()).toMatchSnapshot();
   });
 
-  test('non-string defaultValue', async () => {
+  test('non-string and non-number defaultValue', async () => {
```
ninariccimarie, line 68: it would be better if you can separate the numbers vs non-string and non-number tests to cover it's case since the test had already been there. we will be losing it's coverage if we replace it since they are 2 different cases

src/Input/Select/Select.numberChecking.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+      const { queryByRole, unmount } = render(<Component />);
+      const selectInput = queryByRole('combobox') as HTMLInputElement;
+      expect(selectInput.value).toEqual(expectedValue);
+      unmount();
+    };
+
+    [
+      [1, '1'],
+      [{ key: 'value' }, '[object Object]'],
```
ninariccimarie: i think you don't need to include this since the test is for numberChecking?

src/Input/Select/Select.numberChecking.test.tsx
```diff
@@ -0,0 +1,123 @@
...
```
ninariccimarie: same comment for this one.
https://github.com/glints-dev/glints-aries/pull/996/files#r1424903960

src/Input/Select/Select.numberChecking.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+    const selectInput = queryByRole('combobox');
+    expect(selectInput.getAttribute('value')).toEqual('Rerender');
+
+    rerender(<Component value={1} />);
+    expect(selectInput.getAttribute('value')).toEqual('1');
+
+    rerender(<Component value={[1, 2, 3]} />);
+    expect(selectInput.getAttribute('value')).toEqual('1,2,3');
+
+    rerender(<Component value={{ key: 'value' }} />);
```
ninariccimarie: same comment for this one.
https://github.com/glints-dev/glints-aries/pull/996/files#r1424903960

ninariccimarie (approved):
LGTM

PR #998, opened 2023-12-14
@next Select: sync input value from searchable props

music1353:


src/@next/Select/Select.tsx
```diff
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from 'react';
+import { isEmpty } from 'lodash';
```
ninariccimarie: can you use lodash-es instead?

src/@next/Select/Select.tsx
```diff
@@ -136,14 +145,21 @@ export const Select = ({
     setPopoverActive(!popoverActive);
   };
 
+  useEffect(
+    function syncInputValueFromSearchableProps() {
+      setInputValue(searchableProps?.inputValue);
+    },
+    [searchableProps?.inputValue]
+  );
+
```
ninariccimarie, line 154: can you share what’s the bug that this useEffect is fixing?

ninariccimarie (changes requested):
can you change the branch name to prefix with `feature` instead of `hotfix` it will actually skip some parts of the pipeline because of this

src/@next/Select/Select.tsx
music1353: Sure! you can refer to this video
https://www.loom.com/share/52216222926c4bd98e67d7188d3acdd0?sid=fc28029c-67ef-40cd-a2dc-0481a8b8568b

the context is when user clears the country fields, would auto-clear the city field at the same time.
But you can see the video, if I search any cities, then change the country, the city field will displaying the keyword that I used previously.
I found that it's because select component doesn't sync up the `searchableProps?.inputValue`

src/@next/Select/Select.tsx
music1353: thanks for catching, fixed!

PR #1000, opened 2023-12-15
@next Select: sync input value from searchable props

music1353:


ninariccimarie (changes requested):
@music1353 can you check [Async Searchable Single With Input State](https://615abff875b8f5004aa277ac-urobngfhrm.chromatic.com/?path=/story/next-select--async-searchable-single-with-input-state) 
Kapture 2023-12-15 at 15 56 18
the issue is, when i select the option, it doesn't get selected anymore. same with the https://615abff875b8f5004aa277ac-urobngfhrm.chromatic.com/?path=/story/next-select--async-searchable-single-with-input-state-custom-popover-placeholder as well. 

i think they both broke? 🤔 current version: https://aries.glints.design/?path=/story/next-select--async-searchable-single-with-input-state

multi select searchable also unable to select
Kapture 2023-12-15 at 16 01 36
 after searching and clicking the checkbox

ninariccimarie (approved):
LGTM

PR #1002, opened 2023-12-22
@next Select: Close select options on blur

yongchenglow:
This is related to the issue found by @music1353, where opening multiple selectors does not close the previous one.

There is still a lag time because of the 101ms. We can technically reduce it by reducing 

https://github.com/glints-dev/glints-aries/blob/0655a3bc8579d129461ad4fafd0f03788ef6e94e/src/@next/Select/components/SearchableSelectInput/SearchableSelectInput.tsx#L134

the initial timeout of 100 to 80. However I decided not to do so as we are unclear about the repercussions. I noticed that if we reduce it to 50, we are unable to select a value using the selector.

This is a hacky solution. As the 4pm team have discussed, a proper way requires a rewrite to somehow remove that 100ms blocker.

### DIFF
--- a/src/@next/Select/Select.tsx
+++ b/src/@next/Select/Select.tsx
@@ -114,6 +114,14 @@ export const Select = ({
     setMenuOptions(newState);
   };
 
+  const handleOnBlur = () => {
+    setTimeout(() => {
+      if (searchableSelectState.showSelected) setPopoverActive(false);
+    }, 101);
+
+    onBlur?.();
+  };
+
   const handleClose = () => {
     setPopoverActive(false);
     onClose?.();
@@ -185,7 +193,7 @@ export const Select = ({
           placeholder={placeholder ?? 'Search'}
           width={width}
           selectedValues={selectedValues}
-          onBlur={onBlur}
+          onBlur={handleOnBlur}
           onSelect={onSelect}
           onFocus={handleFocus}
           inputValue={inputValue}


src/@next/Select/Select.tsx
```diff
@@ -114,6 +114,14 @@ export const Select = ({
     setMenuOptions(newState);
   };
 
+  const handOnBlur = () => {
+    setTimeout(() => {
+      setPopoverActive(false);
+    }, 101);
```
yongchenglow, line 120: We have to set a timeout of more than 100 because of https://github.com/glints-dev/glints-aries/blob/0655a3bc8579d129461ad4fafd0f03788ef6e94e/src/@next/Select/components/SearchableSelectInput/SearchableSelectInput.tsx#L134

yongchenglow: This is a hacky solution but I guess it prevents a re-write

ninariccimarie: @yongchenglow thanks for submitting this.
upon checking, this solution still has the same issues found here: https://github.com/glints-dev/glints-aries/pull/1000#pullrequestreview-1783309335

ninariccimarie: @yongchenglow it's still having some issues on other types like the searchable single select:
Kapture 2023-12-27 at 10 51 12

it seems it could still lead to some issues since onBlur event goes first then the onClick. I've read from [here](https://stackoverflow.com/questions/17769005/onclick-and-onblur-ordering-issue) we can try replacing our onClick to onMouseDown to let the browser reorder the event handlers. Maybe we can try this?

yongchenglow: Will be closing this MR. We will adjust to use our custom component combo box which are not facing these issues. We are facing several other issues as well which is heavily affecting our forms

PR #1010, opened 2024-02-28
IndexTable: specify to apply only on mozilla

ninariccimarie:
### Issue
- `scrollbar-color` is introducing a scrollbar issue with Windows OS on Chrome

### Solution
- apply Firefox specific styles only to itself


chiahou (approved):
lgtm, thx

aguss787 (approved):
don't we all love different browser env :)
lgtm! 10/10 PR

PR #1011, opened 2024-02-29
Icon: add ri-user-add-fill

ninariccimarie:


src/@next/Icon/icons/icons.ts
```diff
@@ -164,6 +164,7 @@ export const iconNames = [
   'ri-twitter-line',
   'ri-upload-fill',
   'ri-upload-line',
+  'ri-user-add-fill',
   'ri-user-line',
```
chiahou, line 169: hi @ninariccimarie what about user-add-line? Line style is used in default state, where Fill style appears on hover

ninariccimarie: @chiahou do we need it? we usually just add what we need

ninariccimarie: as discussed, we need the user-add-line 👍 I've included it in @chiahou

aguss787 (approved):
lgmt!

chiahou (approved):

PR #1016, opened 2024-03-27
Modal: custom left and right component

ganiirsyadi:
https://glints.atlassian.net/browse/EP-2901

### DIFF
--- a/src/@next/Modal/Modal.tsx
+++ b/src/@next/Modal/Modal.tsx
@@ -14,29 +14,41 @@ import {
   StyledButtonContainer,
   StyledModalCloseButton,
   StyledModalBackButton,
+  StyledModalLeftAndRightComponent,
 } from './ModalStyle';
 
-export interface ModalProps extends React.HTMLAttributes<HTMLDivElement> {
-  isOpen?: boolean;
-  header?: React.ReactNode;
-  headerDescription?: React.ReactNode;
-  children?: React.ReactNode;
-  /** This action will be tied to primary button on footer */
-  primaryAction?: ComponentAction;
-  /** This action will be tied to basic button on footer */
-  secondaryAction?: ComponentAction;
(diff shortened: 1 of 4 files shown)


src/@next/Modal/Modal.tsx
```diff
@@ -55,6 +58,8 @@ export const Modal = React.forwardRef<HTMLDivElement, ModalProps>(
       onClose,
       onBack,
       zIndexOverride = 999,
+      leftComponent,
```
Afsal102, line 70: wondering if we can make it a bit more meaningful, ex: forward, backward icon/component

src/@next/Modal/ModalStyle.ts
```diff
@@ -12,7 +12,7 @@ export const StyledModalWrapper = styled.div<
   inset: 0;
   background-color: rgba(45, 45, 45, 0.5);
   display: flex;
-  flex-direction: column;
```
Afsal102, line 15: is this expected? i think by default it should be column instead of a row, if we have the left and right icons then this should be row?

src/@next/Modal/ModalStyle.ts
```diff
@@ -147,3 +147,9 @@ export const StyledModalBackButton = styled(StyledModalButton)`
 export const StyledModalCloseButton = styled(StyledModalButton)`
   margin-left: 16px;
 `;
+
+export const StyledModalLeftAndRightComponent = styled.div`
+  @media (max-width: ${Breakpoints.large}) {
+    display: none;
+  }
```
Afsal102, line 154: do we need to have a max width here?  not sure if the plan is to only allow icon buttons to show or any kind of custom component can be rendered?

Afsal102 (approved):

Afsal102 (approved):

Afsal102 (approved):

PR #1022, opened 2024-04-24
@next IndexTable: fix display nothing issue if all rows disabled

music1353:
Index Table will display empty if all data is unselectable, a design mistake from the official Polaris library.
Can refer to https://codesandbox.io/p/sandbox/sweet-resonance-jtw9yw
This MR is the solution for if all rows are disabled state.


ninariccimarie (approved):
LGTM! thanks

PR #1024, opened 2024-05-02
feat: update radio button input types

Afsal102:
changes the label and helperText prop to react node

### DIFF
--- a/src/@next/RadioButton/RadioButton.tsx
+++ b/src/@next/RadioButton/RadioButton.tsx
@@ -1,16 +1,16 @@
-import React from 'react';
+import React, { ReactNode } from 'react';
+import { Colors } from '..';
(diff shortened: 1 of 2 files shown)


ganiirsyadi:
image

This will be inside a Typography span component, we might need to handle this, if it's string we put it inside Typography, else no need to wrap it with Typography

ganiirsyadi:
> image
> This will be inside a Typography span component, we might need to handle this, if it's string we put it inside Typography, else no need to wrap it with Typography

> See Button implementation with the children props

@ganiirsyadi thats not ideal as the typography should be handled by the component itself (ex: size) else there would be no consistency as the clients would need to wrap the label with a typography on their side

ganiirsyadi (approved):

PR #1027, opened 2024-05-30
Button: Update the onClick handler

pawan-glints:
Update onClick handler to no trigger the callback if button is in loading state


src/@next/Button/Button.tsx
```diff
@@ -45,6 +46,13 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       return null;
     };
 
+    const onClickHandler = (...args: Parameters<ButtonProps['onClick']>) => {
+      if (loading) {
```
Afsal102: can we add the disabled condition too

Afsal102 (approved):
LGTM!

PR #1028, opened 2024-06-03
@next Icon: add ri-flag

music1353:


src/@next/Icon/icons/ri-flag-line.svg
```diff
@@ -0,0 +1,3 @@
+<svg viewBox="0 0 18 19" xmlns="http://www.w3.org/2000/svg">
```
ninariccimarie: @music1353 can you double check the exported file? it seems the size is a bit different 🤔

music1353: @ninariccimarie thanks adjusted!

ninariccimarie: @music1353 it should probably be the other way around. the size is usually the one with the 0 0 24 24 instead. with the hollow space around the icon.

src/@next/Icon/icons/ri-flag-fill.svg
ninariccimarie, line 1: Screenshot 2024-06-03 at 4 23 07 PM
@music1353 please check for this one. it seems the lines were missed during the export?

ninariccimarie (approved):
LGTM!

PR #1030, opened 2024-06-05
feat: Add sublabel to select component options

farhanabi:


chiahou (approved):

src/@next/Menu/components/MenuOptionLabel.tsx
```diff
@@ -1,11 +1,33 @@
...
 import { Neutral } from '../../utilities/colors';
+import { MenuOptionLabelContainer } from './MenuOptionLabelStyle';
+
+export const MenuOptionLabel = ({
+  label,
+  sublabel,
+}: {
+  label: React.ReactNode;
+  sublabel?: React.ReactNode;
+}) => {
```
ganiirsyadi: can we extract this Props type

```ts
type Props = {
  // ...
}
```

src/@next/Menu/Menu.tsx
```diff
@@ -11,6 +11,7 @@ export interface Option {
   disabled?: boolean;
   id?: string;
   label: string | React.ReactNode;
+  sublabel?: string | React.ReactNode;
```
ganiirsyadi: string is a member of ReactNode type, so no need to re-state it.

```ts
    type ReactText = string | number;
    type ReactChild = ReactElement | ReactText;
    type ReactFragment = {} | Iterable<ReactNode>;
    type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
```

src/@next/Menu/components/MenuOptionLabel.tsx
```diff
@@ -1,11 +1,33 @@
...
-export const MenuOptionLabel = ({ label }: { label: React.ReactNode }) => {
   return (
-    <Typography as="span" variant="body1" color={Neutral.B18}>
-      {label}
-    </Typography>
+    <MenuOptionLabelContainer sublabel={sublabel}>
+      <Typography as="span" variant="body1" color={Neutral.B18}>
+        {label}
+      </Typography>
+      {sublabel && (
```
ganiirsyadi: This && condition is unnecessary, we already have an early return, subLabel should be defined at this point

ganiirsyadi (changes requested):

src/@next/Menu/components/MenuOptionLabel.tsx
farhanabi: Updated!

src/@next/Menu/Menu.tsx
farhanabi: Updated!

src/@next/Menu/components/MenuOptionLabel.tsx
farhanabi: Sorry I missed it. Updated!

ganiirsyadi (approved):

PR #1035, opened 2024-06-24
Tabs: add disabled props to TabContent

ganiirsyadi:


src/@next/Tabs/Tabs.tsx
```diff
@@ -13,6 +13,8 @@ export type TabModel = {
   id?: string;
   /** Content of the type header */
   content: React.ReactNode;
+  /** Disable onClick handler */
```
Afsal102: comment is not needed?

src/@next/Tabs/TabStyle.ts
```diff
@@ -98,12 +98,14 @@ export const StyledTabButton = styled.button`
   width: 100%;
   padding: 0 ${space4};
 
-  &:hover {
-    color: ${Neutral.B18};
-  }
+  &[data-disabled='false'] {
```
Afsal102, line 101: do we have some style overrides for disabled state?

Afsal102 (changes requested):

src/@next/Tabs/TabStyle.ts
ganiirsyadi: Based on the Figma here

img_v3_02c6_67c26822-2c0a-4573-ba9d-f8ae6d43c1hu

We don't show the HOVER ui when the tab is disabled, no underline no bolder text

For the tooltip part I will confirm with bernard tomorrow if it's okay to show the tooltip only if the cursor hover on the text (in this case I can just wrap the Tooltip on my Employer project tab content component)

Cause, if I need to show the tooltip as soon as the cursor enter the tab area (even outside the text) I will need to update the Aries again to support it.

ganiirsyadi: https://www.loom.com/share/f4ea888bd9924770a071ab8e517d635d

Afsal102: what about the cursor?

src/@next/Tabs/Tabs.tsx
ganiirsyadi: Removed

src/@next/Tabs/TabStyle.ts
ganiirsyadi: not so sure, will wait for bernard answer tomorrow

ganiirsyadi: I scheduled the message lol, don't want to disturb him late at night

Afsal102 (approved):

src/@next/Tabs/TabStyle.ts
ganiirsyadi: Updated

- Added custom wrapper, this will allow us to wrap the individual tab with any component, including the tooltip
- Cursor will be default if tab is disabled, pointer otherwise

src/@next/Tabs/Tabs.tsx
```diff
@@ -109,15 +112,21 @@ export const Tabs = React.forwardRef<HTMLDivElement, TabsProps>(function Tabs(
...
     return (
       <StyledLi key={`${tabId}-${index}`} className="tab-item">
-        <Tab
-          id={tabId}
-          key={`tab-${tabId}-${index}`}
-          content={tab.content}
-          onSelect={() => handleSelectedIndexChanged(index)}
-          selected={index === selectedTabIndex}
-        ></Tab>
+        {tab.contentWrapper ? tab.contentWrapper(tabContent) : tabContent}
```
ganiirsyadi, line 129: Reason on why I don't place the wrapper outside StyledLi is, it will break current styling and will add more changes that may be risky.

Having it inside the StyledLi doesn't affect the styling.

Afsal102 (approved):

PR #1042, opened 2024-07-11
feat: adds icons

Afsal102:


src/@next/Icon/icons/ri-questionnaire-line.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg"
+    viewBox="0 0 24 24"
+    fill="currentColor">
```
ganiirsyadi: I recall nina mentioned that we should remove the fill attribute from the svg element so it will use the default fill color from Icon component

src/@next/Icon/icons/ri-questionnaire-fill.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg"
+    viewBox="0 0 24 24"
+    fill="currentColor">
```
ganiirsyadi: same here

src/@next/Icon/icons/ri-questionnaire-line.svg
Afsal102: yup, not sure why the export from remix had this fill property

ganiirsyadi (approved):

PR #1085, opened 2025-04-19
Add support for loading state on switch component

Yayo-Arellano:
Add support for loading state on switch component

PR #1086, opened 2025-04-21
feat: Add support for loading state on switch component #1085

Yayo-Arellano:
Add support for loading state on switch component


davidzzz (approved):

src/@next/Switch/SwitchStyle.tsx
```diff
@@ -2,6 +2,13 @@ import styled from 'styled-components';
 import { Breakpoints, Spacing, Typography } from '..';
 
 import { Blue, Neutral } from '../utilities/colors';
+import { keyframes } from 'styled-components';
+
+const spin = keyframes`
```
Afsal102, line 7: can we use the `<Spinner />` ?

Yayo-Arellano: What do you mean. Can you elaborate in the comment? @Afsal102

src/@next/Switch/SwitchStyle.tsx
```diff
@@ -67,13 +74,25 @@ export const SwitchStyle = styled.div`
...
+    border: 2px solid ${Neutral.B100};
+    border-top-color: transparent;
+    background: transparent;
+    animation: ${spin} 0.6s linear infinite;
+    transform: rotate(0deg);
+  }
+
+  &[data-loading='true'] {
+    pointer-events: none;
+  }
```
Afsal102, line 95: please add test cases for the loading state of the switch

src/@next/Switch/Switch.tsx
```diff
@@ -85,6 +87,7 @@ export const Switch = React.forwardRef<HTMLInputElement, SwitchProps>(
         />
         <SwitchStyle
           data-disabled={disabled}
```
Afsal102, line 89: should we disable this as well?

Yayo-Arellano: No, we should not disable this. When is in loading we disable the switch to be clickable but the UI does not change tp disabled state.

src/@next/Switch/SwitchStyle.tsx
Yayo-Arellano: Done. I added one more commit with switch variants.

src/@next/Switch/SwitchStyle.tsx
Yayo-Arellano: I tried but could not make it work, the problem is that I have to replace the knob with the spinner. The knob is declared in the style and If I want to use the Spinner It has to be on the component. I am not sure how can I make the spinner replace the knob, even if the position is left or right. 

Even I try to do 
```
export const spin = styled(Spinner)``;

// Later in the style using it does not work:

  &[data-loading='true']::before {
    border: 2px solid ${Neutral.B100};
    border-top-color: transparent;
    background: transparent;
    animation: ${spin};
    transform: rotate(0deg);
  }
```

Afsal102 (approved):

PR #1104, opened 2026-01-04
Update colors to new branding colors

Yayo-Arellano:
Update colors to match new branding colors


nmtuan2001 (changes requested):
Should we update Blue.S95 too? We use it almost interchangeably with Blue.Brand on employers

nmtuan2001 (approved):

